PR #87, opened 2020-12-02
fix: update certificate resource when changes are made

### LINKED ISSUE #43: Don't get valid certs when creating new tenants
After an install, add a new tenant to the cluster config file, then run the create command again (this is the simplest way to reproduce by effectively "adding" a tenant). The new tenant will be created, but the cert is invalid for this tenant's domain. Cert is fine for original domains. @sreis seems like the cert isn't getting copied over to the new tenant's namespace or maybe we're not generating a new cert to cover this domain, now that we combine all subdomains into the same cert?

**This cluster has `letsencrypt-prod` certs and all other tenants created during install have valid certs**.

Adding a tenant after the original install no longer get's a valid cert:

Screen Shot 2020-11-30 at 10 40 23 PM

sreis:
Fix #43

When we add a new tenant, we need to check if the Certificate spec changed to trigger an update of the certificate resource in kubernetes. This will make cert-manager reconcile the resource and issue a new certificate with Let's Encrypt.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -530,7 +530,7 @@ node_modules:
 
 .PHONY: unit-tests
 unit-tests: yarn-frozen-lockfile
-	cd lib/kubernetes && yarn test
+	cd lib/kubernetes && CI=true yarn test
 
 .PHONY: preamble
 preamble:
--- a/Makefile
+++ b/Makefile
@@ -528,6 +528,10 @@ yarn.lock: package.json $(PACKAGE_JSON) node_modules
 node_modules:
 	mkdir -p node_modules
 
+.PHONY: unit-tests
+unit-tests: yarn-frozen-lockfile
+	cd lib/kubernetes && yarn test
+
 .PHONY: preamble
 preamble:
 	bash "ci/preamble.sh"
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -24,11 +24,19 @@ steps:
       - ci/check-docs-fastpath-helper.sh fastpath-stdout FASTPATHEXIT
     soft_fail:
       - exit_status: 42
+  - label: "🔨 unit tests"
+    key: "unit-tests"
+    depends_on:
+      - "preamble"
+      - "check-docs"
+    command:
+      - cd "${OPSTRACE_PREBUILD_DIR}" && make ci-unit-tests
   - label: "🔨 main test (GCP)"
     key: "maintest-gcp"
     depends_on:
       - "preamble"
       - "check-docs"
+      - "unit-tests"
     env:
       OPSTRACE_CLUSTER_NAME: "bk-${BUILDKITE_BUILD_NUMBER}-${BUILDKITE_COMMIT:0:3}-g"
       OPSTRACE_BUILD_DIR: "/tmp/opstrace-bk-${BUILDKITE_BUILD_NUMBER}-${BUILDKITE_COMMIT:0:7}-1-gcp"
@@ -42,6 +50,7 @@ steps:
     depends_on:
       - "preamble"
       - "check-docs"
+      - "unit-tests"
     env:
       OPSTRACE_CLUSTER_NAME: "bk-${BUILDKITE_BUILD_NUMBER}-${BUILDKITE_COMMIT:0:3}-a"
       OPSTRACE_BUILD_DIR: "/tmp/opstrace-bk-${BUILDKITE_BUILD_NUMBER}-${BUILDKITE_COMMIT:0:7}-1-aws"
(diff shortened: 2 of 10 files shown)


sreis:
Decided to add unit tests in `lib/kubernetes` using `jest` as in the other modules (for example, `packages/app`).

This PR also adds a new step to the CI pipeline to run the unit tests in `lib/kubernetes` directory. We will add more as we go.

This is the output of running them locally:
```
    > yarn test
    yarn run v1.22.5
    $ jest --coverage
     PASS  src/equality/Certificate.spec.ts
      ✓ should return true when certificates are empty (3 ms)
      ✓ should not be equal when certificate annotations do not match
      ✓ should not be equal when certificate commonName does not match (1 ms)
      ✓ should not be equal when certificate dnsNames does not match
      ✓ should not be equal when certificate isCa does not match
      ✓ should not be equal when certificate issuerRef does not match (1 ms)
      ✓ should not be equal when certificate secretName does not match
    
    ----------------|---------|----------|---------|---------|-------------------
    File            | % Stmts | % Branch | % Funcs | % Lines | Uncovered Line #s
    ----------------|---------|----------|---------|---------|-------------------
    All files       |      96 |    92.86 |     100 |   95.83 |
     Certificate.ts |      96 |    92.86 |     100 |   95.83 | 26
    ----------------|---------|----------|---------|---------|-------------------
    Test Suites: 1 passed, 1 total
    Tests:       7 passed, 7 total
    Snapshots:   0 total
    Time:        1.148 s
    Ran all test suites.
    Done in 1.67s.
```

`jest` also has code coverage. The output is saved to `lib/kubernetes/coverage` and can be opened using a web browser. It should look like the following screenshot:

Screenshot from 2020-12-04 16-16-23

sreis:
Tests failed with a new CI instability https://github.com/opstrace/opstrace/issues/101. Retrying...

jgehrcke:
Last build was affected by instability #114. Started a new build: https://buildkite.com/opstrace/prs/builds/3165

jgehrcke:
> Started a new build: https://buildkite.com/opstrace/prs/builds/3165

Five loki tests failed: https://buildkite.com/opstrace/prs/builds/3165#6357df46-61ad-4aa0-bda7-881c7c8b0b14/3880-5681

sreis:
> Five loki tests failed: https://buildkite.com/opstrace/prs/builds/3165#6357df46-61ad-4aa0-bda7-881c7c8b0b14/3880-5681

So we have a new bug then since the AWS build run was green. Created https://github.com/opstrace/opstrace/issues/115 to track.

Makefile
```diff
@@ -508,6 +508,11 @@ website-build:
 deploy-testremote-teardown:
 	bash "ci/deploy-testremote-teardown.sh"
 
+.PHONY: unit-tests
+unit-tests:
+	yarn --frozen-lockfile
```
jgehrcke, line 513: Okay... :) in the current setup _within_ the preamble we have a containerized environment where `node_modules` is set up through a big `yarn install` (_in_ the container file system) and practically the only build artifact of relevance that is carried over to other steps is the single-binary CLI build (through the "prebuild" directory).

Here, we're doing another biiig `yarn install` (_in_ the container file system) "just" for being able to run these unit tests. 

Certainly something to consolidate in the future.

What are your thoughts about running these unit tests as part of the preamble, like we currently do for the go unit tests?

(note1: not suggesting to change anything _now_ :)).

(note2: sharing a cache between these `yarn install`s failed rather miserably.)

sreis: > Certainly something to consolidate in the future.

Well #119 helps with that already. The [build dir](https://github.com/opstrace/opstrace/blob/main/Makefile#L406) is available in the container so the `node_modules` are only installed once.

I'm waiting for the currents tests to finish since yesterday they triggered some flakes. I'll update this PR with the merge conflicts after they finish.

> What are your thoughts about running these unit tests as part of the preamble, like we currently do for the go unit tests?

I was thinking of following up with a PR to move the tests out of the preamble. The preamble step would check everything builds. The unit-tests step would run the (well) the unit tests. And then the next steps would continue as today. 

The point is to make it obvious to know know what's broken. It's the build. It's the unit tests. It's the integration/e2e tests. Etc.

jgehrcke: > The build dir is available in the container

Huh. I created this system and keep having wrong intuitions about what's happening :). Too complex.

> I was thinking of following up with a PR to move the tests out of the preamble. The preamble step would check everything builds. The unit-tests step would run the (well) the unit tests. And then the next steps would continue as today.

certainly a +1!

I forgot to say that before; a build job that is meant to run unit tests is certainly great; much better semantics than piggy-backing this into the preamble stage.

Big thanks for taking this on.

Issue #51, opened 2020-12-01
ci: lint entire TS codebase with ESLint


jgehrcke:
The goal is that the entire TypeScript code base is linted (and passing!) using ESLint (and a consistent set of rules).

This is a bigger effort (sometimes requiring quite involved code changes) and we should probably decompose this:
- installer
- uninstaller
- lib/aws
- lib/gcp
-...

Started doing that for the CLI:

https://github.com/opstrace/opstrace/blob/1e4297517961d552fc0238ed3693b9b96e734626/packages/cli/package.json#L19

We certainly want to end up in a state where we do not have `"lint": "echo done"` anymore in our code base :-) ([example](https://github.com/opstrace/opstrace/blob/77bf499be15d87ed319f192063a47aae0572187d/packages/uninstaller/package.json#L16))

The rules (respected by ESLint in CI, and also by the ESLint extension in VS Code) live here: https://github.com/opstrace/opstrace/blob/main/.eslintrc.js (and can be adjusted).

jgehrcke:
> we could probably say it's a duplicate 

Oh, yes, sorry. My original intent was to close the private issue, somehow forgot that -- so, this is a half-way transferred issue! Leaving it to @juliachvyrova now :).

Issue #62, opened 2020-12-01
ci: gcp: create SQLInstance: failed: reached the max instance per project/creator limit


sreis:
**Describe the bug**

```
[2020-12-01T20:13:53Z] 2020-12-01T20:13:53.517Z error: error during cluster creation (attempt 3):
[2020-12-01T20:13:53Z] GaxiosError: Failed to create instance because the project or creator has reached the max instance per project/creator limit.
[2020-12-01T20:13:53Z]     at Gaxios._request (/snapshot/build/node_modules/gaxios/src/gaxios.ts:117:15)
[2020-12-01T20:13:53Z]     at runMicrotasks (<anonymous>)
[2020-12-01T20:13:53Z]     at processTicksAndRejections (internal/process/task_queues.js:97:5)
[2020-12-01T20:13:53Z]     at JWT.requestAsync (/snapshot/build/node_modules/google-auth-library/build/src/auth/oauth2client.js:343:18) {
[2020-12-01T20:13:53Z]   response: [Object],
[2020-12-01T20:13:53Z]   config: [Object],
[2020-12-01T20:13:53Z]   code: 403,
[2020-12-01T20:13:53Z]   errors: [Array]
[2020-12-01T20:13:53Z] }
```
https://buildkite.com/opstrace/scheduled-main-builds/builds/1285#f7b824b4-4945-488d-8367-0cc01229c654/200-1075

triclambert:
Request approved, up to the _maximum_ allowable instances.

![DEE50250-27AC-4056-BF19-D36EA4091C3C_1_105_c](https://user-images.githubusercontent.com/19239758/100953937-6f479d80-34c8-11eb-9666-73bd596d14de.jpeg)

spahl:
damn I bet we already do more that 100 build a week. Maximum for sure?

triclambert:
I didn't specifically ask about that one, but I did ask if the 7 days was fixed...

> Regarding your question, I’m afraid that this limit of waiting 7 days is fixed for now.

spahl:
Ah the time window is fine but we probably need 250-300 to get a margin of
safety. Did you explain our use case that we spin these up in CI?

triclambert:
I did; they explicitly asked.  I'll reply with a thanks and one last check if the 100 can increase.  

(FWIW, we're dealing with a limit that is the function of the max over a time window...  if we shortened the window to 2 days we'd be fine.)

triclambert:
Their reply...

[...]

PR #97, opened 2020-12-03
Fix eslint

juliachvyrova:
Part of task #51 

Fix errors and warnings in:

#### /packages:
- [x] packages/app
- [x] packages/cli
- [x] packages/config
- [ ] packages/controller
- [x] packages/controller-config
- [x] packages/e2e-testing
- [x] packages/installer
- [x] packages/tenants
- [x] packages/uninstaller

Comments:
We have bunch of warnings ( ~71 ) for `port: 8080 as any`. As types are autogenerated based on schema and those part uses
`anyOf: [
                                        {
                                          type: "integer"
                                        },
                                        {
                                          type: "string"
                                        }
                                      ]` type is inferred as `object` instead of `String | number`.
I suppose we need either ignore such errors for now or check how to change tool to correctly generate types.

### DIFF
--- a/packages/config/src/aws.ts
+++ b/packages/config/src/aws.ts
@@ -48,9 +48,26 @@ export const REGION_EKS_AMI_MAPPING: Record<string, string> = {
 
 export const KNOWN_AWS_REGIONS: string[] = Object.entries(
   REGION_EKS_AMI_MAPPING
-).map(([region, _]) => region);
+).map(([region]) => region);
 
-export function getAWSConfig(ccfg: NewRenderedClusterConfigType) {
+type AWSConfigType = {
+  k8sVersion: string;
+  endpointPublicAccess: boolean;
+  endpointPrivateAccess: boolean;
+  region: string;
+  zone: string;
+  vpc: {
+    CidrBlock: string;
+  };
+  subnets: { CidrBlock: string; AvailabilityZone: string; Public: boolean }[];
+  masterAuthorizedNetworks: string[];
+  instanceType: string;
+  imageId: string;
+};
+
+export function getAWSConfig(
+  ccfg: NewRenderedClusterConfigType
+): AWSConfigType {
   // Using k3s instead of EKS will wipe a lot of these config options, which is
   // why I'm hardcoding the region, vpc, and subnets for now.
   // EKS has region/zone specific imageId so avoid computing that
--- a/packages/config/src/aws.ts
+++ b/packages/config/src/aws.ts
@@ -60,6 +60,11 @@ type AWSConfigType = {
     CidrBlock: string;
   };
   subnets: { CidrBlock: string; AvailabilityZone: string; Public: boolean }[];
+  rdsSubnets: {
+    CidrBlock: string;
+    AvailabilityZone: string;
+    Public: boolean;
+  }[];
   masterAuthorizedNetworks: string[];
   instanceType: string;
   imageId: string;
--- a/packages/uninstaller/src/gcp.ts
+++ b/packages/uninstaller/src/gcp.ts
@@ -36,31 +36,37 @@ export function* destroyGCPInfra(): Generator<
   // eslint-disable-next-line @typescript-eslint/no-explicit-any
   any
 > {
+  if (!destroyConfig.gcpProjectID) {
+    throw Error("gcp project id not specified in credentials");
+  }
+  if (!destroyConfig.gcpRegion) {
+    throw Error("gcp region not specified in credentials");
+  }
   log.info(`Ensure cert-manager service account deletion`);
   yield call(ensureServiceAccountDoesNotExist, {
     name: `${destroyConfig.clusterName}-cert-manager`,
-    projectId: destroyConfig.gcpProjectID ?? "",
+    projectId: destroyConfig.gcpProjectID,
     role: "roles/dns.admin"
   });
 
   log.info(`Ensure external-dns service account deletion`);
   yield call(ensureServiceAccountDoesNotExist, {
     name: `${destroyConfig.clusterName}-external-dns`,
-    projectId: destroyConfig.gcpProjectID ?? "",
+    projectId: destroyConfig.gcpProjectID,
     role: "roles/dns.admin"
   });
 
   log.info(`Ensure cortex service account deletion`);
   yield call(ensureServiceAccountDoesNotExist, {
     name: `${destroyConfig.clusterName}-cortex`,
-    projectId: destroyConfig.gcpProjectID ?? "",
+    projectId: destroyConfig.gcpProjectID,
     role: "roles/storage.admin"
   });
 
   log.info(`Ensure loki service account deletion`);
   yield call(ensureServiceAccountDoesNotExist, {
     name: `${destroyConfig.clusterName}-loki`,
-    projectId: destroyConfig.gcpProjectID ?? "",
+    projectId: destroyConfig.gcpProjectID,
     role: "roles/storage.admin"
   });
 
@@ -86,14 +92,14 @@ export function* destroyGCPInfra(): Generator<
   yield call(
     ensureGatewayDoesNotExist,
     destroyConfig.clusterName,
-    destroyConfig.gcpRegion ?? ""
+    destroyConfig.gcpRegion
   );
 
   log.info(`Destroying Subnet`);
   yield call(
     ensureSubNetworkDoesNotExist,
     destroyConfig.clusterName,
-    destroyConfig.gcpRegion ?? ""
+    destroyConfig.gcpRegion
   );
 
   log.info(`Destroying VPC`);
--- a/packages/uninstaller/src/gcp.ts
+++ b/packages/uninstaller/src/gcp.ts
@@ -14,7 +14,7 @@
  * limitations under the License.
  */
 
-import { call } from "redux-saga/effects";
+import { call, CallEffect, ForkEffect, JoinEffect } from "redux-saga/effects";
 import {
   ensureGKEDoesNotExist,
   ensureGatewayDoesNotExist,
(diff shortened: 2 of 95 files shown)


MatApple:
> We have bunch of warnings ( ~71 ) for port: 8080 as any. As types are autogenerated based on schema and those part uses
anyOf: [ { type: "integer" }, { type: "string" } ] type is inferred as object instead of String | number.
I suppose we need either ignore such errors for now or check how to change tool to correctly generate types.

I think for now the most practical approach would be to ignore these errors specifically!

packages/config/src/aws.ts
```diff
@@ -46,9 +46,31 @@ export const REGION_EKS_AMI_MAPPING: Record<string, string> = {
 
 export const KNOWN_AWS_REGIONS: string[] = Object.entries(
   REGION_EKS_AMI_MAPPING
-).map(([region, _]) => region);
+).map(([region]) => region);
```
jgehrcke, line 49: Ha! :) Thank you.

jgehrcke:
Thanks for this (sometimes quite tedious) work!

Let's try to integrate some of this asap; before merge conflicts arise!

> I suppose we need either ignore such errors for now or check how to change tool to correctly generate types.

Well, we certainly do the "for now" approach and then merge early and often :).

packages/uninstaller/src/gcp.ts
```diff
@@ -28,33 +28,39 @@ import { destroyDNS } from "@opstrace/dns";
...
+  JoinEffect | CallEffect | ForkEffect | Generator<ForkEffect, Task[], Task>,
+  void,
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  any
+> {
   log.info(`Ensure cert-manager service account deletion`);
   yield call(ensureServiceAccountDoesNotExist, {
     name: `${destroyConfig.clusterName}-cert-manager`,
-    projectId: destroyConfig.gcpProjectID!,
+    projectId: destroyConfig.gcpProjectID ?? "",
```
sreis, line 42: @juliachvyrova what would be the typescript best practice to ensure `destroyGCPInfra` would only be called if these (and other) properties are set correctly?

jgehrcke:
Ah, by the way: this is the corresponding public issue: https://github.com/opstrace/opstrace/issues/51

packages/uninstaller/src/gcp.ts
MatApple: TS doesn't do runtime type checking, so you have to do that yourself. Best way would be to assert the properties you expect are indeed there, as you'd expect them and throw and Error if not.

Only the properties that can __change__ at runtime (i.e. user provided config for example) need to be checked at runtime, since the rest will benefit from static analysis during compilation. If you don't want to do your own simple typecheck, some libraries that do runtime checking:

https://github.com/jquense/yup
https://github.com/gcanti/io-ts

There are many.

In the case here, it is probably safer to add this before the call to `ensureServiceAccountDoesNotExist`:
```
if (!destroyConfig.gcpProjectID) {
  throw Error("gcp project id not specified in credentials");
}
```

MatApple:
Just to be clear

> Well, we certainly do the "for now" approach and then merge early and often :).

We are going to **explicitly ignore** the port errors on Kubernetes objects with `// eslint-disable-next-line`.

packages/uninstaller/src/gcp.ts
juliachvyrova: @MatApple Good idea! I thought about checking it, because in some places error will be raised anyway ( like when we get path from env variables and then try to get file by that path ), but having explicit error is always better for future debugging. :)

jgehrcke: > but having explicit error is always better for future debugging. :)

:100:

jgehrcke:
Great progress here! I noticed this TSC failure in the last CI run:
```
[2020-12-05T23:04:57Z] src/destroy.ts:55:24 - error TS2345: Argument of type '(e: Error, detail: { sagaStack: string; }) => void' is not assignable to parameter of type '(e: Error, detail: unknown) => void'.
[2020-12-05T23:04:57Z]   Types of parameters 'detail' and 'detail' are incompatible.
[2020-12-05T23:04:57Z]     Type 'unknown' is not assignable to type '{ sagaStack: string; }'.
[2020-12-05T23:04:57Z]
[2020-12-05T23:04:57Z] 55   await destroyCluster(util.smErrorLastResort);
```

pushed a quick workaround/fix for that to this branch: hoping that this helps -- maybe the last tiny step to green CI here? :) Excited to land this!

jgehrcke:
Controller log prefix, from build artifacts from https://buildkite.com/opstrace/prs/builds/3162#59d8325e-0eb2-41eb-b985-ac93bbfe1593/246-2556:
```
2020-12-06T12:40:17.192Z [32minfo[39m: Load kubeconfig. Assume to run in cluster, rely on privileged access
2020-12-06T12:40:17.194Z [32minfo[39m: fetching tenants
2020-12-06T12:40:17.227Z [32minfo[39m: starting kubernetes informers
2020-12-06T12:40:17.228Z [32minfo[39m: Starting informers
2020-12-06T12:40:17.243Z [32minfo[39m: Waiting for the Kubernetes cache to hydrate...
2020-12-06T12:40:17.306Z [31mwarning[39m: starting informer failed (will retry):  { response: [IncomingMessage], body: '404 page not found\n' }
2020-12-06T12:40:17.308Z [31mwarning[39m: starting informer failed (will retry):  { response: [IncomingMessage], body: '404 page not found\n' }
2020-12-06T12:40:17.309Z [31mwarning[39m: starting informer failed (will retry):  { response: [IncomingMessage], body: '404 page not found\n' }
2020-12-06T12:40:17.310Z [31mwarning[39m: starting informer failed (will retry):  { response: [IncomingMessage], body: '404 page not found\n' }
2020-12-06T12:40:17.313Z [31merror[39m: undefined
2020-12-06T12:40:17.332Z [31mwarning[39m: starting informer failed (will retry):  { response: [IncomingMessage], body: '404 page not found\n' }
2020-12-06T12:40:17.333Z [31mwarning[39m: starting informer failed (will retry):  { response: [IncomingMessage], body: '404 page not found\n' }
2020-12-06T12:40:17.334Z [31mwarning[39m: starting informer failed (will retry):  { response: [IncomingMessage], body: '404 page not found\n' }
2020-12-06T12:40:17.364Z [31mwarning[39m: starting informer failed (will retry):  { response: [IncomingMessage], body: '404 page not found\n' }
2020-12-06T12:40:18.244Z [32minfo[39m: Waiting for the Kubernetes cache to hydrate...
2020-12-06T12:40:19.245Z [32minfo[39m: Waiting for the Kubernetes cache to hydrate...
2020-12-06T12:40:20.246Z [32minfo[39m: Waiting for the Kubernetes cache to hydrate...
...
```

packages/uninstaller/src/gcp.ts
```diff
@@ -36,31 +36,37 @@ export function* destroyGCPInfra(): Generator<
   // eslint-disable-next-line @typescript-eslint/no-explicit-any
   any
 > {
+  if (!destroyConfig.gcpProjectID) {
+    throw Error("gcp project id not specified in credentials");
```
jgehrcke, line 40: This error message isn't quite true, it would in fact be misleading when this would blow up. When this blows up it indicates a "programmer error", whereas the error message with "in credentials" suggests an operational/user error :)

I think this should be a simple assertion in code (based on NodeJS' assert module in strict mode (which tsc understands)) that will lead to a proper error message:

```
import { strict as assert } from "assert";
...
assert(destroyConfig.gcpProjectID);
...
<here, tsc knows, that `destroyConfig.gcpProjectID` is not undefined / falsy>
```

jgehrcke:
I had a quick look at the packages/controller diff and didn't find the culprit leading to `error: undefined` based on code alone. We should probably separate out the controller changes into a different PR. We should probably also change error handling in the controller so that we see a stack trace corresponding to the problem hidden behind `error: undefined`.

packages/uninstaller/src/gcp.ts
sreis: I was thinking it should be possible to check the config object once (somewhere) and then we wouldn't have to rely on `obj!.someProperty` or `obj?.someProperty` type. I guess that's what `yup` is for but I was expecting not to do `!` or `?` afterwards.

packages/uninstaller/src/gcp.ts
jgehrcke: Oh, and here is another example with an explicit comment:

https://github.com/opstrace/opstrace/blob/45f8cf57264fe3fb5d59a86744d14d38db9a60a2/packages/installer/src/aws.ts#L188

```
  // done assert instead of N non-null-assertion.
  assert(internetGateway.InternetGatewayId);
```

MatApple:
Yes splitting out the work for the controller package into a separate PR makes sense. We can then land the rest and have a focussed effort on the controller linting separately.

jgehrcke:
In the last build during cluster creation we see that there is a new problem, another `error: undefined` in the logs:
```
[2020-12-07T21:16:15Z] 2020-12-07T21:16:15.432Z info: waiting for controller deployment (not yet found in informer-based state)
[2020-12-07T21:16:15Z] 2020-12-07T21:16:15.453Z warning: starting informer failed (will retry):  { response: [IncomingMessage], body: '404 page not found\n' }
[2020-12-07T21:16:15Z] 2020-12-07T21:16:15.461Z error: undefined
```
(https://buildkite.com/opstrace/prs/builds/3188#d36cca1f-4db8-4063-9a77-d731e9df68c3/252-3352)

jgehrcke:
For the record, copying my last comment from Slack:
> @jamakase @juliachvyrova I have just landed https://github.com/opstrace/opstrace/pull/129 -- do you maybe want to rebase #97 on latest head of main, and force-push? (I can also do so, and might do so if you don't get to replying soon :-)). This will hopefully help debugging. Thaaaanks for patience.

jgehrcke:
`main-test-gcp` is currently expected to fail as of #62

jgehrcke:
Here we go, that is what was hidden behind `error: undefined`:
```
[2020-12-08T16:00:24Z] 2020-12-08T16:00:24.356Z info: at least one replica is READY: desired state reached, continue
[2020-12-08T16:00:24Z] 2020-12-08T16:00:24.374Z warning: error during `yield put(event)`: SyntaxError: Unexpected end of JSON input
[2020-12-08T16:00:24Z]     at JSON.parse (<anonymous>)
[2020-12-08T16:00:24Z]     at Object.deserialize (/snapshot/build/node_modules/@opstrace/controller-config/build/utils.js:0)
```
(https://buildkite.com/opstrace/prs/builds/3200#fd53adde-48ed-4ad4-9ae9-6f5ca0ff763d/252-4735)

Jamakase:
@jgehrcke Good point! 

@juliachvyrova 
I suppose that instead of 
`JSON.parse(configMap.spec.data?.STORAGE_KEY || "")`
it should be
`JSON.parse(configMap.spec.data?.[STORAGE_KEY] ?? "")`

jgehrcke:
> `JSON.parse(configMap.spec.data?.[STORAGE_KEY] ?? "")`

Yes, for this PR this is fine! Thanks for fixing this :)

In the future, note to self: before we JSON-deserialize a string we should explicitly confirm that it's a string and that it's of non-zero length! All other cases should be handled deliberately / explicitly, depending on whether they are allowed to happen or not --, by either throwing an error in a controlled fashion or... :)

jgehrcke:
`main-test-aws` went green! :heavy_check_mark: 
:rocket: 
Thanks, great collaboration. Merging!

jgehrcke (approved):
:rocket:

PR #94, opened 2020-12-03
chore: improve docs only pr check

sreis:
Close #93

Tested locally by running the script.

This is the output of a docs only PR:
```
> BUILDKITE_PULL_REQUEST=92 bash ./ci/check-if-docs-pr.sh
+ echo 'BUILDKITE_PULL_REQUEST: 92'
BUILDKITE_PULL_REQUEST: 92
+ '[' -z 92 ']'
+ '[' 92 = false ']'
++ curl --silent -H 'Accept: application/vnd.github.v3+json' https://api.github.com/repos/opstrace/opstrace/pulls/92/files
++ jq -r '.[].filename'
+ FILES_EDITED_IN_PR='.markdownlint.json
README.md
ci/check-if-docs-pr.sh
docs/README.md
docs/quickstart.md'
+ ALLOWLIST='^docs/|.gitattributes|^.github|^README.md|^.markdownlint.json'
++ echo .markdownlint.json README.md ci/check-if-docs-pr.sh docs/README.md docs/quickstart.md
++ egrep -v '^docs/|.gitattributes|^.github|^README.md|^.markdownlint.json'
++ tr -d '[:space:]'
+ DOCS_ONLY_CHANGES=
+ '[' -z '' ']'
+ echo '--- docs only PR (92) - skipping next steps'
--- docs only PR (92) - skipping next steps
+ exit 0
```

The output of a PR that must run all the CI steps:
```
> BUILDKITE_PULL_REQUEST=89 bash ./ci/check-if-docs-pr.sh
+ echo 'BUILDKITE_PULL_REQUEST: 89'
BUILDKITE_PULL_REQUEST: 89
+ '[' -z 89 ']'
+ '[' 89 = false ']'
++ curl --silent -H 'Accept: application/vnd.github.v3+json' https://api.github.com/repos/opstrace/opstrace/pulls/89/files
++ jq -r '.[].filename'
+ FILES_EDITED_IN_PR=ci/deploy-testremote-teardown.sh
+ ALLOWLIST='^docs/|.gitattributes|^.github|^README.md|^.markdownlint.json'
++ echo ci/deploy-testremote-teardown.sh
++ egrep -v '^docs/|.gitattributes|^.github|^README.md|^.markdownlint.json'
++ tr -d '[:space:]'
+ DOCS_ONLY_CHANGES=ci/deploy-testremote-teardown.sh
+ '[' -z ci/deploy-testremote-teardown.sh ']'
[...]

PR #100, opened 2020-12-04
fix: retry curl command on transient errors

sreis:
Follow up to [this review request](https://github.com/opstrace/opstrace/pull/94#discussion_r535627347).

### DIFF
--- a/ci/utils.sh
+++ b/ci/utils.sh
@@ -1,0 +1,35 @@
+#
+# This file is meant to be imported by the scripts in the ci/ directory. It
+# contains a few useful helper functions.
+
+#
+# curl is a bash function with the same name as the binary to ensure we always
+# call this function instead of the command directly. The purpose is to set a
+# few default options in the curl command.
+#
+# curl manpage section https://curl.se/docs/manpage.html about retry flag:
+#
+# "If a transient error is returned when curl tries to perform a transfer, it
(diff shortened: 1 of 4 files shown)


ci/utils.sh
```diff
@@ -0,0 +1,35 @@
...
+#
+# If this option is used several times, the last one will be used."
+#
+curl() {
+    # circumvent normal bash function lookup
+    command curl \
+    --connect-timeout 10 \
+    --retry 3 \
+    --retry-delay 5 \
+    "${@}"
```
jgehrcke, line 34: Yes!!! :rocket:

jgehrcke: @sreis do we also want to add verbosity? For example, when curl retries now upon a 5xx HTTP response -- maybe we want to log that. On the other hand, we can try to remember adding `-v` in those places where we use this curl shim.

jgehrcke (approved):
This is great; and also necessary towards getting CI stability into a nice regime.

Next up: docker pull, docker push... haha.

ci/utils.sh
sreis: It prints a `warning` by default. Example message for a timeout:
```Warning: Transient problem: timeout Will retry in 1 seconds. 3 retries left.```

jgehrcke: Perfect.

PR #103, opened 2020-12-05
Use Redis as session store

MatApple:
This PR adds a very small Redis deployment and configures the Opstrace app to use Redis as the backing store for session management.


packages/app/src/server/server.ts
```diff
@@ -84,6 +85,15 @@ function createServer() {
   }
   app.use(sessionParser);
 
+  app.use(function sessionCheck(req, res, next) {
+    if (!req.session) {
+      return next(
+        new GeneralServerError(500, "connection to session backend lost")
```
jgehrcke, line 91: About the topic of potentially misleading error messages: are we sure that "connection to session backend lost" is correct / useful in all cases when this condition (`!req.session`) is true? As in: TCP connection lost? In which cases is `!req.session` true, architecturally? Is there any original error message here exposed by the framework/library that we could emit instead / additionally?

jgehrcke (approved):
Thank you!

Much better than hitting the cloud db for each UI-related HTTP request.

For the record, the rationale: we can afford to lose session state (worst case: users are logged out).

jgehrcke:
Let's integrate this. If this is by any means to early or not what was wanted then we have learned something :-).

sreis:
@MatApple any obvious advantages over `memcached`? We already deploy and have to support `memcached` because of `loki` and `cortex` so I'm wondering if we can consolidate the architecture around one caching layer.

packages/app/src/server/server.ts
MatApple: This is recommended by the library itself: https://github.com/tj/connect-redis#how-do-i-handle-lost-connections-to-redis

MatApple: Of course, in practice, we shall see

MatApple:
@sreis Memcached would have been a great alternative for this feature. However, the reason Redis has been selected is with some forward thinking in mind. The UI requires pub/sub for upcoming features and Memcached can't help with that. We should be able to have a highly available and horizontally scalable UI with two pependencies: Postgres as the datastore and Redis for key/value and pub/sub.

spahl:
Note that we can move loki and Cortex to redis if needed too. They support it. But not urgent at all.

packages/app/src/server/server.ts
jgehrcke: To be fair - they write `return next(new Error('oh no'))` which is _sad_. Error detail being swallowed.

PR #119, opened 2020-12-07
chore: prevent yarn install from running twice

sreis:
This improves developer productivity (slightly) by reducing the compile time since we only need to run `yarn install` when a `package.json` or `yarn.lock` changed.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -13,6 +13,7 @@ out
 .cache
 .build
 yarn-error.log
+yarn.lock.md5
 
 opstrace-config.yaml
 prometheus.yml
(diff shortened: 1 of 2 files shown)


jgehrcke (commented):
I was typing up a response here yesterday and have lost it in my forest of browser tabs. Meh.

I was basically saying that I hope this is worth the complexity / mental overhead :-).

Most importantly, I think we should talk to each other about what we run locally most often. Personally, I execute `yarn install` pretty much only manually, explicitly. I don't even run `make cli` anymore. Locally, I either run a CI-built CLI (that I freshly downloaded), or I use my local CLI build via `node ...index.js...`: I've tried to document that here: https://github.com/opstrace/opstrace/blob/main/docs/guides/contributor/workflows.md#local-checkout-quick-iteration-while-changing-the-cli

That is, I don't expect this to make things faster for "my" local workflows :).

But hey. Let's see! Thanks for doing work in this dirty corner of our endeavor.

PR #123, opened 2020-12-07
Fix lint in packages/controller

juliachvyrova:
Part of task #51

Fix errors and warnings in packages/controller

### DIFF
--- a/packages/controller/src/dbClient.ts
+++ b/packages/controller/src/dbClient.ts
@@ -19,7 +19,7 @@ import { getSdk } from "./dbSDK";
 // Make this optional, so that running the controller locally during dev
 // doesn't require a connection to the db.
 const endpoint = process.env.GRAPHQL_ENDPOINT;
-const adminSecret = process.env.HASURA_GRAPHQL_ADMIN_SECRET;
+const adminSecret = process.env.HASURA_GRAPHQL_ADMIN_SECRET ?? "";
 
 if (endpoint && !adminSecret) {
   throw Error(
@@ -30,14 +30,15 @@ if (endpoint && !adminSecret) {
 const client = endpoint
   ? new GraphQLClient(endpoint, {
       headers: {
-        "x-hasura-admin-secret": adminSecret!
+        "x-hasura-admin-secret": adminSecret
       }
     })
   : null;
 
 export type PromiseReturnType<T> = T extends PromiseLike<infer U> ? U : T;
-export type ClientResponse<T extends (args?: any) => {}> = PromiseReturnType<
-  ReturnType<T>
->;
+export type ClientResponse<
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  T extends (args?: any) => Record<string, any>
+> = PromiseReturnType<ReturnType<T>>;
 
 export default client ? getSdk(client) : null;
--- a/packages/controller/src/dbSDK.ts
+++ b/packages/controller/src/dbSDK.ts
@@ -29,8 +29,8 @@ export type Scalars = {
   Boolean: boolean;
   Int: number;
   Float: number;
-  timestamp: unknown;
-  timestamptz: unknown;
+  timestamp: number;
+  timestamptz: number;
   uuid: unknown;
 };
 
--- a/packages/controller/src/dbSDK.ts
+++ b/packages/controller/src/dbSDK.ts
@@ -17,7 +17,6 @@ import { GraphQLClient } from "graphql-request";
 import { print } from "graphql";
 import { GraphQLError } from "graphql-request/dist/types";
 import gql from "graphql-tag";
-
 export type Maybe<T> = T | null;
 export type Exact<T extends { [key: string]: unknown }> = {
   [K in keyof T]: T[K];
@@ -29,9 +28,9 @@ export type Scalars = {
   Boolean: boolean;
   Int: number;
   Float: number;
-  timestamp: number;
-  timestamptz: number;
-  uuid: unknown;
+  timestamp: any;
+  timestamptz: any;
+  uuid: any;
 };
 
 /** expression to compare columns of type Boolean. All fields are combined with logical 'AND'. */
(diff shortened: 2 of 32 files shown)


packages/controller/src/dbSDK.ts
```diff
@@ -28,9 +29,9 @@ export type Scalars = {
   Boolean: boolean;
   Int: number;
   Float: number;
-  timestamp: any;
-  timestamptz: any;
-  uuid: any;
+  timestamp: unknown;
```
Jamakase, line 32: Shouldn't it actually be number? Will it cause issues somewhere if we change it to number?

packages/controller/src/dbSDK.ts
```diff
@@ -28,9 +29,9 @@ export type Scalars = {
...
+  timestamptz: unknown;
```
Jamakase, line 33: Same here as above

packages/controller/src/dbSDK.ts
```diff
@@ -17,6 +17,7 @@ import { GraphQLClient } from "graphql-request";
 import { print } from "graphql";
```
MatApple, line 17: Unfortunately this file is generated by https://graphql-code-generator.com/ in `packages/app` and then copied to the controller package, so we'll have to find a way to ignore this file

packages/controller/src/resources/cortex/index.ts
```diff
@@ -327,7 +327,7 @@ export function CortexResources(
   let serviceAccountName: string | undefined = undefined;
   if (target === "gcp") {
     annotations = {
-      "iam.gke.io/gcp-service-account": gcp!.cortexServiceAccount
+      "iam.gke.io/gcp-service-account": gcp?.cortexServiceAccount
```
MatApple, line 330: We should explicitly assert that this exists before we rely on it with `assert(gcp?.cortexServiceAccount)`

packages/controller/src/resources/ingress/externalDns.ts
```diff
@@ -47,8 +47,8 @@ export function ExternalDnsResources(
   let annotations = {};
   if (target == "gcp") {
     annotations = {
-      "iam.gke.io/gcp-service-account": state.config.config!.gcp!
-        .externalDNSServiceAccount
+      "iam.gke.io/gcp-service-account":
+        state.config.config?.gcp?.externalDNSServiceAccount
```
MatApple, line 51: again here, we should explicitly `assert()` before we use it to ensure we throw an error if it doesn't exist

packages/controller/src/resources/ingress/certManager.ts
```diff
@@ -269,7 +269,7 @@ export function CertManagerResources(
   let annotations = {};
   if (target == "gcp") {
     annotations = {
-      "iam.gke.io/gcp-service-account": gcp!.certManagerServiceAccount
+      "iam.gke.io/gcp-service-account": gcp?.certManagerServiceAccount
```
MatApple, line 272: Assert this one too

packages/controller/src/resources/loki/index.ts
```diff
@@ -248,7 +248,7 @@ export function LokiResources(
   let serviceAccountName: string | undefined = undefined;
   if (target === "gcp") {
     annotations = {
-      "iam.gke.io/gcp-service-account": gcp!.lokiServiceAccount
+      "iam.gke.io/gcp-service-account": gcp?.lokiServiceAccount
```
MatApple, line 251: Here too

MatApple (changes requested):
Just a few changes to make sure we check the values exist and throw an error if they don't!

packages/controller/src/dbClient.ts
```diff
@@ -19,7 +19,7 @@ import { getSdk } from "./dbSDK";
 // Make this optional, so that running the controller locally during dev
 // doesn't require a connection to the db.
 const endpoint = process.env.GRAPHQL_ENDPOINT;
-const adminSecret = process.env.HASURA_GRAPHQL_ADMIN_SECRET;
+const adminSecret = process.env.HASURA_GRAPHQL_ADMIN_SECRET ?? "";
```
jgehrcke, line 22: If this is a required input parameter and if an empty string is not an allowed value I think we should terminate the process here with a proper error message, saying that the environment variable `HASURA_GRAPHQL_ADMIN_SECRET` must be set to a non-empty value.

@MatApple what do you think?

packages/controller/src/resources/app/index.ts
```diff
@@ -165,7 +167,7 @@ export function OpstraceApplicationResources(
   if (secretToCopy) {
     // use the value of the secret we want to copy
     secretValue = Buffer.from(
-      secretToCopy.data!.HASURA_ADMIN_SECRET,
+      secretToCopy.data?.HASURA_ADMIN_SECRET ?? "",
```
jgehrcke, line 170: This code section looks dangerous now, from a security point of view. Using a low-entropy secret like an empty string should be of no particular use/value -- not for testing, not in any dev environment, not in CI (and _obviously_ not in "prod").

I think security-wise I'd therefore love to be sure that `HASURA_ADMIN_SECRET` can never be an empty string.

That is, I'd love to be sure that we exit the process around the entry point when validating the config when `HASURA_ADMIN_SECRET` is not set to a string of the desired length.

After config validation, we can then rely on this invariant: `HASURA_ADMIN_SECRET` _is_ a secret of desired length (often implying desired entropy).

Deferring to @MatApple for his opinion, of course.

packages/controller/src/dbClient.ts
```diff
@@ -30,14 +30,15 @@ if (endpoint && !adminSecret) {
...
   : null;
 
 export type PromiseReturnType<T> = T extends PromiseLike<infer U> ? U : T;
-export type ClientResponse<T extends (args?: any) => {}> = PromiseReturnType<
-  ReturnType<T>
->;
+export type ClientResponse<
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  T extends (args?: any) => Record<string, any>
+> = PromiseReturnType<ReturnType<T>>;
```
MoSattler, line 42: I am curious about what this is doing. Naively I would have thought that the above is equivalent to:
```suggestion
// eslint-disable-next-line @typescript-eslint/no-explicit-any
export type ClientResponse = PromiseReturnType<Record<string, any>>;
```

packages/controller/src/dbClient.ts
MatApple: Actually we do terminate with an Error on the very next line if `adminSecret` is falsy (emtpy string is falsy) so this is fine. Good eye though.

packages/controller/src/resources/app/index.ts
MatApple: Hasura rejects an empty value for admin secret (I tested to be also be sure), so we could validate earlier in the process if we want to throw an error before the downstream components do, which there is no harm in doing, but I'd say this is not an issue since the validation does indeed happen in the most important places. Our app will also throw an error if the value is empty and so will the controller.

MatApple: Good eye though @jgehrcke :-)

MatApple (approved):
Nice work @juliachvyrova and @Jamakase 💯 🎉

Issue #73, opened 2020-12-02
ci: make sure that `addlicense ... -check` really covers all the files of interest


jgehrcke:
`addlicense` implements file discovery using a "pattern" concept -- however, that's not well documented, and the code is also a little dubious. 

Also see
https://github.com/google/addlicense/blob/a0294312aa76d31c0bd7e49083d88a2a04d9b3d1/main.go
google/addlicense#38
and 
https://github.com/opstrace/opstrace/pull/70

jgehrcke:
Related: https://github.com/opstrace/opstrace/issues/1#issue-751735293

jgehrcke:
I think that's done with #132.

PR #132, opened 2020-12-08
ci: perform license headers check across entire repo

jgehrcke:
Will also address https://github.com/opstrace/opstrace/issues/73.

todo:
- fix the problems that it finds
- remove the yarn entry points
- document

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -267,10 +267,8 @@ check-license-headers:
 		-not -path '*/kubernetes/src/custom-resources*' \
 		-type d -exec bash -c \
 		$$'echo "dir: {}"; shopt -s extglob \n addlicense -c "Opstrace, Inc." -l apache {}/@(*.ts|*.tsx|*.json|*.go|*.css)' \;
-	# Show the changes.
-	git --no-pager diff
-	# This returns non-zero when there is a diff.
-	git diff --exit-code
+	# Show the changes. This returns non-zero when there is a diff.
+	git --no-pager diff --exit-code
 
 
 .PHONY: cli-set-build-info-constants
--- a/Makefile
+++ b/Makefile
@@ -242,6 +242,28 @@ cli-test-s3-latest:
 	./opstrace --help
 
 
+.PHONY: check-license-headers
+check-license-headers:
+	@echo "does addlicense work?"
+	command -v addlicense
+	# Walk through the directory tree. For every directory (that is not
+	# ./.git), execute a bash process. Echo the current directory. Abort if no
+	# files in the directory (flat) have a matching extension. Else: run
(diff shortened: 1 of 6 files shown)


Makefile
```diff
@@ -242,6 +242,35 @@ cli-test-s3-latest:
...
+	# show a "no such file or directory" error when the extended glob does not
+	# match a single file in the current directory. That's fine. Note: we start
+	# bash so that we can rely on bash's extended glob behavior. Also note that
+	# `addlicense` does not have well-behaved recursive directory tree walking
+	# behavior (with predictable pattern matching!). That's why we use
+	# find/bash+extglob to control _that_ part. Fail the target when this
+	# technique produces file changes (show the changes, too).
+	find . -not -path '*/.git/*' \
+		-not -path '*/node_modules*' \
+		-not -path '*/kubernetes/src/custom-resources*' \
```
jgehrcke, line 267: Note: add exclude patterns here, when new ones come up.

Makefile
```diff
@@ -242,6 +242,35 @@ cli-test-s3-latest:
...
+		-type d -exec bash -c \
+		$$'echo "dir: {}"; shopt -s extglob \n addlicense -c "Opstrace, Inc." -l apache {}/@(*.ts|*.tsx|*.json|*.go|*.css)' \;
```
jgehrcke, line 269: Note: add new file extensions here, when they come up.

jgehrcke: Also note: the `{}` is where `find` fills in the path to the current direction (as part of the file system walk).

jgehrcke: Also note: launching a constant number of processes process per directory (bash and addlicense) is probably faster here than launching a constant number of processes per file, assuming that we probably have ~10 times more files than directories.

In CI, this walk passes within ~1 second. On my machine, it's a little slower (5 seconds or so).

jgehrcke: CC @triclambert not sure if you're interested in seeing this ugly piece of code, but here it is :shrug:

package.json
```diff
@@ -96,7 +96,7 @@
       "gofmt -s -w",
       "yarn golint"
     ],
-    "*.{js,css,json,ts}": [
+    "*.{js,css,json,ts,tsx}": [
```
jgehrcke, line 99: Note: this is just for the commit hook stage. Not quite sure if within a commit this really catches everything. Well. Let's leave this here.

Makefile
triclambert: I've been following along.  And steering clear.  😁   

But in all seriousness, thanks for doing this.  It's unfortunate that it's needed, but your solution looks good.

jgehrcke:
main-test-gcp failed as of #62. Merging.

PR #133, opened 2020-12-08
fix: set liveness and readiness probes

### LINKED ISSUE #131: Setup liveness probes

> Many applications running for long periods of time eventually transition to broken states, and cannot recover except by being restarted. Kubernetes provides liveness probes to detect and remedy such situations.

https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-startup-probes/

Check which deployments that have a readiness probe and are missing a liveness probe. Add the liveness probe where it's missing.

## Context

Ingester deployment for cortex and loki is missing a liveness probe. 

Fix #127

sreis:
Fix #131 

* Adds equality check for pod readiness and liveness probes. This required also setting the Kubernetes default values otherwise the desired spec and existing spec equality check would always fail.
* Adds liveness probes to the deployments/statefulsets/daemonsets that didn't have them.
* Changed `initialDelaySeconds` in cortex and loki deployments following the official helm chart values (for example, [cortex](https://github.com/cortexproject/cortex-helm-chart/blob/14ee59e7b3e8772f19a12ab16979e5143f51ae92/values.yaml#L250-L254)).

### DIFF
--- a/packages/controller/src/resources/cortex/index.ts
+++ b/packages/controller/src/resources/cortex/index.ts
@@ -1049,13 +1049,31 @@ export function CortexResources(
                       name: "grpc"
                     }
                   ],
+                  // https://github.com/cortexproject/cortex-helm-chart/blob/14ee59e7b3e8772f19a12ab16979e5143f51ae92/values.yaml#L250-L254
                   readinessProbe: {
                     httpGet: {
                       path: "/ready",
-                      port: 80 as any
+                      port: 80 as any,
+                      scheme: "HTTP"
                     },
-                    initialDelaySeconds: 15,
-                    timeoutSeconds: 1
+                    initialDelaySeconds: 45,
+                    timeoutSeconds: 1,
+                    periodSeconds: 10,
+                    successThreshold: 1,
+                    failureThreshold: 3
+                  },
+                  // https://github.com/cortexproject/cortex-helm-chart/blob/14ee59e7b3e8772f19a12ab16979e5143f51ae92/values.yaml#L245-L249
+                  livenessProbe: {
+                    httpGet: {
+                      path: "/ready",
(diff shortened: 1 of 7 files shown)


packages/controller/src/resources/cortex/index.ts
```diff
@@ -1049,13 +1049,31 @@ export function CortexResources(
...
                     },
-                    initialDelaySeconds: 15,
-                    timeoutSeconds: 1
+                    initialDelaySeconds: 45,
+                    timeoutSeconds: 1,
+                    periodSeconds: 10,
+                    successThreshold: 1,
+                    failureThreshold: 3
+                  },
+                  // https://github.com/cortexproject/cortex-helm-chart/blob/14ee59e7b3e8772f19a12ab16979e5143f51ae92/values.yaml#L245-L249
```
jgehrcke, line 1065: Thanks for adding the comment. This is great!

jgehrcke:
Great improvement. Looks like we're starting to know what we're doing around the k8s-based readiness and liveness probes. :muscle:

sreis:
@jgehrcke  We now get this in the pod events when one of the probes fail:
```
> kubectl get event --namespace loki --field-selector involvedObject.name=ingester-2 
LAST SEEN   TYPE      REASON      OBJECT           MESSAGE
6s          Warning   Unhealthy   pod/ingester-2   Liveness probe failed: HTTP probe failed with statuscode: 503
1s          Warning   Unhealthy   pod/ingester-2   Readiness probe failed: HTTP probe failed with statuscode: 503
```
Which is (or could be) the stepping stone to automate failure handling.

jgehrcke:
> We now get this in the pod events when one of the probes fail:

This is great. Thanks for sharing!!

PR #136, opened 2020-12-08
docs: address user feedback and behavior

triclambert:
See #2128.

Also, see Google Analytics.

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -3,7 +3,7 @@
 
 # Opstrace Introduction
 
- _**A secure observability platform, deployed in your own network.**_
+ _**A secure open source observability platform, deployed inside your own network.**_
 
 ## What does it do?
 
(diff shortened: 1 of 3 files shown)


triclambert:
There is a ~65% drop-off rate on the quick start page.  On 12/3, there were only 5 downloads of the CLI (via GoLinks).  So far, only 2 people have created new Auth0 users.  Looks like we have some work to do.

Commit 6563fff attempts to reduce the time-to-download by reducing the text before that action section.  I wonder if perhaps we even moved it above the fold—perhaps to the top—if at least people would download it.  After that, we could have them check for permissions and such.  (Thinking that having the CLI locally would contribute to "stickiness.")

docs/README.md
```diff
@@ -17,24 +17,26 @@ You can point your existing Prometheus or [FluentD](https://fluentd.org)/[Promta
 We also plan to support a wide variety of other APIs, such as the Datadog agent.
 
 Creating an Opstrace Cluster requires our [command-line interface](./references/cli.md), which talks directly to your cloud provider.
-It orchestrates the tedious process of setting everything up.
-After your cluster is running, our Controller inside the cluster will maintain things over time.
-All of your data resides safely (and inexpensively) in your own S3 buckets.
+It orchestrates the toilsome process of setting everything up inside your account, for example creating [a variety of resources](https://buildkite.com/organizations/opstrace/pipelines/scheduled-main-builds/builds/1322/jobs/d3de7e2d-95c0-4605-baf9-64ca03e1c9d6/artifacts/747bbf79-8dfe-447c-9d5b-68e04bbcf9d1) (which includes its own VPC to live in).
```
jgehrcke, line 20: suggestion: let's say "cloud account" instead of just "account".

jgehrcke: the build log is not public

jgehrcke: Let's add another 'advanced topics' section where we lay out the resources we create. It's a rather constant set of resources by now for AWS. Then let's link to that instead of to a build log :).

PR #143, opened 2020-12-09
makefile: move all unit tests under a single target

### LINKED ISSUE #135: Move all unit tests to a single CI pipeline step

After #87 move all the unit tests to a the `unit-tests` pipeline step.

## Context

> The preamble step would check everything builds. The unit-tests step would run the (well) the unit tests. And then the next steps would continue as today.

https://github.com/opstrace/opstrace/pull/87#discussion_r538283414

sreis:
Close #135

*  Move all unit tests under a single makefile target
* Cleanup pipeline dependencies since unit-tests already depends on docs and preamble steps.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -165,6 +165,7 @@ clean:
 
 .PHONY: run-app-unit-tests
 run-app-unit-tests:
+	@echo "--- run app unit tests"
 	CI=true yarn workspace @opstrace/app test
 
 .PHONY: cli-crashtest
@@ -175,11 +176,12 @@ cli-crashtest:
 	# interested in confirming whether grep as the last command in the pipeline
 	# exits non-zero (did not find match -> error), or with exit code 0 (did
 	# find match -> test passed).
+	@echo "--- make cli-crashtest"
 	./build/bin/opstrace crashtest || exit 0 # so that the output is visible in build log
 	set +o pipefail && ./build/bin/opstrace crashtest 2>&1 | grep 'cli/src/index.ts:'
 
 
-lint-codebase.js:
+lint-codebase.js: yarn-frozen-lockfile
 	yarn run lint
 
 lint-codebase.go:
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -560,8 +562,30 @@ yarn.lock: package.json $(PACKAGE_JSON) node_modules
...
+	@echo "--- run opstrace CLI tests (cli-tests-pre-cluster.sh)"
+	CHECKOUT_VERSION_STRING=${CHECKOUT_VERSION_STRING} source ci/test-cli/cli-tests-pre-cluster.sh
+
+.PHONY: go-unit-tests
+go-unit-tests:
+	@echo "--- run go unit tests"
+	cd go && make unit-tests
+
+.PHONY: js-unit-tests
+ts-unit-tests:
```
jgehrcke, line 587: js/ts typo I think :)

ci/preamble.sh
```diff
@@ -36,19 +36,15 @@ bash ci/check-if-docs-pr.sh && exit 0
 echo "--- detect missing license headers"
 make check-license-headers
 
+echo "--- lint codebase: quick feedback"
+make lint-codebase
```
jgehrcke, line 40: The original reason to put this after `make cli-tsc` was that linting errors were often boring and not so insightful, and tsc errors were more exciting, authoritative, and I wanted to see them _even if_ linting failed. Maybe we can still achieve this later. I am ok with moving this :) Especially now that linting feedback is getting better and better.

jgehrcke: and the implicit `yarn install` within `make cli-tsc` now probably needs to be pulled out / up so that `make lint-codebase` has all its dependencies.

sreis: > Maybe we can still achieve this later. I am ok with moving this :) Especially now that linting feedback is getting better and better.

Figured we wanted to fail as fast as possible.

> and the implicit yarn install within make cli-tsc now probably needs to be pulled out / up so that make lint-codebase has all its dependencies.

Looks like it.... forgot to test in a clean checkout.

jgehrcke (approved):

jgehrcke:
main-test-gcp failed as of #62. I looked at the new job's output -- this is great! Let's land this.

PR #149, opened 2020-12-10
feat(aws): cortex/loki bucket lifecycle rule for data retention

sreis:
Adds a lifecycle rule to the S3 buckets used by loki and cortex to ensure data
is deleted periodically.

Things to know about when S3 data is deleted:

"Amazon S3 rounds the transition or expiration date of an object to midnight
UTC the next day. For example, if you created an object on 1/1/2020 10:30 UTC,
and you set the lifecycle rule to transition the object after 3 days, then the
transition date of the object is 1/5/2020 00:00 UTC. Before you check whether a
lifecycle rule has been satisfied, be sure to verify that enough time has
elapsed."

### DIFF
--- a/lib/aws/src/bucket.ts
+++ b/lib/aws/src/bucket.ts
@@ -24,11 +24,18 @@ import { AWSResource } from "./resource";
 
 export class S3BucketRes extends AWSResource<true> {
   private bname: string;
-
+  // the array of lifecycle rules to apply to this bucket, we don't handle the
+  // deletion since that is taken cared of for us when the bucket is deleted
+  private lifecycleRules: AWS.S3.LifecycleRules = [];
   // overridden in constructor, see below.
   protected rname = "";
 
-  constructor(opstraceClusterName: string, bucketName: string) {
+  constructor(
+    opstraceClusterName: string,
+    bucketName: string,
+    retentionPeriod: number = 0,
+    tenants: string[] = []
+  ) {
     super(opstraceClusterName);
 
     this.changeRname(`s3 bucket ${bucketName}`);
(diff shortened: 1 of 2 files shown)


sreis:
Converted to draft while I work on a test but want to get feedback as soon as possible in case anything is broken.

sreis:
The blocks are not written to S3 in the time the e2e tests run. So my idea to check if the files in S3 had an expiration set didn't work out.

The [lifecyle rules have been applied in the S3 buckets](https://s3.console.aws.amazon.com/s3/management/ci-cortex/lifecycle?region=us-west-2) where the CI builds metrics and logs are stored. It has been running without any issues for some time now.

Screenshot from 2020-12-10 16-34-10

Graph of storage used in S3 by cortex.  It was growing at a steady pace until I applied the lifecycle rule.


Screenshot from 2020-12-10 16-36-56
Graph of storage used in S3 by loki where the lifecycle rule was applied later (around December 1st). Afterwards it released older data.

jgehrcke:
Thanks!

What about consistency between the index that Loki and Cortex use, and the actual data in the cloud storage?

I am hoping that we can have loki/cortex delete objects from index and bulk storage in a consistent fashion, based on retention settings we configure them with :). 

For the record: when we delete data "underneath" those systems then certain queries will crash when index and cloud storage content got out of sync, right?

sreis:
> What about consistency between the index that Loki and Cortex use, and the actual data in the cloud storage?
> For the record: when we delete data "underneath" those systems then certain queries will crash when index and cloud storage content got out of sync, right?

@jgehrcke  The [compactor docs](https://cortexmetrics.io/docs/blocks-storage/compactor/#soft-and-hard-blocks-deletion) explain it better than I can:

> This soft deletion mechanism is used to give enough time to queriers and store-gateways to discover the new compacted blocks before the old source blocks are deleted. If source blocks would be immediately hard deleted by the compactor, some queries involving the compacted blocks may fail until the queriers and store-gateways haven’t rescanned the bucket and found both deleted source blocks and the new compacted ones.

PR #154, opened 2020-12-10
docs: remove aws/gcp tabs from cluster config ref

jgehrcke:
The tabs should be used here: both, `aws` and `gcp` are high-level optional configuration parameters that should be enumerated like all other high-level optional configuration parameters.

In addition to that, the tabbing leads to problems:
Screenshot from 2020-12-10 22-13-00

- notes and warnings are mixed
- the tab headings don't look great because of lower case letters (but lower case letters are correct, because that's how the parameters are named)


triclambert (approved):

jgehrcke:
:muscle:

PR #169, opened 2020-12-14
chore: deploy loki query-frontend

sreis:
https://grafana.com/docs/loki/latest/configuration/query-frontend/#frontend-service

> (...) A common bottleneck (in Loki) is on the read path: queries that executed effortlessly on small data sets may churn to a halt on larger ones. Sometimes we can solve this with more queriers. However, that doesn’t help when our queries are too large for a single querier to execute. Then we need the query frontend

> One of the most important functions of the query frontend is the ability to split larger queries into smaller ones, execute them in parallel, and stitch the results back together.


sreis:
I will work on it in the background.

PR #170, opened 2020-12-14
Show keybindings for commands

### LINKED ISSUE #155: Show keybinding for commands that have a key binding in the command picker
When a command in the command palette has a keybinding, show the keybinding on the right.

We should detect the platform using `navigator.platform` so we can show the correct mod key for the platform i.e. `⌘` for mac, `^` for others.

Today:

Example of what is desired:

juliachvyrova:
Close #155


MatApple (commented):
🎉 Very cool!

MatApple (approved):

PR #178, opened 2020-12-16
ci: acquire lock only in linux systems

### LINKED ISSUE #177: ci/dns_cache.sh: line 55: flock: command not found
**Describe the bug**

`make test-remote` and `make ci-*` targets do not run on macOS by default without an additional dependency: https://github.com/discoteq/flock

sreis:
Fix #177


Jamakase (approved):
Awesome!

PR #185, opened 2020-12-17
docs: refresh algolia index and re-deploy website when docs change

PatrickHeneise:
Add a GitHub Action to re-index /docs with Algolia and re-deploy the website with the latest changes on `main`

# Have you...

### DIFF
--- a/.github/workflows/docs-sync.yml
+++ b/.github/workflows/docs-sync.yml
@@ -1,0 +1,27 @@
+name: Sync Algolia
+
+on:
+  push:
+    branches:
+      - "main"
+    paths:
+      - "docs/**"
+
+jobs:
+  algolia:
+    runs-on: ubuntu-latest
+    name: Algolia Sync
+    steps:
+      - uses: actions/checkout@v2
+
+      - uses: opstrace/algolia-docs-sync@v1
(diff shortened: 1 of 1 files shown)


jgehrcke:
Thanks! Is this deploying to a staging environment (or something like that) or to opstrace.com/docs?

PatrickHeneise:
Doesn't need deployment anywhere. It runs when there are changes in the `docs` folder on `main`

jgehrcke:
> This refreshes opstrace.com/docs 

Okay, Yolo! :) :rocket: 

Would be cool to have a way to see how the result will look like (render) on the live site _before merging_. Is there?

PatrickHeneise:
Little tricky, but could be possible. I'll open a ticket and investigate separately.

triclambert:
> Would be cool to have a way to see how the result will look like (render) on the live site before merging. Is there?

There is a way to do this that involves a little bit of work.  Tl;dr: in a local clone of the website, you update it to point to your branch.  However, hopefully, GitHub rendering is good enough since we tried to match that very closely.

We'll probably come up with something better in the future.

triclambert (approved):

PR #186, opened 2020-12-17
chore: await axios.request()

jgehrcke:
We want to move to `got` and use established patterns, for good reasons :)
Until then...well.

Quick improvement for https://github.com/opstrace/opstrace/issues/187.

Also an explanation.

For more in-depth considerations, see commit msg.

I'd like to stress that this topic has bothered us before, by a _lot_. We initiated this conversation: https://github.com/eslint/eslint/issues/13567 -- which was super insightful for me.

### DIFF
--- a/lib/dns/src/opstrace.ts
+++ b/lib/dns/src/opstrace.ts
@@ -17,7 +17,7 @@
 import * as fs from "fs";
 
 import qs from "qs";
-import axios, { AxiosRequestConfig } from "axios";
+import axios, { AxiosResponse, AxiosRequestConfig } from "axios";
 import open from "open";
 
 import { log } from "@opstrace/utils";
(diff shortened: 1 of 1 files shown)


lib/dns/src/opstrace.ts
```diff
@@ -144,11 +144,11 @@ export class DNSClient {
       url,
       headers: this.headers
     };
-    const r = await axios.request(getRequest);
```
jgehrcke, line 147: Here we _did_ wait for the response, and likewise any corresponding axios error would be thrown _in here_.

lib/dns/src/opstrace.ts
```diff
@@ -158,10 +158,11 @@ export class DNSClient {
         clustername
       }
     };
-    return axios.request(deleteRequest);
```
jgehrcke, line 161: Here we _did not_ wait for the response, but moved on with the program.

Likewise any corresponding axios error corresponding with this dangling task would be thrown in the context of things that are difficult see through.

Potentially, we didn't even wait for the HTTP response to arrive before terminating the program  -- possible, when DNS deletion happens towards the end of the `destroy` operation`. Which it does:

https://github.com/opstrace/opstrace/blob/eb32d8ef4e563d9964388a82d8fc9bb8c187eac9/packages/uninstaller/src/gcp.ts#L108

This code change makes it _much_ easier to reason about where/when an unexpected HTTP response (non-2xx in this case I guess, axios default) is expected to throw an error.

jgehrcke:
GCP run affected by CI instability: https://github.com/opstrace/opstrace/issues/188

PR #190, opened 2020-12-17
Fix lib lint

juliachvyrova:
Part #51

- `dns/aws.ts` HostedZoneId was used as `zone.name` what will happen if its undefined or empty string like it was before? Or is it possbile at all?
- `dns/index.ts` Is it actually safe to assume that if we have no name-servers we can skip AddNameservers call?
- Ignore `lib/gcp/network` and `lib/gcp/subnetworks` there are a lot of warnings with any type. Because we use library without types
- `lib/kubernetes` will be fixed in another PR (there are too much errors and warnings)

### DIFF
--- a/lib/gcp/src/subnetworks.ts
+++ b/lib/gcp/src/subnetworks.ts
@@ -26,7 +26,7 @@ import { log, SECOND } from "@opstrace/utils";
 const doesSubNetworkExist = async (
   client: any,
   { name }: { name: string }
-): Promise<any> => {
+): Promise<unknown> => {
   return new Promise((resolve, reject) => {
     client
       .getSubnetworks()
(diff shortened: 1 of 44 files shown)


lib/gcp/src/subnetworks.ts
```diff
@@ -22,11 +22,15 @@ import { ApiError as GCPApiError } from "@google-cloud/common";
 import Compute from "@google-cloud/compute";
 import { log, SECOND } from "@opstrace/utils";
 
+// Currently as @google-cloud/compute does not provide typings - it doesn't
+// make much sense to fix all lint errors for any
+/* eslint-disable @typescript-eslint/no-explicit-any */
```
jgehrcke, line 27: :+1:

jgehrcke (approved):
Thank you! I skimmed over the changes -- they look good. This is not a thorough review trying to catch new bugs. Instead, I think let's trust CI and merge this -- and see we'll see what breaks (if something breaks) soon.

PR #217, opened 2020-12-26
Save sidebar width in DisplayService

### LINKED ISSUE #148: Save Sidebar width to DisplayService
The width for the sidebar should be stored in the DisplayService (just like the Sidebar's visibility) so that a user can drag the sidebar to the width they like and it will remain there while they navigate the app. This should only persist for the browsing session (when the tab is closed and reopened, the width will be reset to the default value)

juliachvyrova:
Closes #148


MatApple:
GCP failed due to reached quota. All other tests pass so merging.

MatApple (approved):

PR #232, opened 2021-01-04
yarn lockfile update, dev dependency bumps, use TS 4.1.x everywhere

jgehrcke:


yarn.lock
```diff
@@ -2112,50 +2112,6 @@
...
-  version "0.4.1"
-  resolved "https://registry.yarnpkg.com/@cypress/listr-verbose-renderer/-/listr-verbose-renderer-0.4.1.tgz#a77492f4b11dcc7c446a34b3e28721afd33c642a"
-  integrity sha1-p3SS9LEdzHxEajSz4ochr9M8ZCo=
-  dependencies:
-    chalk "^1.1.3"
-    cli-cursor "^1.0.2"
-    date-fns "^1.27.2"
-    figures "^1.7.0"
-
-"@cypress/request@^2.88.5":
```
jgehrcke, line 2125: this was still part of the cypress removal

package.json
```diff
@@ -24,31 +24,31 @@
...
     "redux-devtools-extension": "^2.13.8",
     "rimraf": "^3.0.0",
     "source-map-support": "^0.5.19",
-    "ts-loader": "^8.0.11",
-    "ts-node": "^9.0.0",
-    "typescript": "4.0.5",
-    "wsrun": "^5.2.0"
+    "ts-loader": "^8.0.13",
+    "ts-node": "^9.1.1",
+    "typescript": "4.1.3",
```
jgehrcke, line 50: had some fragmentation here, parts of the code base already asked for this. should aim for a common typescript version across code base.

lib/aws/src/policy.ts
```diff
@@ -59,7 +59,7 @@ export const deletePolicy = ({
   PolicyArn
 }: {
   PolicyArn: string;
-}): Promise<IAM.Policy | undefined> => {
+}): Promise<void> => {
```
jgehrcke, line 62: `resolve()` is what we do below. The type signature didn't reflect that, right? (why did this compile?)

Issue #53, opened 2020-12-01
cli/(un)installer: properly handle a 401 response upon dns service login


jgehrcke:
When the DNS service login fails with a permanent error (which a 401 error  is supposed to be: a non-retryable error indicating that credentials are bad) then we have to handle that situation properly.

What's currently happening is that the error is unhandled, emits a stack trace, and consumes a high-level `create` attempt -- until exhausted:

```
[2020-12-01T13:54:24Z] 2020-12-01T13:54:24.270Z info: setting up DNS
[2020-12-01T13:54:24Z] 2020-12-01T13:54:24.270Z debug: DNSClient.GetAll()
[2020-12-01T13:54:24Z] 2020-12-01T13:54:24.342Z error: error during cluster creation (attempt 3):
[2020-12-01T13:54:24Z] Error: Request failed with status code 401
[2020-12-01T13:54:24Z]     at createError (/snapshot/build/node_modules/axios/lib/core/createError.js:16:15)
[2020-12-01T13:54:24Z]     at settle (/snapshot/build/node_modules/axios/lib/core/settle.js:17:12)
[2020-12-01T13:54:24Z]     at IncomingMessage.handleStreamEnd (/snapshot/build/node_modules/axios/lib/adapters/http.js:236:11)
[2020-12-01T13:54:24Z]     at IncomingMessage.emit (events.js:327:22)
[2020-12-01T13:54:24Z]     at IncomingMessage.EventEmitter.emit (domain.js:485:12)
[2020-12-01T13:54:24Z]     at endReadableNT (_stream_readable.js:1224:12)
[2020-12-01T13:54:24Z]     at processTicksAndRejections (internal/process/task_queues.js:84:21) {
[2020-12-01T13:54:24Z]   config: [Object],
[2020-12-01T13:54:24Z]   request: [ClientRequest],
[2020-12-01T13:54:24Z]   response: [Object],
[2020-12-01T13:54:24Z]   isAxiosError: true,
[2020-12-01T13:54:24Z]   toJSON: [Function (anonymous)]
[2020-12-01T13:54:24Z] }
[2020-12-01T13:54:24Z] 2020-12-01T13:54:24.342Z error: JSON representation of err: {
[2020-12-01T13:54:24Z]   "message": "Request failed with status code 401",
[2020-12-01T13:54:24Z]   "name": "Error",
[...]

PR #238, opened 2021-01-05
Auth0 device code login (for DNS service authentication): consolidate (no axios, clearer polling loop, credential refresh, ...)

jgehrcke:
Among others, this resolves issue #53.

```
When the DNS service responds to an HTTP
request with a 401 HTTP response then
this means that the authentication proof
(set of credentials) presented in the request
was invalid.

A common reason for that is authentication
proof / credential expiry. Therefore, a valuable
course of action is to obtain a fresh set of
credentials, by going through another login
process.
```

This also removes the last bits of usage of the Axios HTTP client in the installer/uninstaller (replaced with got) -- and therefore allows for closing #29.

### DIFF
--- a/lib/dns/src/opstrace.ts
+++ b/lib/dns/src/opstrace.ts
@@ -37,7 +37,11 @@ import {
 
 const ACCESS_TOKEN_FILE_PATH = "./access.jwt";
 const ID_TOKEN_FILE_PATH = "./id.jwt";
+
+// Use e.g. https://httpbin.org/status/500
+// for testing behavior upon 5xx.
 const DNS_SERVICE_URL = "https://dns-api.opstrace.net/dns/";
+
 const OIDC_ISSUER = "https://opstrace-dev.us.auth0.com";
 const OIDC_CLIENT_ID = "fT9EPILybLT44hQl2xE7hK0eTuH1sb21";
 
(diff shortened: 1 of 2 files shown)


lib/dns/src/opstrace.ts
```diff
@@ -30,44 +30,204 @@ import {
...
+  const descr = data["error_description"];
+  if (descr !== undefined) {
+    log.error("verification failed: %s", descr);
+    // Note(JP): the legacy behavior here was to keep polling. That does not
+    // seem to be the right approach. This should be looked at as a permanent
+    // DNS service login error. Leave the program.
+    die("DNS service login failed");
+  }
+
+  // TODO: think about which cases to handle how. Don't retry everything.
```
jgehrcke, line 114: But for now, we keep polling :). Might be fine, we'll see.

lib/dns/src/opstrace.ts
```diff
@@ -207,43 +298,62 @@ export class DNSClient {
...
+          }
+
+          if (e.response?.statusCode === 401) {
+            log.info("perform another login to refresh authentication state");
+            this.clearAuthnState();
+            await this.login();
+
+            log.info(
+              "retry the request that previously failed with a 401 response"
+            );
```
jgehrcke, line 344: Note: if that were to fail again, this indicates a permanent error. Well.

jgehrcke:
gcp ci failed as of https://github.com/opstrace/opstrace/issues/62

PR #242, opened 2021-01-06
chore: bump axios to 0.21.1 to fix CVE

sreis:
Fix https://github.com/advisories/GHSA-4w2v-q235-vp99

### DIFF
--- a/packages/app/package.json
+++ b/packages/app/package.json
@@ -106,6 +106,7 @@
     "react-virtualized-auto-sizer": "^1.0.2",
     "react-window": "^1.8.5",
     "redis": "^3.0.2",
+    "redux-saga": "^1.1.3",
     "reselect": "^4.0.0",
     "resolve": "1.15.0",
     "retry-axios": "^2.4.0",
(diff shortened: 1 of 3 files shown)


packages/app/package.json
```diff
@@ -106,6 +106,7 @@
     "react-virtualized-auto-sizer": "^1.0.2",
     "react-window": "^1.8.5",
     "redis": "^3.0.2",
+    "redux-saga": "^1.1.3",
```
jgehrcke, line 109: Oh yeah, probably as of my changes yesterday, hidden by app image not having rebuilt.

sreis: Seems like it https://github.com/opstrace/opstrace/commit/f4e890bee17cb0d071eddf936a2d9525fe1496f0#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519L55-L56 was this.

Dependencies management is hard... I think we can move it here safely. Let's wait for the builds.

sreis: > I think we can move it here safely 

I guess not...

```
[2021-01-06T14:02:47Z]   Try `npm i --save-dev @types/got` if it exists or add a new declaration (.d.ts) file containing `declare module 'got';`
```
https://buildkite.com/opstrace/prs/builds/3365#133eb72f-5b47-4183-a989-234808e4f611/1061-1170

Issue #196, opened 2020-12-21
ci instability: loki query: Timeout awaiting 'request' for 60000ms


jgehrcke:
```
[2020-12-20T12:48:45Z]   17 passing (1m)
[2020-12-20T12:48:45Z]   1 failing
[2020-12-20T12:48:45Z]
[2020-12-20T12:48:45Z]   1) Loki API test suite
[2020-12-20T12:48:45Z]        short dummystream insert, validate via query:
[2020-12-20T12:48:45Z]      RequestError: Timeout awaiting 'request' for 60000ms
```
https://buildkite.com/opstrace/scheduled-main-builds/builds/1402#0de7d3fe-da0f-4226-a317-bfc5bf2d5421/2801-4864

sreis:
https://buildkite.com/opstrace/scheduled-main-builds/builds/1456#a2ba52c3-41b3-45f8-8e81-aee08a209782/1965-4034

```
[2020-12-29T12:52:08Z]   1) Loki API test suite
[2020-12-29T12:52:08Z]        short dummystream insert, validate via query:
[2020-12-29T12:52:08Z]      RequestError: Timeout awaiting 'request' for 60000ms
```

sreis:
https://buildkite.com/opstrace/prs/builds/3385#8f175cff-7d2e-4610-8ec1-28c61c644e9e/1744-3817

```
[2021-01-12T14:13:28Z]   1) system logs test suite
[2021-01-12T14:13:28Z]        check loki ingester logs:
[2021-01-12T14:13:28Z]      RequestError: Timeout awaiting 'request' for 60000ms
```

triclambert:
https://buildkite.com/opstrace/scheduled-main-builds/builds/1636#6a374349-e251-43b0-aa08-885d199f873c/3453-5510

```
[2021-01-27T12:49:14Z]   18 passing (1m)
[2021-01-27T12:49:14Z]   1 failing
[2021-01-27T12:49:14Z]
[2021-01-27T12:49:14Z]   1) Loki API test suite
[2021-01-27T12:49:14Z]        short dummystream insert, validate via query:
[2021-01-27T12:49:14Z]      RequestError: Timeout awaiting 'request' for 60000ms
```

jgehrcke:
```
[2021-02-06T00:41:13Z]   18 passing (2m)
[2021-02-06T00:41:13Z]   1 failing
[2021-02-06T00:41:13Z]
[...]

PR #256, opened 2021-01-12
controller: allow pvc volume expansion

sreis:
This can be used in an emergency. For example, when the ingester volume reaches the default limit, this triggers a crash loop of that pod because it can't write to a full volume. Allowing the PVC to be resized makes it easier to address the issue proactively. 

[Kubernetes docs](https://kubernetes.io/docs/concepts/storage/persistent-volumes/)
[GCP docs](https://cloud.google.com/kubernetes-engine/docs/how-to/persistent-volumes/volume-expansion#using_volume_expansion) 
[AWS docs](https://kubernetes.io/docs/concepts/storage/storage-classes/#aws-ebs)

### DIFF
--- a/packages/controller/src/resources/storage/aws.ts
+++ b/packages/controller/src/resources/storage/aws.ts
@@ -36,7 +36,8 @@ export function StorageResources(
           type: "gp2"
         },
         provisioner: "kubernetes.io/aws-ebs",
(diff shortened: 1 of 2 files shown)


sreis:
AWS failed with https://github.com/opstrace/opstrace/issues/196#issuecomment-758708411

MatApple:
Makes sense!

PR #261, opened 2021-01-13
chore: reduce default instance size everywhere

triclambert:


packages/cli/src/schemas.ts
```diff
@@ -28,7 +28,7 @@ const CONTROLLER_IMAGE_DEFAULT = `opstrace/controller:${BUILD_INFO.VERSION_STRIN
 // `us-west-2a` encodes availability zone "a" within that region.
 export const infraConfigSchemaAWS = yup
   .object({
-    instance_type: yup.string().default("t3.2xlarge"),
+    instance_type: yup.string().default("t3.xlarge"),
```
jgehrcke, line 31: If we make this change a documentation upadte has to be part of this :)

https://github.com/opstrace/opstrace/blob/3f405cd4baa709c5d624d8966b8e2820b28ea37f/docs/references/cluster-configuration.md#aws

jgehrcke: edit: oh, the default values are not yet explicitly documented. OK.

jgehrcke:
context: https://github.com/opstrace/opstrace/issues/262

sreis:
Pushed a commit to update the docs and stopped the build. Although not explicit, might as well keep the config example in sync with the defaults.

jgehrcke:
> Although not explicit, might as well keep the config example in sync with the defaults.

yes, further down the line we should document (as with `gcp`) that what's shown _are_ the defaults, and even further down the line this kind of thing should be automated :)

PR #265, opened 2021-01-17
chore: update dev credentials

triclambert:


jgehrcke (approved):
:rocket:

Makefile
```diff
@@ -104,7 +104,6 @@ cli: cli-tsc cli-pkg
 .PHONY: fetch-secrets
 fetch-secrets:
 	@# Fetch secrets, expected to be done before every cloud deployment.
-	aws s3 cp "s3://buildkite-managedsecretsbucket-100ljuov8ugv2/vast-pad-240918-1445f1830b0e.json" secrets/gcp-credentials.json
```
jgehrcke, line 107: I understand that part!

Makefile
```diff
@@ -114,6 +113,7 @@ fetch-secrets:
 	--include "ci.id_rsa.pub" \
 	--include "opstrace-collection-cluster-authtoken-secrets.yaml" \
 	--include "dns-service-login-for-ci.json" \
+	--include "gcp-svc-acc-dev-dns-service.json" \
```
jgehrcke, line 116: But that part is unclear :-). Why is that needed? We don't use this Makefile in the context of the DNS service, right?

jgehrcke (commented):
Left a question about why we would want to adde `--include "gcp-svc-acc-dev-dns-service.json"` as part of this Makefile's `fetch-secrets` target.

triclambert:
We can certainly omit it.  However, it seems like `fetch-secrets` is the canonical way to get all of the credentials we use.  For example, documentation from @sreis on where to find the credentials for it relies on these being present here in the `secrets` directory.

sreis (approved):

sreis:
>  @sreis on where to find the credentials for it relies on these being present here in the secrets directory.

Created a ticket to add a similar target in dns-service. I'm ok merging this as is for now.

jgehrcke (approved):

PR #271, opened 2021-01-20
Grafana: use postgres db

spahl:
Before this PR grafana was still using sqlite in a volume.
This PR takes advantage of the aurora and cloudsql databases to fix that.

### DIFF
--- a/packages/controller/src/resources/monitoring/tenants/grafana.ts
+++ b/packages/controller/src/resources/monitoring/tenants/grafana.ts
@@ -22,7 +22,6 @@ import {
   Deployment,
   Service,
   ServiceAccount,
-  PersistentVolumeClaim,
   V1ServicemonitorResource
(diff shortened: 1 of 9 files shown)


jgehrcke:
Cool!

packages/controller/src/resources/monitoring/tenants/grafana.ts
```diff
@@ -172,6 +169,31 @@ export function GrafanaResources(
               }
             },
             spec: {
+              initContainers: [
+                {
+                  image: "tmaier/postgresql-client",
```
sreis, line 174: This should be moved to docker-images.json and also needs an image tag.

spahl: Thanks. https://github.com/opstrace/opstrace/pull/278

PR #278, opened 2021-01-20
controller: pin postgresql-client image

spahl:


sreis:
It seems [the check](https://github.com/opstrace/opstrace/blob/main/ci/check-deployed-docker-images.sh)  doesn't handle init container images.

sreis:
A fix was pushed so I'll rebase with main and trigger new tests.

sreis (approved):

PR #304, opened 2021-01-26
cortex: introduce ruler & alertmanager

spahl:


spahl:
Alertmanager UI is not exposed yet. It will be done in a separate PR, no reason to hold this up.

go/cmd/loki/main.go
```diff
@@ -71,7 +71,8 @@ func main() {
 	log.Infof("tenant name: %s", tenantName)
 	log.Infof("API authentication enabled: %v", !disableAPIAuthentication)
 
-	reverseProxy := middleware.NewReverseProxy(tenantName, lokiqurl, lokidurl, disableAPIAuthentication)
+	reverseProxy := middleware.NewReverseProxy(tenantName, lokiqurl, lokidurl,
+		lokiqurl, lokiqurl, disableAPIAuthentication) // dirty: pass the querier since there is no ruler nor alertmanager
```
jgehrcke, line 75: Haha. Dirty indeed. :-)

This is where we need to have two different abstractions for the Loki and Cortex API proxies. 

Maybe they can inherit from each other (have a common base).

spahl: I'm going to be moving all the ruler/alert proxy to @nickbp api/config code today after chatting with him and @MatApple. It makes more sense for users than having it under cortex. Will clean this at the same time.

PR #306, opened 2021-01-26
ci: add test upgrade pipeline to main branch

### LINKED ISSUE #267: task: Test upgrades
## Context

Going forward, we'll need to test upgrades before a new release. For now, we will simplify this story and set the initial goal of gathering more data for the upgrade process.

In this scenario, an upgrade is defined as issuing a `create` command for a running cluster with a new CLI version.

We run daily tests in CI for the main branch. These tests build the CLI and push the artifact to an S3 bucket. We will start by testing that we can upgrade between the last two available CLI versions. 

## Technical Solution

Create a buildkite pipeline that starts a job that:

1. Downloads the two latest CLI versions
2. Creates a cluster with the oldest version of the two
3. Runs test-remote against the cluster 
4. Runs an upgrade by issuing the `create` command with the latest available CLI version
5. Runs test-remote against the cluster

sreis:
Close #267

### TODO

I'll follow up with tickets for:

- How to manage flaky tests in main branch when the test remote fails before running the upgrade?
- Schedule upgrade tests that run after main branch tests

### DIFF
--- a/.buildkite/test-upgrade-pipeline.yml
+++ b/.buildkite/test-upgrade-pipeline.yml
@@ -1,0 +1,11 @@
+steps:
+  - label: "🔨 main upgrade test (AWS)"
+    key: "maintest-upgrade-aws"
+    env:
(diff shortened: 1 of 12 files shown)


sreis:
I'll create tickets for follow up work.

PR #316, opened 2021-02-01
ci: link check

PatrickHeneise:
Run the [markdown link check](https://github.com/marketplace/actions/markdown-link-check) action on each Pull Request that modifies docs.

# Have you...

### DIFF
--- a/.github/workflows/link-check.yml
+++ b/.github/workflows/link-check.yml
@@ -7,7 +7,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@master
-      - uses: gaurav-nelson/github-action-markdown-link-check@v1.0.1
+        uses: gaurav-nelson/github-action-markdown-link-check@1.0.12
         with:
           use-quiet-mode: "yes"
           use-verbose-mode: "yes"
--- a/.github/workflows/link-check.yml
+++ b/.github/workflows/link-check.yml
@@ -6,8 +6,9 @@ jobs:
   markdown-link-check:
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@master
-        uses: gaurav-nelson/github-action-markdown-link-check@1.0.12
+      - uses: actions/checkout@v2
+
+      - uses: gaurav-nelson/github-action-markdown-link-check@1.0.12
         with:
           use-quiet-mode: "yes"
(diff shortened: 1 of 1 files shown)


.github/workflows/link-check.yml
```diff
@@ -0,0 +1,17 @@
...
+jobs:
+  markdown-link-check:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+
+      - uses: gaurav-nelson/github-action-markdown-link-check@1.0.12
+        with:
+          use-quiet-mode: "yes"
+          use-verbose-mode: "yes"
```
jgehrcke, line 14: Aren't these two settings a little contradictory? :)

The most detailed level of output seems to doable with quiet mode off, and verbose mode on. Well. :) Maybe we want to stick to the defaults here?

jgehrcke: for the record: https://github.com/gaurav-nelson/github-action-markdown-link-check#custom-variables

jgehrcke:
That's a great addition, thanks for this @PatrickHeneise! I've asked a stupid question w.r.t. verbosity, but we can tweak that later.

For the record, log output from the run performed in this PR:

```
NOTE: See https://github.com/tcort/markdown-link-check#config-file-format to know more about
customizing markdown-link-check by using a configuration file.
USE_QUIET_MODE: yes
USE_VERBOSE_MODE: yes
FOLDER_PATH: docs
MAX_DEPTH: -1
CHECK_MODIFIED_FILES: yes
FILE_EXTENSION: .md
FILE_PATH: 
BASE_BRANCH: main
From https://github.com/opstrace/opstrace
 * branch            main       -> FETCH_HEAD
 * [new branch]      main       -> origin/main
All good!
```

It would maybe be good, initially, to increase verbosity. I think it makes sense that no file was actually checked here, because of `CHECK_MODIFIED_FILES: yes`. But I'm not sure, based on the log output (something like "0 files(s) checked" would be helpful).

In any case, we can go ahead with merging this and iterate on this as we learn.

.github/workflows/link-check.yml
PatrickHeneise: those sound confusing indeed, this is my takeaway:

- quiet mode: don't show 2xx's
- verbose-mode: show http code for errors (404, 500, ...)

PR #335, opened 2021-02-06
cli: require user to accept privacy policy

triclambert:


packages/cli/src/create.ts
```diff
@@ -127,16 +127,23 @@ export async function create(): Promise<void> {
...
+        "By selecting 'y' you agree to our terms of service: %s\n\n",
+      url,
+      tos
     );
-    await util.promptForProceed();
   }
 
-  // if (ccfg.cloud_provider === "gcp") {
-  // }
+  if (ccfg.cloud_provider === "gcp") {
```
triclambert, line 140: I added this here in order to have all messages related to the prompt in a single message to make it as easy-to-read as possible.

packages/cli/src/create.ts
```diff
@@ -125,18 +125,26 @@ export async function create(): Promise<void> {
 }
 
 async function promptForResourceCreation(ccfg: NewRenderedClusterConfigType) {
+  const tos = "https://go.opstrace.com/terms-of-service";
```
jgehrcke, line 128: `"go/terms-of-service" was not found`

triclambert: Yes, I converted to draft until we have the text.

triclambert:
After reconsidering this, I don't believe it is strictly necessary.  Let's abandon it.

PR #348, opened 2021-02-10
DD API: start adding system dashboard, improve logging, enable Prom middleware

jgehrcke:


jgehrcke:
Screenshot from 2021-02-10 13-15-03

Prom middleware in the API implementation wasn't enabled yet (changing this in this PR) -- which is why the two panels on the left are still empty. I prepared them w/o being able to plot. They might work or not work after this PR. This is just a starting point.

PR #349, opened 2021-02-10
make check-license-headers: set year range (and update notices across repo)

jgehrcke:
We started having fragmentation of copyright notices across files. Some mentioning 2020, others 2021.

I think we want a consistent year range in the copyright notice: the same in all files.

This is just a proposal. Please review @triclambert!

See commit messages for helpful cross-repo replacement commands.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -287,7 +287,7 @@ check-license-headers:
 		-not -path '*/.npm/*' \
 		-not -path './build/*' \
 		-type d -exec bash -c \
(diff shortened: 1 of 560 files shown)


triclambert:
I appreciate the simplicity of this approach.  However, as discussed offline, this is a business decision that we will defer for now.  We will continue to follow [Google's approach](https://opensource.google/docs/copyright/) for including a copyright notice in each source file.  

One other approach we will consider is the [CNCF's method](https://github.com/cncf/foundation/blob/master/copyright-notices.md) for simplifying the source file header by omitting each individual copyright holder.  You can read their reasoning [here](https://github.com/cncf/foundation/blob/master/copyright-notices.md#why-not-list-every-copyright-holder).

PR #351, opened 2021-02-11
config: Rename 'graphql' API to 'config' API, accept tenant via bearer

nickbp:
rename service: The prior `graphql` name was quickly confusing for everyone involved because it doesn't actually serve graphql. Instead it's just an implementation detail that it acts as a frontend to the graphql service. Since the service handles creating, updating, and deleting credentials/exporters entries in graphql, and might be extended to support changing other settings in the future, it felt like `config` would be a reasonable name for the service, at least for now.

change tenantname handling: The config service previously was "locked" on startup to a single tenant via CLI argument, with instances running in each tenant namespace. However, testing quickly showed that the tenant namespaces don't have access to the hasura/graphql secret. To resolve this (and avoid needing to copy the hasura secret into per-tenant namespaces), a few internal changes are made to the config service:
- The `--tenantname` argument is removed and the service is no longer "locked" to a single tenant on startup
- The service instead reads the tenant name from the (validated) bearer token in HTTP requests (or an `X-Tenant` header if `-disable-api-authn` is specified for testing)
- The `middleware` package has an API call added for retrieving the validated tenant name from a bearer token (in addition to existing calls for comparing against an expected name)

With these changes, the service can instead be run as a single instance from the `application` namespace, where it will have access to Hasura secrets. It would have been a bit redundant to run identical copies (only changing `-tenantname`) in each tenant namespace anyway. 

# Have you...

* [X] Added an explanation of what your changes do?
* [X] Thought about which docs need updating?


jgehrcke:
Thank you for this lovely description, @nickbp.

go/cmd/config/README.md
```diff
@@ -1,6 +1,6 @@
-# GraphQL HTTP service
+# HTTP Config API service
```
jgehrcke, line 1: :+1:

go/cmd/config/README.md
```diff
@@ -18,18 +18,19 @@ Terminal B:
...
 HASURA_GRAPHQL_ADMIN_SECRET=myadminsecret \
-./graphql \
+./config \
   --loglevel debug \
   --listen "127.0.0.1:8989" \
-  --tenantname tenant-foo \
   --disable-api-authn
 ```
 
+In normal use, the config service extracts the tenant name from the bearer token that must be provided with requests. When instead testing with `--disable-api-authn`, the service requires that we provide the tenant name using an `X-Tenant` header, as provided in the examples below.
```
jgehrcke, line 32: This paragraph makes me think a bit about terminology and concepts, both of which of course are inherently evolving as our product evolves.

Instead of "the bearer token" I so far thought of this as the "Opstrace data API authentication token".

Of course, in colloquial language we will always use shorter versions of this. But, based on experience: instead of using implementation details in the names we call it (such as 'bearer' and 'jwt'), let's maybe say "data API auth token".

That then raises an interesting question: does it make sense to use the data API authentication token as a valid credential against the config API? Probably yes!

Does it make sense to maybe adjust the name "Opstrace data API authentication token" to something else? Probably yes!

"Data API" and "config API" together could be something like "tenant API"?

Maybe Opstrace tenant API authentication token.

A decisive property of the data API authentication tokens that they are _long-lived_ primary credentials.

They have to be, in the trade off space between security and robustness (the observability data ingest path can't be bothered with the hiccups you automatically get by using short-lived authentication tokens).

Does the config API need long-lived authentication tokens? Should the config API maybe, for starters, only be exposed through the UI, involving short-lived secondary credentials?

No need to find answers now.

nickbp: The naming could go against the identity being attested in the token. So if the token just identifies the tenant, then something like "tenant key" or "tenant token" might work. If it also identifies both a user + a tenant the user has access to, it could be a "tenant user key" or "tenant user token"? But I'm not very good with names.

PR #355, opened 2021-02-12
controller: Add config API pod to application deployment

nickbp:
This deploys the `config` service from `go/cmd/config/`. It serves endpoints for adding/updating/removing credentials and exporters in the GraphQL service. It is accessible at e.g. `config.cluster-name.opstrace.io`. The endpoints are currently documented in `go/cmd/config/README.md`, planning on writing a nice user-facing guide once the controller sync support is complete.

The controller deploys a single common `config` ingress/service/pod into the `application` namespace, rather than the per-tenant namespaces, so that it can use the existing Hasura access secrets in the `application` namespace. This felt like a better tradeoff than running instances in each of the tenant namespaces, which would have meant needing to add a copy of the Hasura secret into the per-tenant namespaces as well. So, when the config service receives a request over HTTP, it extracts the tenant name from the (required and validated) bearer token, rather than being "locked" to a single tenant. See also #351.

As of this PR, we support writing/updating/deleting credentials and exporter configs in GraphQL via `curl`. The interface itself has pretty good validation and error handling for the configs, and should be pretty much ready to go. However, there isn't anything that actually consumes the configs yet. Next steps:
- Updating the controller itself to automatically sync GraphQL credentials/exporters into K8s tenant namespaces as Secrets and Deployments.
- Once things are known-working, writing a user guide for the docs describing how to configure and deploy exporters (and their prerequisite credentials) in various environments, using `curl` commands to the config service.

# Have you...

* [X] Added an explanation of what your changes do?
* [X] Thought about which docs need updating?

### DIFF
--- a/packages/controller-config/src/docker-images.json
+++ b/packages/controller-config/src/docker-images.json
@@ -2,6 +2,7 @@
   "addonResizer": "k8s.gcr.io/addon-resizer:1.8.4",
   "certManagerCAInjector": "quay.io/jetstack/cert-manager-cainjector:v1.1.0",
   "certManagerController": "quay.io/jetstack/cert-manager-controller:v1.1.0",
+  "configApi": "opstrace/config-api:c5b1a59208bc8712918010e90546b9c96e2d0e40",
   "cortex": "cortexproject/cortex:v1.7.0-rc.0",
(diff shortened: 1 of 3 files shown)


nickbp:
As a sanity check, I did a quick test cluster creation with all the controller changes and the config pod came up and was reachable at `https://config.<cluster-name>.opstrace.io`, so this part is looking good at this point.

PR #374, opened 2021-02-17
maintenance: bump golang to 1.16

sreis:
A new version is out https://blog.golang.org/go1.16


sreis:
Ok, spent enough time on this one already. Golang 1.16 requires the use of Go modules by default and this seems to break golangci-lint running in the CI Docker container.

PR #375, opened 2021-02-17
ddapi: initial service check support (WIP)

jgehrcke:


jgehrcke:
Using this custom service check:
```python
import time
import random
import datadog as dd


dd.initialize({"statsd_host": "127.0.0.1", "statsd_port": 8125})

while True:

    if random.random() > 0.04:
        dd.statsd.service_check(
            check_name="app.service_check",
            status=0,
            message="Application is OK",
        )

    else:
        dd.statsd.service_check(
            check_name="app.service_check",
            status=2,
            message="Application is not OK",
        )

    time.sleep(1.0)
```
It generates a bad (non-zero) status with low probability.

Here is how this appears in the readout in my current development version:

Screenshot from 2021-02-17 15-06-39

One interesting decision to achieve this result is to _not_ use the `message` property as a time series label value. Querying and plotting `app_service_check` would otherwise yield two different time series, one for each message value. Which could be compensated for through advanced query techniques, but I think that's not what we want for starters. At least, I'd like to learn from here.

The message for non-zero status check updates is not entirely lost, but logged by the API implementation. Example:
```
Message for check app.service_check with status 2 (timestamp: 1613570905): Application is not OK  source="translate.go:133"
```

jgehrcke:
Naive query:

Screenshot from 2021-02-17 15-38-28

Query that has the goal to find all instances of non-zero service check status updates over a rolling time window:
Screenshot from 2021-02-17 15-38-11

jgehrcke:
Failed only on `main-test-gcp`:
```
[2021-02-17T15:27:03Z]   19 passing (2m)
[2021-02-17T15:27:03Z]   1 failing
[2021-02-17T15:27:03Z]
[2021-02-17T15:27:03Z]   1) DD API test suite
[2021-02-17T15:27:03Z]        dd_api_run_agent_container_query_sysuptime:
[2021-02-17T15:27:03Z]      Error: DD agent container setup failed: deadline hit waiting for log needle
[2021-02-17T15:27:03Z]       at startDDagentContainer (test_dd.ts:214:13)
[2021-02-17T15:27:03Z]       at processTicksAndRejections (internal/process/task_queues.js:93:5)
[2021-02-17T15:27:03Z]       at Context.<anonymous> (test_dd.ts:323:32)
```

PR #381, opened 2021-02-18
DD API: unit test work

jgehrcke:


jgehrcke:
rebased on main and force-pushed

jgehrcke:
rebased on main and force-pushed

Issue #24, opened 2020-11-30
review cluster name length limitation (13 chars right now)


jgehrcke:
The Opstrace cluster name is used in contexts that impose limitations on length and character set.

Historically, that's how we ended up with current cluster name limitations.

https://github.com/opstrace/opstrace/blob/7b6b7f589f069c53357be3a83be36e459998b6d1/packages/cli/src/util.ts#L25

Some of these limitations go back to infrastructure components that we don't use anymore, such as Bigtable. See this commit: https://github.com/opstrace/opstrace/commit/b1ea161f62d3be3a30fbea0c00a435f999aac3fe

```
  // const infraNamePrefix = getInfrastructureName(stack.org, stack.name);
  // return `${infraNamePrefix}-idx`;
  // Note that the derived Bigtable instance cluster ID must not get longer
  // than 30 chars.
  return `${stack.name}-idxdb`;
};
```

TODO: play with this, see which length limitation is as of 'today' still justified. Then adjust the limit.

jgehrcke:
@triclambert what are your findings?

triclambert:
I have not yet had time to investigate.  I will report back when I do.

jgehrcke:
Investigation work and changes can be found in https://github.com/opstrace/opstrace/pull/423.

jgehrcke:
The new cluster name length constraint is 23 (max length). Has been tested in https://github.com/opstrace/opstrace/pull/423.

PR #383, opened 2021-02-19
update dashboards and other operational changes

triclambert:
* Fix dashboards
  - [x] Update 'container_name' label
  - [x] Make tenant list generated dynamically
  - [x] Add PVC graphs to overview dashboard
  - [x] Add pod detail graphs
  - [x] Remove swap graph and pod states graph
  - [x] Polish things here and there (e.g., remove legends where messy)
  - [x] Strip `-tenant` from the label_values results on some dashboards, as it was breaking them
  - [x] Remove unused dashboards


packages/cli/src/util.ts
```diff
@@ -32,8 +32,8 @@ import * as list from "./list";
  * messages.
  */
 export function validateClusterNameOrDie(cn: string): void {
-  if (cn.length > 13) {
-    die(`cluster name must not be longer than 13 characters`);
+  if (cn.length > 21) {
+    die(`cluster name must not be longer than 21 characters`);
```
triclambert, line 36: @spahl @sreis @jgehrcke any issues with this?

MatApple: It was originally set to 13 Chars because that was the limit for the BigTable instance name (when we had it for Loki/Cortex). Since that's no longer used then this should be fine. We may hit some issue elsewhere with another cloud service that has aggressive char limits on their Ids, since we use the cluster name in _most_ cloud infra Ids.

I'd suggest testing this with a cluster name that uses 21 chars in AWS and GCP

jgehrcke: Let's please update https://github.com/opstrace/opstrace/issues/24

triclambert:
Dashboards looking good.  Local build of the installer did not pick up the validation change; need to dig further.

MatApple:
It’s the controller that bundles the dashboards

> 
> 
>

triclambert:
Yup, the controller worked just fine.

packages/controller/src/resources/monitoring/system/dashboards/cortex-overview-per-tenant.json
```diff
@@ -158,7 +158,7 @@
       "reverseYBuckets": false,
       "targets": [
         {
-          "expr": "sum by (le) (rate(cortex_api_proxy_request_duration_seconds_bucket{job=\"$tenant-tenant.cortex-api\",method=\"GET\",status_code=\"200\",route=~\"api_prom_api_v1_query\"}[$timewindow]))",
+          "expr": "sum by (le) (rate(cortex_api_proxy_request_duration_seconds_bucket{job=\"$tenant-tenant.cortex-api\",method=\"POST\",status_code=\"200\",route=~\"api_v1_push\"}[$timewindow]))",
```
jgehrcke, line 161: Why this change?

jgehrcke: The center column in these dashboards is supposed to show the read direction, as indicated in the title of this panel:

```
API proxy /api/prom/api/v1/query, 200-acked GETs, request duration, 
```
changing this to keep track of POST requests would result in a pretty dangerous mismatch between title and actual plot :)

packages/controller/src/resources/monitoring/system/dashboards/cortex-overview-per-tenant.json
```diff
@@ -71,7 +71,7 @@
       "steppedLine": true,
       "targets": [
         {
-          "expr": "sum by (status_code) (rate(cortex_api_proxy_request_duration_seconds_count{job=\"$tenant-tenant.cortex-api\",method=\"POST\",route=\"api_prom_push\"}[$timewindow]))",
+          "expr": "sum by (status_code) (rate(cortex_api_proxy_request_duration_seconds_count{job=\"$tenant-tenant.cortex-api\",method=\"POST\",route=\"api_v1_push\"}[$timewindow]))",
```
jgehrcke, line 74: can you update the title of this panel, too, please?

jgehrcke: (it says `API proxy /api/prom/push` still right now)

packages/controller/src/resources/monitoring/system/dashboards/cortex-overview-per-tenant.json
```diff
@@ -2453,6 +2423,54 @@
...
+            "selected": false,
+            "text": "v3",
+            "value": "v3"
+          },
+          {
+            "selected": false,
+            "text": "launchpad",
+            "value": "launchpad"
+          }
+        ],
```
jgehrcke, line 2463: I hope we can make this dynamic at some point! A drop-down, showing the actual tenants in the current cluster.

packages/controller/src/resources/monitoring/system/dashboards/opstrace-overview.json
```diff
@@ -348,7 +205,7 @@
       "timeFrom": null,
       "timeRegions": [],
       "timeShift": null,
-      "title": "[LOGS] API proxy /loki/api/v1/push, POSTs, rate by status code (5m mean)",
+      "title": "[LOGS] API proxy /loki/api/v1/push, POSTs, rate by status code (5m mean) ALL tenants",
```
jgehrcke, line 208: :rocket:

packages/controller/src/resources/monitoring/system/dashboards/cortex-overview-per-tenant.json
triclambert: This was a mistake, reverting.

packages/controller/src/resources/monitoring/system/dashboards/cortex-overview-per-tenant.json
triclambert: This is a WIP.  The regex below does make this dynamic, but it saves the results here in the config.  I need to test omitting that part, or at least reverting back to "default".

packages/controller/src/resources/monitoring/system/dashboards/cortex-overview-per-tenant.json
```diff
@@ -643,7 +643,7 @@
       "steppedLine": true,
       "targets": [
         {
-          "expr": "sum by (status_code) (rate(cortex_api_proxy_request_duration_seconds_count{job=\"$tenant-tenant.cortex-api\",method=\"GET\",route=~\"api_prom_api_v1_query.*\"}[$timewindow]))",
+          "expr": "sum by (status_code) (rate(cortex_api_proxy_request_duration_seconds_count{job=\"$tenant-tenant.cortex-api\",method=\"GET\",route=~\"metrics\"}[$timewindow]))",
```
triclambert, line 646: The only route available on the read path appears to be 'metrics'.  Is this expected @sreis @jgehrcke?

packages/controller/src/resources/monitoring/system/dashboards/opstrace-overview.json
```diff
@@ -545,26 +402,27 @@
         {
           "expr": "sum(rate(loki_distributor_lines_received_total[5m]))",
           "interval": "",
-          "legendFormat": "log entry rate",
+          "legendFormat": "entry rate",
           "refId": "A"
         },
         {
-          "expr": "avg(sum(rate(loki_distributor_lines_received_total[12h])))",
+          "expr": "sum(rate(loki_distributor_bytes_received_total[5m]))/sum(rate(loki_distributor_lines_received_total[5m]))",
```
triclambert, line 409: Is this "bytes per entry" calculation correct and useful?

packages/controller/src/resources/monitoring/system/dashboards/cortex-overview-per-tenant.json
triclambert: I think this is working now.  Will continue testing, and then update other dashboards.  Resolving this comment.

packages/controller/src/resources/monitoring/system/dashboards/cortex-overview-per-tenant.json
jgehrcke: The Grafana/browser based query path is expected to be a blind spot when using the data API proxy-based metrics! (those queries do not go through the data API proxies).

triclambert: Not quite following, are you saying that the `metrics` route is for browser-based queries?

Stepping back to the problem:  this graph appears empty without a change.  The only route that appears to be available is "metrics".

triclambert:
@sreis I added a few panels to the overview dashboard based on our work today.  PTAL.

packages/controller/src/resources/monitoring/system/dashboards/opstrace-overview.json
triclambert: FWIW, I have now added a panel just like this that is "`BY tenants`", i.e., it is broken down by tenant rather than summed across them.  Please let me know how useable this is, especially the title of the panel.

packages/controller/src/resources/monitoring/system/dashboards/opstrace-overview.json
triclambert: After using it a bit, we should keep it around but revisit in the future.

Issue #386, opened 2021-02-19
ci: cert check: grep for Fake LE Intermediate fails (permanent main breakage?)


jgehrcke:
https://buildkite.com/opstrace/prs/builds/3595#ab3cc734-0cee-4b38-beec-b957a3a20521/1981-1982
Seen bei @MatApple on #220 
```
checking cluster is using certificate issued by LetsEncrypt
[2021-02-19T06:32:58Z] + retry_check_certificate loki.system.bk-3595-9c6-g.opstrace.io:443
[2021-02-19T06:32:58Z] + count=0
[2021-02-19T06:32:58Z] + retries=3
[2021-02-19T06:32:58Z] + check_certificate loki.system.bk-3595-9c6-g.opstrace.io:443
[2021-02-19T06:32:58Z] + openssl x509 -noout -issuer
[2021-02-19T06:32:58Z] + tee /dev/stderr
[2021-02-19T06:32:58Z] + timeout --kill-after=10 10 openssl s_client -showcerts -connect loki.system.bk-3595-9c6-g.opstrace.io:443
[2021-02-19T06:32:58Z] + grep 'Fake LE Intermediate'
[2021-02-19T06:32:58Z] depth=1 C = US, O = (STAGING) Let's Encrypt, CN = (STAGING) Artificial Apricot R3
[2021-02-19T06:32:58Z] verify error:num=20:unable to get local issuer certificate
[2021-02-19T06:32:58Z] verify return:1
[2021-02-19T06:32:58Z] depth=0 CN = bk-3595-9c6-g.opstrace.io
[2021-02-19T06:32:58Z] verify return:1
[2021-02-19T06:32:58Z] DONE
[2021-02-19T06:32:58Z] issuer=C = US, O = (STAGING) Let's Encrypt, CN = (STAGING) Artificial Apricot R3
[2021-02-19T06:32:58Z] + retcode=1
[2021-02-19T06:32:58Z] + wait=5
[2021-02-19T06:32:58Z] + count=1
[2021-02-19T06:32:58Z] + '[' 1 -lt 3 ']'
[2021-02-19T06:32:58Z] + echo 'failed checking if cluster is using certificate issued by LetsEncrypt, retrying in 5 seconds...'
[2021-02-19T06:32:58Z] failed checking if cluster is using certificate issued by LetsEncrypt, retrying in 5 seconds...
[2021-02-19T06:32:58Z] + sleep 5
[...]

PR #389, opened 2021-02-19
ci: cert check: update issuer grep text

jgehrcke:
LE changed the issuer name of the certificate used to sign staging environment certificates on Feb 18: https://community.letsencrypt.org/t/staging-hierarchy-changes/145633

We check for this issuer name in CI to confirm that the certs exposed by API endpoints are indeed derived from
LE infrastructure. After the LE-internal naming change we have to update our check. See issue #386.


jgehrcke:
the grep check passed. next failure, as expected is, https://github.com/opstrace/opstrace/issues/393.

PR #394, opened 2021-02-19
test-remote: tmp-disable test as of #393

jgehrcke:
sad tmp fix for https://github.com/opstrace/opstrace/issues/393.

PR #398, opened 2021-02-19
test-remote: simpler test deactivation

jgehrcke:
yolo-merging https://github.com/opstrace/opstrace/pull/394 was not a good idea. It introduced a new problem.

It's sad that there is no simple way to deactivate a test (like a 'marker').

Another pragmatic idea, returning early from a function, makes compilation fail.

Now I try outcommenting the entire test function. Hopefully this does not yield compilation errors.

### DIFF
--- a/test/test-remote/test_dd.ts
+++ b/test/test-remote/test_dd.ts
@@ -168,7 +168,8 @@ export async function startDDagentContainer() {
   const logNeedle = `Successfully posted payload to "${TENANT_DEFAULT_DD_API_BASE_URL}/api/v1/series`;
 
   // It's known that this may take a while after DD agent startup.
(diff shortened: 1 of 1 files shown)


MatApple:
I'm assuming you're referring to the fact there is no utility for disabling a TDD style test? The only testing I've previously done with Mocha has all been with the BDD syntax which I'm assuming functions the same as `test("` or TDD style. IIRC the only difference is readability between the two styles.

https://stackoverflow.com/questions/32723167/how-to-programmatically-skip-a-test-in-mocha

You should be able to use `xit("` instead of `it("` to disable a test (the BDD way of writing `test("`).

Instead of `suite("` you can use the BDD style `describe("` and then disable it with `describe.skip("`.

PR #402, opened 2021-02-20
bump: cortex 1.7.0 rc0 -> rc2

jgehrcke:
Cortex 1.7.0-rc.2 was released about a day ago.

Since rc0, there have been two important fixes:

> [BUGFIX] Handle missing samples due to large steps and single point extents
> [BUGFIX] Fix ring tokens sorting regression introduced in #3601

https://github.com/cortexproject/cortex/releases


jgehrcke:
Both, main-test-aws and main-test-gcp were affected by (only) https://github.com/opstrace/opstrace/issues/405. Merging.

Issue #379, opened 2021-02-17
Move alertmanager and credential/exporter config APIs into common /v1/ path under the root cluster domain


nickbp:
DEADLINE: Try to have this figured out by the end of the week (Feb 19)

There are currently multiple endpoints for setting configuration for tenants:
- `config.cluster.example/api/v1/credentials/*`
- `config.cluster.example/api/v1/exporters/*`
- `cortex.cluster.example/ruler/*`
- `cortex.cluster.example/api/v1/rules/*`
- `cortex.cluster.example/api/v1/alerts/*`
- `cortex.cluster.example/alertmanager/*`
- `cortex.cluster.example/multitenant_alertmanager/*`

This current structure has a few issues:
- The `config` subdomain used by credential/exporter config may conflict with a tenant named `config`, and it's a bit confusing since the subdomains are ideally supposed to all be about tenants anyway.
- For alertmanager configuration, we'd ideally like configuration endpoints to be separate from data ingest endpoints.

So we'd ideally like to only use subdomains for per-tenant data ingest, and then have these config endpoints at the root domain (shared with the UI). Requests to the config endpoints would identify the tenant name using the `Bearer` token, as is currently the case.

The changes will be to move the endpoints to the root domain, under a reserved `/api/*` path like the following:
- Move `config.cluster.example/api/v1/credentials/*` to `cluster.example/api/v1/credentials/*`
- Move `config.cluster.example/api/v1/exporters/*` to `cluster.example/api/v1/exporters/*`
- `cortex.cluster.example/ruler/*` to `cluster.example/api/v1/ruler/*`
- `cortex.cluster.example/api/v1/rules/*` to `cluster.example/api/v1/rules/*`
- `cortex.cluster.example/api/v1/alerts/*` to `cluster.example/api/v1/alerts/*`
- `cortex.cluster.example/alertmanager/*` to `cluster.example/api/v1/alertmanager/*` (may go to `cluster.example/alertmanager/*` in the future)
- `cortex.cluster.example/multitenant_alertmanager/*` to `cluster.example/api/v1/multitenant_alertmanager/*` (may go to `cluster.example/multitenant_alertmanager/*` in the future)

The changes would be applied in a few places:
- Move the alertmanager config endpoint handling into the API config executable (see `/go/cmd/config/`)
[...]

Issue #358, opened 2021-02-12
Write guide for adding credentials and configuring/deploying exporters


nickbp:
Write a guide covering:
- (brief, ideally external link) Getting credentials from AWS/GCP
- Adding those credentials into opstrace via curl
- Adding stackdriver/cloudwatch exporters into opstrace via curl, using a basic config example
- Observing the metrics appearing in opstrace

triclambert:
Should this be a user guide or an administrator guide?

nickbp:
I'm thinking it'd fall under administrator guides, since it'd mainly be about configuring retrieving metrics from other running systems.

But this assumes that a "user guide" would mainly be about building dashboards etc against metrics that are already being collected.

nickbp:
WIP: https://docs.google.com/document/d/1DW2uRwLBkaDFhw81yfo3EC1aW40_oTCii_kKUQDqBZE/edit?usp=sharing

Issue #312, opened 2021-02-01
Update backend to deploy cloudwatch/stackdriver into tenant namespace, get prometheus scraping working


nickbp:
Following #311, get the exporters deployed into per-tenant namespaces in K8s, via the controller deploying them.

This would also involve providing some hardcoded configuration for what the exporters should be collecting from, and any credentials needed to access that information. Next step/ticket would be to expose the API to provide configuration, and replace the hardcoded config used in this phase.

To validate that data is flowing with stub configuration, figure out how to configure prometheus to scrape from the exporters as well.

nickbp:
Currently working on the remaining parts which are all in the controller:
- Deploying the graphql service pod in tenant namespaces (implemented locally, needs testing)
- Syncing credentials from graphql as K8s Secrets (working on this now)
- Syncing exporters from graphql as K8s Deployments (working on this next - should be functionally similar to credential sync)

nickbp:
As of #355 just getting merged, the code for deploying the `config` pod in the application namespace is in. It was renamed from `graphql` to be less confusing since it's not serving graphql endpoints, and it was moved from per-tenant namespaces (and "locked" to working on one tenant at a time) to instead be deployed as part of the application namespace and accept the tenant token dynamically from the request. This was mainly because the hasura secret is only created in the application namespace, and because having multiple identical instances of the config interface is a bit redundant anyway.

For this ticket the main thing left is to add credential and exporter sync support to the controller, where additions/changes/deletions are automatically reflected in K8s. To keep it simple I'm planning on just having it periodically poll the list of credentials/exporters from graphql, and reconcile that with the current K8s state.

nickbp:
I've been working on adding credential sync to the controller, where it listens for created/updated/deleted credentials in GraphQL and automatically writes them as Secrets against the tenant namespaces in K8s. After a few false starts I feel like I've started getting the hang of how the controller works, and have a structure that should do the job. There's still some work around the GraphQL interaction that I still need to figure out - mainly just to figure out either a polling-based method or subscription-based method of querying the Credential data itself that's compatible with how Informers work. Once I've gotten things working for GraphQL Credential->Kube Secret sync, I'm expecting it to be relatively trivial to do something similar for GraphQL Exporter->Kube Deployment sync. The sync logic should be identical, so it should just be a matter of writing the templating part that takes an Exporter config and converts it into a Deployment object, as opposed to the existing code that converts a Credential config to a Secret object.

[...]

Issue #310, opened 2021-02-01
Collect metrics and logs from CloudWatch and Stackdriver


nickbp:
There are many interesting metrics that existing only in cloud vendors' own metrics system. We should be able to ingest those (with an allowlist) into Opstrace.

Details TBD.

opstracy:
###### do-not-edit-start-codetree-epic-issues

#### Issues in this epic:

Title | Milestone | Assignees | Stage | State
----- | --------- | --------- | ----- | -----
![](https://cdnjs.cloudflare.com/ajax/libs/octicons/4.4.0/svg/issue-opened.svg) [Read cloudwatch/stackdriver exporter docs, get running locally #311](https://github.com/opstrace/opstrace/issues/311) | N/A | nickbp | Open


###### do-not-edit-end-codetree-epic-issues

nickbp:
Quick update:
- The graphql interface and associated golang service for accepting YAML payloads are "complete" but not merged into main yet (#314). This included a lot of initial investigative work around getting golang code to interact with graphql. There were some issues with the graphql code generator we're using but they had an easy temporary workaround. In the meantime it should be relatively straightforward to add more golang<->graphql interaction in the future.
- Next up is getting the controller to subscribe to the credentials/exporters tables and automatically syncing changes to K8s, and then ensuring that prometheus/cortex is able to find/scrape the exporters (#312).

nickbp:
The main implementation was completed via #403. A user should be able to invoke the Go HTTP API to add/update/delete credentials/exporters, where the controller deploys them automatically to K8s. In manual testing, I was able to see exported metrics appearing in cortex.

Some loose ends are broken out into separate tickets, and may tie into the UI config interface when it's being implemented:
- #358: User/admin facing documentation for using curl (can derive from the current go/cmd/config/README.md)
- #373: Using Hasura Actions for running the Go config validation code on behalf of the UI

PR #403, opened 2021-02-22
controller: Apply credential/exporter changes from graphql to Kubernetes

nickbp:
When a given credential or exporter config is created/updated/deleted in GraphQL, the controller will automatically apply that change to Kubernetes. Credentials are converted to Kubernetes `Secret`s, while exporters are converted to Kubernetes `Deployment`s (which may link to the `Secret`s referenced by the exporter config).

The synchronization itself works against the existing Kubernetes reconciliation loop in the Controller. The main addition is that the reconciliation state now includes entries originating from GraphQL, but the reconciliation flow itself remains the same structurally and ends up being pretty straightforward. This structure would potentially be extendable to other state retrieved from GraphQL in the future.

Summary:
- Add GraphQL queries for dumping a "snapshot" of credentials and exporters when starting the sync, in addition to the existing subscription queries.
- Add subscription client to controller, largely copied from app.
- Implement `informers` for fetching and subscribing to the GraphQL credentials and exporters, and applying this data to the Redux state object.
- Implement `resources` for converting the Redux state to Kubernetes objects, with type-specific logic for how to deploy each credential and exporter type.

I've so far tested this by hand by creating credentials/exporters via `curl` against the go graphql `config` service (which is due to get moved/renamed), and was able to verify that actual metrics successfully appeared in prometheus/cortex from each of the current two types of exporters. As such this should complete issue #312 and epic #310 . There remains some doc work via #358 which in turn depends on the graphql config API naming that's being updated via #379 .

# Have you...


packages/app/src/state/credential/queries/subscribeToCredentialList.gql
```diff
@@ -1,8 +1,13 @@
+# Subscribes to credential updates, INCLUDING SECRET VALUES, across all tenants.
+# Used for syncing credential secrets to Kubernetes.
+# TODO: ability to restrict access to this API to just the controller?
```
MatApple, line 3: :+1:

packages/controller/package.json
```diff
@@ -35,6 +36,7 @@
     "js-yaml": "^3.14.0",
     "redux": "^4.0.4",
     "redux-saga": "^1.1.3",
+    "subscriptions-transport-ws": "^0.9.18",
```
MatApple, line 39: cool, subscriptions in node turned out to be easier than expected!

packages/controller/src/dbSubscriptionClient.ts
```diff
@@ -0,0 +1,82 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { WebSocketLink } from "@apollo/client/link/ws";
+import { onError } from "@apollo/client/link/error";
+// Import from apollo/client/core, not apollo/client, to avoid React dependency.
```
MatApple, line 19: 🥇

packages/controller/src/reducers/graphql/credentials.ts
```diff
@@ -0,0 +1,145 @@
...
+    if (!dbClient) {
+      log.warning(
+        "skipping credential sync due to missing env vars GRAPHQL_ENDPOINT & HASURA_GRAPHQL_ADMIN_SECRET"
+      );
+      return;
+    }
+    // Do initial load up-front
+    try {
+      const res = await dbClient.GetCredentialsDump();
+      if (res.data?.credential && res.data.credential.length > 0) {
```
MatApple, line 105: Should we emit a "success" event if credentials is an empty array?

MatApple (approved):
Great work @nickbp 🎉

MatApple:
Merging since GCP test failures are expected and resolved in main.

Issue #380, opened 2021-02-18
UI: Create an interface that allows easy setup of the Cortex Alertmanager configuration


terrcin:
This should be a domain aware html form interface that automatically generates the raw `yaml` configuration file and submits it to the new api endpoint.

The documentation for the `yaml` file format is here: https://www.prometheus.io/docs/alerting/latest/configuration/#slack_config
Documentation on Cortext API is here: https://cortexmetrics.io/docs/api/#get-alertmanager-configuration

Important feature elements:

* Seperate configuration per`Tenant` - this is mainly aimed at the Rule section of the `yaml`.
* Where apporpriate sharing configuration elements, such as smtp configuration, slack api endpoint, and receivers between `Tenants`

terrcin:
Slight spanner in the works. Users can directly upload the config.yaml via the cli which won't be picked up by the GUI I'm creating as it stores the config seperately in a structured data format in postgres/hasura. Any changes directly made via the CLI won't be reflected in the GUI and will infact be override by subsequent changes made in the UI.

The GUI is important as lots of users won't be familiar with the config.yaml format so we'll need to be able to parse the yaml into the structured data format. This has the added benefit of an "import yaml" button in the GUI being straight forward to do. I've done a bit of snooping and there is hopefully existing libraries that allow me to define a "schema" for the config.yaml and then easily import and parse it. Yaml is pretty straight forward anyways.

In the mean time the current goal has temporarily been redirected to providing a UI for uploading and editing the config.yaml directly. Something along these lines:

![Screenshot-20210223194611-1972x1310](https://user-images.githubusercontent.com/17045/108816475-1ac14380-761b-11eb-97ba-c263f2623890.png)

Currently being worked on in: https://github.com/opstrace/opstrace/tree/terrcin/alert-manager-yaml-editor

nickbp:
I'd done a little bit of looking into Hasura Actions via #373. It looks like the case of reading/writing the alertmanager config should be more straightforward than what I was wanting to do in #373 (specifically, that needs the data to also be written to Hasura after the action completes successfully), so I'm going to see if I can get this working first and then use that as a foundation for #373.

I'm thinking the get/update flow would look like this:
```
UI --graphql--> Hasura --action--> AlertManager Config API (get/set) --???--> storage
```

[...]

PR #404, opened 2021-02-22
[WIP] UI: Create an interface that allows easy setup of the Cortex Alertmanager configuration

terrcin:
Initial implementation of feature #380


terrcin:
This is my slightly tidied up initial commit for this work (in progress). I've been learning the code base, libraries used, way things are done etc... so it's still a bit all over the place at the moment. Now that I think I have a handle on things and direction that I'm going in I want to switch to towards a more normal workflow of frequent small commits.

Current aim is for the user to be able to specify a slack endpoint and channel for all a Tenants alerts to be posted to. Get this fully working then start expanding out support for the full configuration yaml.

terrcin:
changed solution to directly editing yaml rather than doing fancy form interface and resolved in different PR - https://github.com/opstrace/opstrace/pull/482

PR #413, opened 2021-02-24
Tune Cortex/Loki

MatApple:
This PR does the following:
1. Increases Persistent Volume size for Cortex and Loki components (10GB is not enough for large ingest volume)
2. Adjusts our scaling formula for Loki/Cortex components. The adjustments that have been made are targeted for a 5 node cluster and above. 3 node clusters stay the same.
3. Adds a dry run mechanism to the controller, so all kubernetes api writes will be logged but not executed. Enable this with the following env var: `DRY_RUN_KUBERNETES_API_WRITES=true`
4. Enables configurable logging level in the controller through the env var: `LOG_LEVEL=<debug|warn|info|error>`
5. Adds debug level logging for Kubernetes reconciliation actions to log resource diffs when resource specs have changed.

To run the controller locally in dry run mode and to see the debug level logging:
`DRY_RUN_KUBERNETES_API_WRITES=true LOG_LEVEL=debug node ./build/cmd.js <instancename> --external`

Example diff output when changing a resource's configuration (in this case I changed the port on the cortex gateway deployment for demo purposes):
Screen Shot 2021-02-24 at 2 00 39 PM

**A note on the logged diff**: Kubernetes will add default values for fields we don't specify e.g. the port `"protocol": "TCP"` in the above image is a default. Kubernetes will also add a number of other fields in the resource spec (relating to status), hence the diff may have a lot of red because we don't specify these values and rely on the defaults. The green `"env": []` addition to the spec is an example of where we specify an empty array of env vars and the Kubernetes api strips the key entirely when it's an "empty-like" value.


jgehrcke:
A brief description / reasoning would be nice!

(our future us will thank us for explaining these commits a little bit)

packages/controller/src/resources/loki/index.ts
```diff
@@ -833,7 +862,7 @@ export function LokiResources(
                 accessModes: ["ReadWriteOnce"],
                 resources: {
                   requests: {
-                    storage: "10Gi"
+                    storage: "30Gi"
```
sreis, line 865: Given the load we have seen thus far this change is not necessary and because it increases the cost profile we should not do it now. Or do you prefer to keep the sizes in sync with Cortex?

MatApple: Ok I reset all the Loki PVCs back to 10Gi

lib/kubernetes/src/api.ts
```diff
@@ -39,24 +50,51 @@ export const createResource = async (resource: K8sResource): Promise<void> => {
...
     if (err.statusCode !== 404) {
-      log.error(err.message);
+      log.info(
+        "api err during deleteResource for %s(ns: %s): %s",
+        resource.name,
+        resource.namespace,
+        err.message
+      );
+    } else {
+      log.debug("deleteResource(): ignored 404 error");
```
jgehrcke, line 73: want to show the main error message here of the ignored error? :) is sometimes super valuable.

lib/kubernetes/src/api.ts
```diff
@@ -18,8 +18,19 @@ import { K8sResource } from "./common";
 import { kubernetesError } from "./utils";
 import { log } from "@opstrace/utils";
 
+const dryRunKubeApiWrites = process.env.DRY_RUN_KUBERNETES_API_WRITES
```
jgehrcke, line 21: In a cluster that was normally set up -- what is the best / intended way to set this env var, and then make it take effect? Can we test this as part of CI with a little bit of a `kubectl` dance?

lib/kubernetes/src/api.ts
MatApple: It's just a generic Kubernetes 404 error message "<type> <name> does not exist". In this case, it's not useful at all and it's already quite noisy because we'll get a lot of these 404s on teardown

lib/kubernetes/src/api.ts
MatApple: I imagine you'd only ever use it when running the controller locally during manual controller testing (to see what _this version of the controller_ would change). At least that's the extent of its purpose at this point.

PR #422, opened 2021-02-25
controller: Restart cloudwatch exporter automatically on config changes

nickbp:
Adds a bash sidecar to the cloudwatch pod that checks the config file in a loop. When a change is detected, the sidecar exits, causing the exporter to restart with the new config. A bit hacky but should get the job done, and better than leaving the pod to ignore config changes.

Inspired by https://github.com/jimmidyson/configmap-reload , which unfortunately doesn't work for us since the cloudwatch exporter doesn't have a reload nor restart HTTP endpoint.

Shouldn't be needed by the stackdriver exporter, whose config is via envvars rather than a config file, so it should just restart automatically.

# Have you...

### DIFF
--- a/packages/controller/src/resources/exporters/index.ts
+++ b/packages/controller/src/resources/exporters/index.ts
@@ -61,6 +61,9 @@ const toKubeResources = (
   const resources: Array<K8sResource> = [];
   let podSpec: V1PodSpec;
   if (exporter.type == "cloudwatch") {
+    // TODO: The exporter pod is NOT automatically restarted if this ConfigMap changes.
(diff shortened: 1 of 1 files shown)


MatApple:
@nickbp let's chat about this more tomorrow. The controller already has a built in mechanism to do this automatically. If a ConfigMap or Secret changes, any deployment with that secret or ConfigMap bound to it should also update https://github.com/opstrace/opstrace/blob/535a0478635a2ee0a3270c49ce9f78ac92568d49/lib/kubernetes/src/reconciliation/reconcile.ts#L913

nickbp:
Aha yep that ought to do it.

PR #425, opened 2021-02-25
Cortex: explicit tsdb config, enable purger

jgehrcke:
https://cortexmetrics.io/docs/configuration/configuration-file/#blocks_storage_config
https://cortexmetrics.io/docs/blocks-storage/

About `block_ranges_period`:

> The in-memory samples are periodically flushed to disk - and the WAL truncated - when a new TSDB block is created, **which by default occurs every 2 hours**. 

About `retention_period`:

> Each newly created block is then uploaded to the long-term storage and kept in the ingester until the configured -blocks-storage.tsdb.retention-period expires, in order to give queriers and store-gateways enough time to discover the new block on the storage and download its index-header.

Powerful knobs, affecting fault tolerance, local disk usage, query performance, etc.

### DIFF
--- a/packages/controller/src/resources/cortex/index.ts
+++ b/packages/controller/src/resources/cortex/index.ts
@@ -330,7 +330,7 @@ export function CortexResources(
       tsdb: {
         dir: "/cortex/tsdb",
         wal_compression_enabled: true,
-        block_ranges_period: "2h", //default
+        block_ranges_period: "2h0m0s", //default
         retention_period: "6h" //default
       },
       backend: storageBackend,
--- a/packages/controller/src/resources/cortex/index.ts
+++ b/packages/controller/src/resources/cortex/index.ts
@@ -329,7 +329,9 @@ export function CortexResources(
     blocks_storage: {
       tsdb: {
         dir: "/cortex/tsdb",
-        wal_compression_enabled: true
+        wal_compression_enabled: true,
+        block_ranges_period: "2h", //default
+        retention_period: "6h" //default
       },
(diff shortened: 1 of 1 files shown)


packages/controller/src/resources/cortex/index.ts
```diff
@@ -332,6 +332,7 @@ export function CortexResources(
         }
       }
     },
+    purger: { enable: true },
```
spahl, line 335: Purger is just for chunks right now so there is no point in enabling this API :-(

jgehrcke: > [ENHANCEMENT] Added new experimental API endpoints: POST /purger/delete_tenant and GET /purger/delete_tenant_status for deleting all tenant data. Only works with blocks storage. Compactor removes blocks that belong to user marked for deletion. #3549 #3558

(https://github.com/cortexproject/cortex/blob/master/CHANGELOG.md#master--unreleased)

! :)

spahl: Right on!

spahl: I imagine this needs a few new URLs to be proxied

jgehrcke:
Huh. Puzzled. Still not the right config value.
```
2021-02-26 15:27:10	
error loading config from /etc/cortex/config.yaml: Error parsing config file: yaml: unmarshal errors:
2021-02-26 15:27:00	
  line 60: cannot unmarshal !!str `2h0m0s` into tsdb.DurationList
```

PR #426, opened 2021-02-26
 use Recreate strategy for Loki compactor so it can transfer PV

MatApple:

### DIFF
--- a/packages/controller/src/resources/loki/index.ts
+++ b/packages/controller/src/resources/loki/index.ts
@@ -910,6 +910,9 @@ export function LokiResources(
           // no more than one compactor instance at any given time, see
           // boltdb shipper docs
           replicas: 1,
+          strategy: {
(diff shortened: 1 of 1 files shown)


jgehrcke:
https://kubernetes.io/docs/concepts/workloads/controllers/deployment/#recreate-deployment

RollingUpdate is default.

Recreate:

> All existing Pods are killed before new ones are created when .spec.strategy.type==Recreate. **Note:** This will only guarantee Pod termination previous to creation for upgrades. If you upgrade a Deployment, all Pods of the old revision will be terminated immediately. Successful removal is awaited before any Pod of the new revision is created. If you manually delete a Pod, the lifecycle is controlled by the ReplicaSet and the replacement will be created immediately (even if the old Pod is still in a Terminating state). If you need an "at most" guarantee for your Pods, you should consider using a StatefulSet.

PR #435, opened 2021-03-01
[WIP] ui: switch to using typed objects rather than arrays in redux

terrcin:
This is about moving away from lists in redux and to an objects for main data storage - the object key being the entity id and value being the entity itself. Typescript `record` types are used for this.

Among the benefits are that to find an entity, say user, we can just look it up by it's id rather than iterate over the whole list - we can still get a list if we want to by accessing the object`s values. Additionally (and primarily) arrays are not mergable, so when updating an array in a reducer we replace the entire dataset for any change no matter how trival which can have signifigant downstream re-rendering consequences.

Users have also been update to have a generated primary key rather than using the email as it's user editable data. This is actually most of the work and was a good way to continue learning the code base as it broke lots of things. :-)


terrcin:
Update: have succssfully tested the upgrade path on my aws cluster

PR #437, opened 2021-03-02
fix additional kubernetes dashboards

triclambert:


packages/controller/src/resources/monitoring/system/dashboards/controller-manager.json
```diff
@@ -504,7 +504,7 @@
           "steppedLine": false,
           "targets": [
             {
-              "expr": "histogram_quantile(0.99, sum(rate(rest_client_request_latency_seconds_bucket{job=\"kube-controller-manager\", instance=~\"$instance\", verb=\"POST\"}[5m])) by (verb, url, le))",
+              "expr": "histogram_quantile(0.99, sum(rate(rest_client_request_latency_seconds_bucket{job=\"apiserver\", instance=~\"$instance\", verb=\"POST\"}[5m])) by (verb, url, le))",
```
sreis, line 507: The graph is empty and it seems the metric is deprecated https://docs.signalfx.com/en/latest/integrations/agent/monitors/kubernetes-scheduler.html

```suggestion
              "expr": "histogram_quantile(0.99, sum(rate(rest_client_request_duration_seconds_bucket{job=\"apiserver\", instance=~\"$instance\", verb=\"POST\"}[5m])) by (verb, url, le))",
```

packages/controller/src/resources/monitoring/system/dashboards/controller-manager.json
```diff
@@ -596,7 +596,7 @@
           "steppedLine": false,
           "targets": [
             {
-              "expr": "histogram_quantile(0.99, sum(rate(rest_client_request_latency_seconds_bucket{job=\"kube-controller-manager\", instance=~\"$instance\", verb=\"GET\"}[5m])) by (verb, url, le))",
+              "expr": "histogram_quantile(0.99, sum(rate(rest_client_request_latency_seconds_bucket{job=\"apiserver\", instance=~\"$instance\", verb=\"GET\"}[5m])) by (verb, url, le))",
```
sreis, line 599: The graph is empty and it seems the metric is deprecated https://docs.signalfx.com/en/latest/integrations/agent/monitors/kubernetes-scheduler.html


```suggestion
              "expr": "histogram_quantile(0.99, sum(rate(rest_client_request_duration_seconds_bucket{job=\"apiserver\", instance=~\"$instance\", verb=\"GET\"}[5m])) by (verb, url, le))",
```

sreis (changes requested):
@triclambert I find it weird they're being collected in the `apiserver` job. Did you check why we're not collecting metrics for the `kube-controller-manager`?

triclambert:
I find it weird too, which is why I want your eyes on this.  I was not able to confirm this via release notes or documentation—this comes from a running instance only.

sreis:
> I find it weird too, which is why I want your eyes on this.  I was not able to confirm this via release notes or documentation—this comes from a running instance only.

Those pods are managed by the cloud-provider so I say we remove the controller-manager dashboard. WDYT?

triclambert:
Sounds good to me.  Are there any others we should remove?

sreis:
> Sounds good to me. Are there any others we should remove?

We can remove Kubernetes/Scheduler dashboard.

The Kubernetes/Kubelet looks like it needs to be fixed/updated too and it seems we don't collect kube-proxy metrics although the pods expose the /metrics endpoint.

triclambert:
> The Kubernetes/Kubelet looks like it needs to be fixed/updated too

I believe this is fixed in this PR.

Removed the other dashboards—please take another look.

sreis (approved):

Issue #450, opened 2021-03-05
upgrades: (cortex|dd|loki)-apis pods crash looping


sreis:
**Describe the bug**

(cortex|dd|loki|opstrace)-apis pods crash looping with the following error

```
time="2021-03-05T19:22:41Z" level=error msg="API_AUTHTOKEN_VERIFICATION_PUBKEY_SET must not be empty. Exit."
```

`loki-api` spec of a crash looping pod:

```
spec:
  progressDeadlineSeconds: 600
  replicas: 2
  revisionHistoryLimit: 10
  selector:
    matchLabels:
      k8s-app: loki-api
  strategy:
    rollingUpdate:
      maxSurge: 25%
      maxUnavailable: 25%
    type: RollingUpdate
  template:
    metadata:
      creationTimestamp: null
      labels:
        k8s-app: loki-api
    spec:
      affinity:
        podAntiAffinity:
          preferredDuringSchedulingIgnoredDuringExecution:
          - podAffinityTerm:
              labelSelector:
                matchLabels:
                  k8s-app: loki-api
              topologyKey: kubernetes.io/hostname
            weight: 100
      containers:
      - args:
        - -listen=:8080
        - -tenantname=system
        - -loki-querier-url=http://querier.loki.svc.cluster.local:1080
        - -loki-distributor-url=http://distributor.loki.svc.cluster.local:1080
        env:
        - name: API_AUTHTOKEN_VERIFICATION_PUBKEY_SET
        image: opstrace/loki-api:b3c0c6cb64ad8d38e65e2a576a4f84c3e2d8981b
        imagePullPolicy: Always
        livenessProbe:
          failureThreshold: 3
          httpGet:
            path: /metrics
            port: 8080
            scheme: HTTP
          periodSeconds: 10
          successThreshold: 1
          timeoutSeconds: 1
        name: loki-api
        ports:
        - containerPort: 8080
          name: http
          protocol: TCP
        readinessProbe:
          failureThreshold: 3
          httpGet:
            path: /metrics
            port: 8080
            scheme: HTTP
          periodSeconds: 10
          successThreshold: 1
          timeoutSeconds: 1
        resources: {}
        terminationMessagePath: /dev/termination-log
        terminationMessagePolicy: File
      dnsPolicy: ClusterFirst
      restartPolicy: Always
      schedulerName: default-scheduler
      securityContext: {}
      serviceAccount: loki-api
      serviceAccountName: loki-api
[...]

Issue #151, opened 2020-12-10
ci: waiting "for Certificate ingress/https-cert to be ready" didn't resolve


jgehrcke:
https://github.com/opstrace/opstrace/issues/62#issuecomment-742631003
https://buildkite.com/opstrace/scheduled-main-builds/builds/1341#d4dae6f8-41a1-4b9f-acdb-03dd9c577f21/221-6770
logs should be in builds.ci.aws.opstrace.io
similar to / same as https://github.com/opstrace/opstrace/issues/60?

jgehrcke:
@sreis did some debugging (thank youu!): https://github.com/opstrace/opstrace/issues/62#issuecomment-742692632

jgehrcke:
https://buildkite.com/opstrace/prs/builds/3303#c6de932f-fc48-47c1-9479-33ad90ef66f1/256-4965

sreis:
Another two:

https://buildkite.com/opstrace/scheduled-main-builds/builds/1423#ed528871-96aa-4227-a288-9e484569c9ce/222-6871

https://buildkite.com/opstrace/scheduled-main-builds/builds/1424#00b2e157-d0c7-4920-877b-b49bfee98d51/222-6819

sreis:
https://buildkite.com/opstrace/scheduled-main-builds/builds/1539#be79e5cf-b481-4338-be27-5758491b1236/221-6820

```
[2021-01-12T05:45:58Z] 2021-01-12T05:45:58.219Z debug:     Waiting for Certificate ingress/https-cert to be ready
[2021-01-12T05:46:02Z] 2021-01-12T05:46:02.179Z warning: cluster creation attempt timed out after 2400 seconds
[2021-01-12T05:46:02Z] 2021-01-12T05:46:02.179Z debug: do not log error detail
```

sreis:
https://buildkite.com/opstrace/scheduled-main-builds/builds/1550#375db81d-0482-4057-86fc-6d99ccdf1359/221-6815

```
[2021-01-14T01:47:15Z] 2021-01-14T01:47:15.706Z warning: cluster creation attempt timed out after 2400 seconds
[2021-01-14T01:47:15Z] 2021-01-14T01:47:15.706Z debug: do not log error detail
[2021-01-14T01:47:15Z] 2021-01-14T01:47:15.706Z error: 3 attempt(s) failed. Stop retrying. Exit.
```

sreis:
Also affected AWS build this time:

[...]

PR #452, opened 2021-03-06
API proxy authenticator: relax key set config validation, wip

jgehrcke:
This addresses https://github.com/opstrace/opstrace/issues/450.

### DIFF
--- a/go/pkg/authenticator/keyset_test.go
+++ b/go/pkg/authenticator/keyset_test.go
@@ -44,3 +44,14 @@ func TestKeysetFromEnv_TwoKeys(t *testing.T) {
 		"map authtokenVerificationPubKeys expected to not be empty",
 	)
 }
+
+func TestKeysetFromEnv_EmtpySetButFallback(t *testing.T) {
+	os.Setenv("API_AUTHTOKEN_VERIFICATION_PUBKEY_SET", "")
+	os.Setenv("API_AUTHTOKEN_VERIFICATION_PUBKEY", TestPubKey)
+
+	// This is now expected to _not_ crash, becuse a fallback key is
+	// configured.
+	ReadConfigFromEnvOrCrash()
+	log.Infof("keyset map:\n%v", authtokenVerificationPubKeys)
+	log.Infof("fallback key:\n%v", authtokenVerificationPubKeyFallback)
+}
--- a/go/pkg/authenticator/keyset.go
+++ b/go/pkg/authenticator/keyset.go
@@ -46,31 +46,47 @@ func keyIDfromPEM(pemstring string) string {
 }
 
 /*
-Read set of public keys from environment variables.
-If key deserialization fails, log an error
-and exit the process with a non-zero exit code.
+Read set of public keys for authentication token verification from environment.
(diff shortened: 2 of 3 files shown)


go/pkg/authenticator/keyset_test.go
```diff
@@ -44,3 +44,14 @@ func TestKeysetFromEnv_TwoKeys(t *testing.T) {
...
+
+func TestKeysetFromEnv_EmtpySetButFallback(t *testing.T) {
+	os.Setenv("API_AUTHTOKEN_VERIFICATION_PUBKEY_SET", "")
+	os.Setenv("API_AUTHTOKEN_VERIFICATION_PUBKEY", TestPubKey)
+
+	// This is now expected to _not_ crash, becuse a fallback key is
+	// configured.
+	ReadConfigFromEnvOrCrash()
+	log.Infof("keyset map:\n%v", authtokenVerificationPubKeys)
+	log.Infof("fallback key:\n%v", authtokenVerificationPubKeyFallback)
```
jgehrcke, line 56: this tests the scenario that led to the #450 breakage.

jgehrcke:
`main-test-gcp` failed with #151: https://github.com/opstrace/opstrace/issues/151#issuecomment-792301414

sreis:
@jgehrcke If I download the [CLI built from this PR](https://buildkite.com/opstrace/prs/builds/3736#12c5fd27-c744-48d1-af51-005036af887d/208-263) and run the upgrade command would that be enough for the controller to deploy the APIs with the correct env variables? Or do we need to do something else?

jgehrcke:
> would that be enough for the controller to deploy the APIs with the correct env variables?

Hm, generally -- as I haven't tested this myself, let's assume things are still broken!

The new API proxy versions are supposed to work in view of the _old_ environment variable being set.

The state is:

- _old_ controller config map
- new controller
- new API proxy versions

When I think about it that way, things are probably still broken. I was thinking of the following state when I built this patch here:

- old controller config map
- old controller
- new API proxy versions

Woof.

This is a great example of complexity that's certainly possible to manage, and even rather easy to manage buuut... where in practice things are just so so likely to break.

jgehrcke:
What one could do, hypothetically, is to make the newer controller version work with an older controller config, transparently.

However, that can't quite be maintained.

What needs to be done is to migrate the controller config to the new schema, I think.

1) Read the current controller config map.
2) Parse current authenticator public key from it
3) Construct new authenticator key set configuration document (JSON), which involves building a key ID, following these instructions: https://github.com/opstrace/opstrace/tree/main/go/pkg/authenticator#key-id-calculation
4) Building a new controller config object, with the parameter `tenant_api_authenticator_pubkey_set_json`.
5) Overwrite the controller config map.

But where? Should the controller itself do this? Sounds like the right thing to do: the controller should perform migrations for its own config, upon startup. What do we think?

----

We could also use this opportunity and re-evaluate the current upgrade strategy (that we're currently trying to make successful). I know it's a potentially hard thing to think about -- but maybe we really do want to use the more easy-to-reason-about hibernation upgrade strategy where a completely new cluster is being used against the 'old' DB and cloud storage.

PR #487, opened 2021-03-15
bump: prometheus, prometheus-operator, prometheus CRDs (fixes #478)

### LINKED ISSUE #478: Errors when writing from non-system tenant prometheus
Steps to reproduce:
- Go to `https://<tenant>.<me>.opstrace.io/prometheus/`
- Enter a prometheus query into the text box, note that autocompletes actually work fine, implying that some endpoint is reachable to fill the autocomplete text
- Execute the query and get an error like `remote_read: remote server http://cortex-api.dev-tenant.svc.cluster.local:8080/api/v1/read returned HTTP status 401 Unauthorized: Authorization header missing` (my guess: Cortex expects `X-Scope-OrgID=tenantname` header?)

Workarounds:
- If you instead click `Try experimental React UI` in the upper right of the prometheus UI, the charts work fine in there. Maybe that's trying a different query that's successful?

nickbp:
WARNING: CLUSTER UPGRADE MAY REQUIRE UPDATING THIS CRD: `prometheuses.monitoring.coreos.com`

Prometheus on non-system tenants was failing to write metrics to cortex.

The root problem was that the tenant prometheus doesn't have an auth token to communicate via cortex-api. As a workaround we can have it talk directly to cortex instead.
However this required adding the `X-Scope-OrgID` tenant header to requests so that cortex knew what tenant to store against.
But then THIS required updating the prometheus stack (prometheus-operator, CRDs, and prometheus itself), because the `remoteWrite.headers` option was only added very recently and each component needed an update to support it.

Some other cleanup along the way:
- Some hardcoded docker images have been moved out to `docker-images.json`
- The process to update prometheus CRDs has been updated to allow generating against arbitrary versions of the prometheus operator (as of writing there were no CRDs generated for operator 0.46.0 yet), and to avoid editing the files by hand.

# Have you...


jgehrcke:
> the tenant prometheus should talk directly to cortex

@nickbp can you please briefly summarize here why the individual tenant prometheus now needs to read and/or write from/to Cortex?

Might be super obvious answers, but still incredibly helpful to write this down. :-).  In the spirit of "The system tenant Prometheus needs to be able to remote_write to Cortex to push the system metrics into Cortex".

nickbp:
It was mainly just because the per-tenant instances were seeing an error when trying to communicate via the `cortex-api` endpoint: They didn't have the credential that the system tenant has. As a result they were getting 401/auth errors when trying to `remote_write` data to `cortex-api`.

Switching to direct cortex was sort of a workaround, and if the non-system tenants are given an auth token then we should be able to switch back to communicating via `cortex-api` rather than directly to cortex. At the time we weren't sure if it made sense for the non-system tenants to a token generated for them so the fix was mainly to get the non-system metrics flowing again in the meantime.

jgehrcke:
> the per-tenant instances were seeing an error when trying to communicate via the cortex-api endpoint

Well, my question was about _why_ this communication became necessary, can you maybe re-read my question above? :-).

You write that

> to get the non-system metrics flowing again

but you might not know that this never needed to work before. :) The tenant Prometheuses had a very narrow purpose. Scraping targets and pushing data out to Cortex was _not_ their purpose before, and that seems to be what's new as far as I understood @spahl now (what's new, specifically, is that a tenant Prometheus scrapes the tenant's cloud metrics exporter(s)).

Might seem to be a boring insight to some of us; but in fact it's an incredibly important change when one wants to understand which component has which responsibilities.

> At the time we weren't sure if it made sense for the non-system tenants to a token generated for them

I would have strongly advocated against this. :)

nickbp:
Yes at the moment the non-system tenant prometheuses (prometheii?) only show the exporters in their scrape config so maybe this problem was specific/new with getting the collected exporter data sent upstream to cortex correctly. So maybe we just weren't exercising the non-system `remote_write` configs until the exporters (and their scrape configs) were added?

FWIW, with the exporters added I was seeing these symptoms for a non-system tenant (`dev`):
- The `dev` tenant Grafana instance wasn't seeing metrics. At the time I wasn't sure if this was expected.
- The `dev` tenant Prometheus instance logs showed ongoing errors about its `remote_write` configuration. It was configured to write metrics to `cortex-api`, where it was getting back HTTP 401 errors from `cortex-api` because it wasn't configured with the needed credentials.

Meanwhile, for the `system` tenant:
- The `system` Grafana instance could see metrics just fine.
- The `system` Prometheus instance had a similar (but different) `remote_write` configuration, where it _was_ being configured with a token to send to `cortex-api`. As a result it wasn't logging errors like the `dev` Prometheus was doing.

To illustrate the exact issue here's the code preceding this PR (`packages/controller/src/resources/monitoring/tenants/prometheus.ts`):
```
const remoteWrite: { url: string; bearerTokenFile?: string } = {
  url: `http://cortex-api.${getTenantNamespace(
    tenant
  )}.svc.cluster.local:8080/api/v1/push`
};

if (tenant.type !== "SYSTEM") {
  ruleNamespaceSelector = serviceMonitorNamespaceSelector = {
    matchLabels: {
      tenant: tenant.name
    }
  };
  // !!!NOTE!!! remoteWrite.bearerTokenFile isn't assigned here!
} else {
  // For the system tenant's Prometheus -- which scrapes Opstrace system
  // targets and pushes system metrics into Cortex via Prom's remote_write
  // protocol -- use the bearer_token_file mechanism to authenticate POST
  // requests towards the authenticator built into the Cortex API proxy.
  promSecrets = ["system-tenant-api-auth-token"];
  promBearerTokenFile =
    "/etc/prometheus/secrets/system-tenant-api-auth-token/system_tenant_api_auth_token";

  remoteWrite.bearerTokenFile = promBearerTokenFile;
  remoteRead.bearerTokenFile = promBearerTokenFile;
}

[... and then later in the Prometheus K8s resource, all tenants get remoteRead/remoteWrite configured: ...]
        remoteWrite: [remoteWrite],
        remoteRead: [remoteRead],
```

PR #488, opened 2021-03-16
ui: added "Remote cluster development" section to app/README.md for working with Hasura actions

terrcin:


terrcin:
Merging and ignoring CI as only readme chages with new yarn scripts that CI doesn't use.

PR #468, opened 2021-03-11
feat: add image processing and upload workflows

PatrickHeneise:
This adds two GitHub Action workflows to the repo:

1. Image Processing: new images in PRs will automatically be optimized for web usage
2. Asset upload: docs/assets will be uploaded to S3 to be used on the website

# Have you...

* [x] Discussed your change with a project contributor in an issue yet?
* [x] Added an explanation of what your changes do?
* [x] Thought about which docs need updating?

PR #490, opened 2021-03-16
controller: Generate controller graphql API alongside other copies

nickbp:
Instead of copying the file from packages/app in the controller build, just update both copies at once.
This avoids potential issues around the controller graphql code falling out of date vs the app graphql code.

As a future improvement, the graphql schema itself could be moved into a separate package. However it depends on Hasura, so we'd probably need to bring that along as well, at which point it starts entering "can of worms" territory in terms of updating build/dev tooling.

# Have you...


nickbp:
`Process Images` CI failure appears related to this PR: https://github.com/opstrace/opstrace/pull/468
The error message (`Error: Unsupported event pull_request_target (currently supported events include push, pull_request)`) sounds like a CI configuration issue rather than being related to the change, so I'll assume it can be ignored for now

PR #497, opened 2021-03-17
looker: cover local build in CI, work on metrics mode, and on dashboarding

jgehrcke:


jgehrcke:
the WAL fsync duration plot was empty as of a broken calculation. Now fixed:
Screenshot from 2021-03-17 17-51-06

PR #504, opened 2021-03-19
Work on Cortex limits config

jgehrcke:


packages/controller-config/src/types.ts
```diff
@@ -26,8 +26,12 @@ export const controllerConfigSchema = yup
...
       .required("must specify a target (gcp | aws)"),
     region: yup.string().required("must specify region"),
-    logRetention: yup.number().required("must specify log retention"),
-    metricRetention: yup.number().required("must specify metric retention"),
+    logRetentionDays: yup
+      .number()
+      .required("must specify log retention in number of days"),
+    metricRetentionDays: yup
+      .number()
+      .required("must specify metric retention in number of days"),
```
jgehrcke, line 34: Another change that should be a no-brainer on the one hand, but that will lead to funny upgrade effects when mixing old controller config with new controller version.

jgehrcke:
`line 41: cannot unmarshal !!str `8d` into time.Duration`

packages/controller-config/src/types.ts
sreis: We should, at some point, move the controller config to a proper Kubernetes CRD to be able to do these kinds of changes with [webhook conversion](https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definition-versioning/#webhook-conversion).

packages/controller/src/resources/cortex/index.ts
```diff
@@ -310,9 +317,30 @@ export function CortexResources(
     // limits_config
     // https://cortexmetrics.io/docs/configuration/configuration-file/#limits-config
     limits: {
+      // Delete blocks containing samples older than the specified retention
+      // period. Default is 0 (disabled). Add one day as a buffer here to be
+      // on the safe side. Note(JP): should this be the primary mechanism to
+      // delete old data from the corresponding cloud storage bucket? Should
+      // this replace the bucket lifecycle-based method?
+      compactor_blocks_retention_period: `${(metricRetentionDays + 1) * 24}h`,
```
sreis, line 325: Nice! This is [pretty recent](https://github.com/cortexproject/cortex/pull/3879) and it doesn't even [show up in the docs yet](https://cortexmetrics.io/docs/blocks-storage/compactor/#compactor_config). I guess we have to test it to be sure?

jgehrcke: this part of the "compactor config" is actually documented in the `limits_config` which I was going through : - )

https://cortexmetrics.io/docs/configuration/configuration-file/#limits_config

PR #514, opened 2021-03-23
installer: mark system-tenant-api-auth-token as managed

nickbp:
The `system-tenant-api-auth-token` secret lacked an `opstrace` annotation, so the controller would then log a (recently-added) notice about the secret not being a known/managed resource. Passing it via a `ResourceCollection` automatically configures the annotation. We also mark the secret as immutable since it's not explicitly included among the controller's expected resources.

As a minor change, the notice message is updated to include the type of the resource being logged (`Secret` in this example).

# Have you...


jgehrcke (approved):
Thanks!

PR #515, opened 2021-03-23
config: Fix credential validation via hasura actions

nickbp:
Fix issues around JSON (via hasura) vs YAML (via HTTP API) input. Had put this down for a bit until it was needed - and it sounds like it's now needed.

# Have you...


jgehrcke:
Have not really looked at code -- but can we / do we want to do some kind of active validation, doing a non-invasive API call using these credentials that can directly confirm many things? Structural validity, but also permissions? Being able to forward error messages to the end users? Sorry for the noise if that is already what we are doing : )

nickbp:
It isn't doing any of that yet, mainly just doing sanity checks like "if the exporter is for AWS, check that the credential it points to is also for AWS" but it does feel like this would be a good place to check that metrics are reachable.

PR #524, opened 2021-03-25
docs(readme): streamline by removing unnecessary content

triclambert:
Proposing a couple of immediate changes to the readme, removing extraneous text.


jgehrcke (approved):
Like that. Thanks!

Issue #517, opened 2021-03-24
Implement hasura actions for getting and setting ruler rules


nickbp:
The config API already has support for passing through the cortex HTTP endpoints directly but doesn't expose them via hasura actions. Should run the proposed graphQL API past Nahum as a sanity check before mergin

nickbp:
Had a discussion with Nahum about this API a few days ago. We agreed that it made sense to have a draft API via Hasura available for the UI to play with - then if any issues come up with the draft API's abilities we can revise the API later. The aforementioned #525 adds a set of APIs for listing/storing/deleting rules, that act largely as a configuration passthrough to the [Cortex APIs](https://cortexmetrics.io/docs/api/#set-rule-group).

PR #525, opened 2021-03-25
config: Hasura Actions for accessing cortex rules (#517)

nickbp:
Rules are under Rule Groups, and Rule Groups are under Namespaces. Each tenant can have one or more Namespaces.

The underlying cortex API has calls at each level, but to keep it simple for now, this just adds the following Hasura calls:
- `listRules(tenant)`: List all Rules for the tenant (across all namespaces/groups)
- `getRuleGroup(tenant, namespace, group_name)`: Get a single Rule Group
- `updateRuleGroup(tenant, namespace, group_yaml)`: Create/update a Rule Group - unlike with `updateAlertmanager` this does not support delete because the `group_yaml` payload contains the Rule Group name
- `deleteRuleGroup(tenant, namespace, group_name)`: Delete a Rule Group (Namespaces auto-delete when empty) - separate call from `update` per note above

ALSO, renames `AlertmanagerUpdateResponse` to `StatusResponse` to reflect its usage in Ruler calls as well

# Have you...

### DIFF
--- a/packages/app/metadata/actions.graphql
+++ b/packages/app/metadata/actions.graphql
@@ -9,32 +9,10 @@ enum ErrorType {
   VALIDATION_FAILED,
 }
 
-# getAlertmanager
-
-type Query {
-  getAlertmanager(tenant_id: String!): Alertmanager
-}
-
-type Alertmanager {
-  tenant_id: String!
-  config: String
-  online: Boolean!
-}
-
-# updateAlertmanager
-
-type Mutation {
-  updateAlertmanager(tenant_id: String!, input: AlertmanagerInput): AlertmanagerUpdateResponse
-}
-
-input AlertmanagerInput {
-  config: String!
-}
-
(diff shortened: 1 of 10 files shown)


packages/app/metadata/actions.graphql
```diff
@@ -9,61 +9,132 @@ enum ErrorType {
...
+# within each tenant, rules are under this hierarchy: namespace -> rulegroup -> rules
 
-  # Must be set if success=false
-  error_type: ErrorType
-  # Must be set if success=false
-  error_message: String
-  # Optional if success=false, message returned by underlying service
-  error_raw_response: String
+# Returns all rules across all namespaces/rulegroups. Example:
+type Query {
```
nickbp, line 50: This is the "meat" the the PR: the added GraphQL queries themselves.

One complication I had was that the RuleGroups are submitted with the name in the YAML payload, so I tried to make the structure work around that. See also https://cortexmetrics.io/docs/api/#set-rule-group

Issue #326, opened 2021-02-03
Build "hibernation" upgrade


sreis:
Part of the goal to implement a reliable, testable upgrade strategy to deliver value to users as soon as possible.

# Current Behavior

Upgrades don't handle infrastructure changes and don't handle controller configuration schema changes. The rapid development pace is impacted when handling upgrades since all scenarios (testing, etc...) need to be accounted for during feature development.

# Proposed Behavior

As discussed previously, we will call this new upgrade behavior "hibernation" until a new or better name comes up. 

The proposed solution is to stop all the Kubernetes deployments and destroy the underlying cloud infrastructure *but keep all the data in the cloud storage* (ex, S3 in AWS). Then proceed to rebuild the cloud infrastructure and start all the Kubernetes deployments and attach them to the cloud storage previously used.

The major downside of this approach is that the maintenance window increases. The maintenance window will be the time to destroy and recreate the cloud infrastructure and all Kubernetes deployments, plus the time to propagate the new DNS entries. 

##  Open questions

* To rebuild the cluster, we need to handle any cluster configuration changes that might happen between releases.
* During the cluster rebuild, the DNS entries will require an update. This update will take time to propagate and potentially bring up some issues.
    * What happens to the certificate request during this timeframe? With #151 and #172, we know the certificate request process has issues with retry behavior.
    * What is the behavior of the scraping clients during this timeframe?
* The DNS service, currently, runs a loop every 60 seconds to create and/or destroy DNS entries. After opstrace/dns-service#8 and opstrace/dns-service#9  are there any other issues we need to account for?

jgehrcke:
Adding a bit to this ticket again. "hibernation" would in practice be a special way to tear down an Opstrace instance. Related ticket and prior art: https://github.com/opstrace/opstrace/issues/27

Motivated by https://github.com/opstrace/opstrace/issues/1169 I want to add two quick ideas:

- `opstrace hibernate`, maybe with an option to leave anything ELB and DNS intact so that one does not needlessly lose time if this is being used in the context of an upgrade.
- `opstrace create ... ` (not a new command!) maybe with an option tied to this overall idea of hibernation wakeup; so that this can error out if the prerequisites / state is not found.

In `create` we should then add explicit code that inspects the state of S3, database etc to see if a migration needs to be done (would very rarely be the case).

jgehrcke:
related: https://github.com/opstrace/opstrace/issues/27

Issue #498, opened 2021-03-17
Update out-of-date CRDs automatically in controller


nickbp:
We've recently updated the CRDs for prometheus in order to support adding custom HTTP headers to remoteWrite (which our non-system tenants use for sending metrics to cortex). This feature was only added in the latest version of prometheus, so running clusters will need a CRD update in order to support custom HTTP headers in remoteWrite.

The controller currently doesn't automatically update CRDs that already exist, however it will automatically recreate CRDs if they are manually deleted (and thereby update them to the new version hardcoded in the controller).

From Nick's previous experience with setting up CRDs in Helm, its possible that we'll need slightly different logic for updating CRDs. For example, maybe we just check for CRD _updates_ once on controller startup, then in the reconciliation loop only do creates for missing CRDs? This wouldn't protect against a user manually editing a CRD behind our back though. Seb also knows a lot about CRDs.

MatApple:
@nickbp can we replace the internals of this (https://github.com/opstrace/opstrace/blob/6e9a9c4894c73d401f1a4790a2e597058ce983da/lib/kubernetes/src/kinds/CustomResourceDefinition.ts#L217) method with a:
```
await this.delete();
return this.create();
```
The CRs should still be fine, but in the background we'll essentially delete the existing CRD and create a new CRD. It's blunt and there is likely a more nuanced way but would be interesting to see if this simplicity would work for now.

PR #526, opened 2021-03-25
feature: handle older cluster config versions

sreis:
Proof of concept code for #326.

The goal is to be able to handle cluster configuration files from previous versions.

The general idea in this implementation is to introduce type aliases that point to the latest version of the cluster config schema. This way we avoid big code changes when changing the cluster config schema.

In this example, we have two versions for the cluster config schema and associated types. V1 will be the base version. V2 is the latest version in the main branch. The difference between V1 and V2 is renaming `log_retention` and `metric_retention` to `log_retention_days` and `metric_retention_days`. We'll address the `tenant_api_authenticator_pubkey_set_json` separately.

Thi is the bulk of the code to define the schemas:

```ts
// Import types from previous versions.
import {
  ClusterConfigFileSchemaTypeV1, ClusterConfigFileSchemaV1
} from "./schemasv1";

// Import types for the most recent version
import {
  AWSInfraConfigSchemaV2,
  GCPInfraConfigSchemaV2,
  ClusterConfigFileSchemaV2,
  RenderedClusterConfigSchemaV2,
  ClusterConfigFileSchemaTypeV2,
  RenderedClusterConfigSchemaTypeV2
} from "./schemasv2";

//
// consts pointing to the latest schema versions
//
export const LatestAWSInfraConfigSchema = AWSInfraConfigSchemaV2;
export const LatestGCPInfraConfigSchema = GCPInfraConfigSchemaV2;
export const LatestClusterConfigFileSchema = ClusterConfigFileSchemaV2;
export const LatestRenderedClusterConfigSchema = RenderedClusterConfigSchemaV2;

//
// type aliases pointing to the latest version types
//
export type LatestRenderedClusterConfigSchemaType = RenderedClusterConfigSchemaTypeV2;
export type LatestClusterConfigFileSchemaType = ClusterConfigFileSchemaTypeV2;

//
// upgrade cluster config from V1 to V2
//
function V1toV2(ucc: ClusterConfigFileSchemaTypeV1): ClusterConfigFileSchemaTypeV2 {
  // unpack v1 cluster config and save the rest the of the fields in restConfig 
  const { log_retention, metric_retention, ...restConfig } = ucc;
  // return V2 object
  return {
    ...restConfig,
    log_retention_days: log_retention,
    metric_retention_days: metric_retention
  }
}

//
// function that takes any user cluster config and upgrades it to the latest
// version if necessary.
//
export function upgradeToLatest(ucc: any): LatestClusterConfigFileSchemaType {
  //
  // try parsing the file into the latest version
  //
  if (LatestClusterConfigFileSchema.isValidSync(ucc, { strict: true })) {
    // validate again, this time "only" to interpolate with defaults, see
    // https://github.com/jquense/yup/pull/961
    log.debug("got latest cluster config file version");
    return LatestClusterConfigFileSchema.validateSync(ucc);
  }

  //
  // it failed, so try parsing it using the V1 version
  //
  if (ClusterConfigFileSchemaV1.isValidSync(ucc, {strict: true})) {
    log.debug("got v1 cluster config file");
    return V1toV2(ClusterConfigFileSchemaV1.validateSync(ucc));
  }

  // Possible user error. Parse again to throw a meaningful error message.
  LatestClusterConfigFileSchema.validateSync(ucc, {strict: true});
}
```

[Relevant unit tests](https://github.com/opstrace/opstrace/compare/sreis/poc-config-schema-versioning?expand=1#diff-50cd2beb1c63ae41ef5adbc5305f9d8c37e64028a47a093874bbf5982d266d9fR171-R264) to show it works.


MatApple:
This is a great solution straight off the bat that is easy to reason about. Is the intention that we'd over time add v3, v4, etc when we have to change more configs in the future?

sreis:
> Is the intention that we'd over time add v3, v4, etc when we have to change more configs in the future?

@MatApple Yes. I don't think we can avoid that. In this scenario, a future v3 change would imply creating a `schemav3` typescript file with the new schemas.

sreis:
Added the same strategy used to handle cluster configuration to version the controller configuration.  This will hopefully help us have a "soft" upgrade strategy (only upgrades the Kubernetes deployments and resources) while we work on the "hibernation" feature (handles infrastructure upgrades).

The controller V1 config schema is set to the one from this [this PR](https://github.com/opstrace/opstrace/pull/413). The `V1toV2` function that handles configuration schema upgrades tries to handle the  `tenant_api_authenticator_pubkey_set_json` with code from #457.

After creating a cluster from commit `c2dbd46` I upgraded it using code in this PR. This upgrade path requires manually creating an S3 bucket for the alert manager. This step was done manuallly.

Cortex and Loki were upgraded successfully as well as the internal APIs. This means the issue reported in #450 was handled successfully using the controller config version strategy in this PR.

The `alertmanager` pods were crash looping with:
```
level=error ts=2021-03-29T15:35:45.358169534Z caller=cortex.go:424 msg="module failed" module=alertmanager err="invalid service state: Failed, expected: Running, failure: failed to list users with alertmanager configuration: Access Denied"
```

Used this command to check we can reuse the auth token:
```bash
curl -i -k -H "authorization: Bearer $(cat sreis088/tenant-api-token-default)" https://loki.default.sreis088.opstrace.io/loki/api/v1/label
```

But found this error message in the loki API logs after I got a 401:
```
time="2021-03-29T15:43:39Z" level=info msg="jwt verification failed: kid not set in auth token, fallback key not set, consider token invalid (unverif. claims: &{opstrace-cluster-sreis088 1932601953  1617025953 opstrace-cli 0 tenant-default})"
```

Some work still left to do:
* Fix the alertmanager upgrade path
* Check if `generateAuthenticatorPubkeySetJSON` from #457 is working as expected

sreis:
> * Fix the alertmanager upgrade path

Replying to myself. It requires the policy defined [here](https://github.com/opstrace/opstrace/pull/441/files#diff-af02c9203600e6fcc9c49a38669a21699af5300e5cb1a9720fd82d9402ca407aR433).

sreis:
> > * Fix the alertmanager upgrade path
> 
> Replying to myself. It requires the policy defined [here](https://github.com/opstrace/opstrace/pull/441/files#diff-af02c9203600e6fcc9c49a38669a21699af5300e5cb1a9720fd82d9402ca407aR433) to be added to the EKS nodes.

Instead of doing it manually, I decided to try and automate it with code. 

Since the installer/create command is idempotent, we can reuse the code that ensures the cloud provider is set up. This means any new infrastructure is created, for example, the new S3 bucket and S3 policies from #441. This upgrade path doesn't handle the future scenario where we might remove some cloud provider resource. In another scenario, when a cloud provider resource is renamed, it'll create a new one, and the old one will need to be deleted manually. We can discuss doing it as part of the upgrade when that day comes.

To reuse the installer/create code required adding a flag to pass the cluster config. Currently, the upgrade command doesn't support changing the cluster configuration for the upgrade. 

I tested this locally, and it resolved the alertmanager crash loop with missing S3 bucket.

TODO:
* Check if generateAuthenticatorPubkeySetJSON from #457 is working as expected

sreis:
> Check if generateAuthenticatorPubkeySetJSON from #457 is working as expected

It was but it required setting `API_AUTHTOKEN_VERIFICATION_PUBKEY` env variable for the internal APIs to fallback using that key for the old tokens.

Changed the `V1toV2` function that handles controller configuration schema upgrades to also set the previous `data_api_authn_pubkey_pem` field in the controller config. When this field is set the internal APIs will read them from the 
`API_AUTHTOKEN_VERIFICATION_PUBKEY` env var and proceed to validate the old auth token.

sreis:
Using this PR, the upgrade test finished locally and I was able to reuse the tenant API tokens.

Found the following error in the opstrace controller logs though.

```
2021-03-30T14:22:19.246Z info: api err during updateResource for compactor(ns: loki): 
{"response":{"statusCode":422,"body":{"kind":"Status","apiVersion":"v1","metadata":{},"status":"Failure","message":"Deployment.apps \"compactor\" is invalid: spec.strategy.rollingUpdate: Forbidden: may not be specified when strategy `type` is 'Recreate'","reason":"Invalid","details":{"name":"compactor","group":"apps","kind":"Deployment","causes":[{"reason":"FieldValueForbidden","message":"Forbidden: may not be specified when strategy `type` is 'Recreate'","field":"spec.strategy.rollingUpdate"}]},"code":422}, <SNIP>
```

sreis:
> Found the following error in the opstrace controller logs, though.
> 
> ```
> 2021-03-30T14:22:19.246Z info: api err during updateResource for compactor(ns: loki): 
> {"response":{"statusCode":422,"body":{"kind":"Status","apiVersion":"v1","metadata":{},"status":"Failure","message":"Deployment.apps \"compactor\" is invalid: spec.strategy.rollingUpdate: Forbidden: may not be specified when strategy `type` is 'Recreate'","reason":"Invalid","details":{"name":"compactor","group":"apps","kind":"Deployment","causes":[{"reason":"FieldValueForbidden","message":"Forbidden: may not be specified when strategy `type` is 'Recreate'","field":"spec.strategy.rollingUpdate"}]},"code":422}, <SNIP>
> ```

This occurs because, in #426, the strategy changed from the default `RollingUpdate` to Recreate. This means, in an upgrade to this version, the strategy field is populated with the default values for the `RollingUpdate` strategy.

sreis:
Requires #498 to handle the CRD upgrade for the Prometheus operator.

sreis:
Recreated the loadtest cluster from version opstrace-c2dbd46e and added some load to it to run an upgrade test.

The upgrade went fine except for the CRD upgrades. I added code to do a replacement but it seems to hit a 402 due to missing resourceVersion.

sreis:
Another issue that I found is that `isDeepStrictEqual` doesn't handle CRD version schemas very well. Tested a workaround that checks if the CRD has any new versions and it seems to do the trick for now.

```
export const hasCustomResourceDefinitionChanged = (
  desired: CustomResourceDefinitionType,
  existing: CustomResourceDefinitionType
): boolean => {
  const desiredNames = desired.spec.spec.versions.map((v) => v.name);
  const existingNames = existing.spec.spec.versions.map((v) => v.name);

  if (!isDeepStrictEqual(desiredNames, existingNames)) {
    log.debug(`CRD ${desired.spec.metadata?.name} requires update with new versions`);
    return true;
  }
  return false;
};
```

Also learned that CustomResourceDefinition doesn't have the [GuaranteedUpdate strategy ](https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/apiextensions-apiserver/pkg/registry/customresource/strategy.go) which prevents us from using `this.api.replaceCustomResourceDefinition` method call since a PUT request for a CRD is not handled properly. We have to manually delete and then recreate the CRD.

```
export class CustomResourceDefinition extends K8sResource {
...
  update(): Promise<{
    response: IncomingMessage;
    body: V1CustomResourceDefinition;
  }> {
    const gracePeriodSeconds = 0;
    return this.api.deleteCustomResourceDefinition(
      this.name,
      undefined,
      undefined,
      gracePeriodSeconds
    ).then(async () => {
      await sleep(1.0);
      return this.create();
    });
  }
```

sreis:
To update a CRD we first need to check if it requires an update. Initially, I used `isDeepStrictEquals` but it doesn't handle arrays with complex objects very well. This meant the controller would enter a loop always trying to update the same CRD. 

To workaround this limitation we now set an annotation on the CRD with the controller version using the opstrace/buildinfo package. To check if the CRD requires an update we check if the annotations match. When a new controller is installed, for example, during an upgrade, the annotation will be updated. The new controller will update the CRD if it contains any changes in the patch request. Check https://github.com/opstrace/opstrace/pull/526/commits/4764302e992a2f9fb5acda031481744ddae483b7 for more details.

sreis:
Set up the `loadtest` cluster with the following config using Opstrace CLI version c2dbd46e 

```
> cat config.yaml 
tenants:
  - default
env_label: loadtest
node_count: 5
cert_issuer: letsencrypt-prod
log_retention_days: 30
metric_retention_days: 60
aws:
  instance_type: r5.xlarge
  region: us-east-2
```
(picked us-east-2 because the latency is better from where I live).

And set up a workload cluster sending 2.5M metrics using avalanche. (Details to be added to the handbook).

Afterward, upgraded the loadtest cluster to this PR build successfully :tada: 

Screenshot from 2021-04-09 15-17-36

Screenshot from 2021-04-09 15-17-14

sreis:
Ingester memory usage seems to have doubled.

Before
Screenshot from 2021-04-09 15-19-59

After
Screenshot from 2021-04-09 15-19-50

sreis:
Small reduction ~2 hours after the upgrade.

Screenshot from 2021-04-09 17-25-07

Screenshot from 2021-04-09 17-27-14

Looks like ingesters did some garbage collection

Screenshot from 2021-04-09 17-25-45

sreis:
Updated PR to fix merge errors and I'll merge it if tests are green.

sreis:
I was able to replicate https://github.com/opstrace/opstrace/pull/526#issuecomment-816838980 by triggering a rolling restart of the cortex ingesters. 

Didn't find any issues in cortex so I figured it would be better to get some golang profiling info to create a ticket. After I got the profiling done I searched for loadWAL and found what we were looking for in https://github.com/cortexproject/cortex/issues/3817. The suggestion was to try the latest cortex master.

Diff of memory heap profile after a statefulset restart of cortex version 1.8.0-rc0 
profile001

Diff of memory heap profile after rollout to latest cortex master (tested with upgrade to #553) 
profile002

PR #545, opened 2021-04-08
chore: update default config values to make config more friendly

### LINKED ISSUE #544: make minimal config more friendly by updating default values
Let's shorten the requirements for a "default" config by providing `3` as the default for node_count.  Let's also better align `cert_issuer` with our value proposition, making the default `letsencrypt-prod` (ensuring CI uses `letsencrypt-staging`).

triclambert:
Resolves #544.

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,14 +20,14 @@ Install Opstrace in your own cloud account with our [quick start](https://go.ops
 ```bash
 cat <<EOF > config.yaml
 tenants:
+  - dev
   - staging
   - prod
-env_label: quickstart
-node_count: 3
+env_label: tryme
 cert_issuer: letsencrypt-prod
 EOF
 
-opstrace create aws me -c config.yaml
+opstrace create aws <choose_a_name> -c config.yaml
 ```
 
 ## Community
--- a/README.md
+++ b/README.md
@@ -30,6 +30,8 @@ EOF
 opstrace create aws <choose_a_name> -c config.yaml
 ```
 
+(Note: `gcp` is also supported.)
+
 ## Community
 
 Authentic collaboration in a community setting is important to us.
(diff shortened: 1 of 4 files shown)


sreis (approved):

jgehrcke:
From #544:

> Let's shorten the requirements for a "default" config by providing 3 as the default for node_count.

I would have loved to see more elaborate arguments here, because the outcome of this reasoning is the opposite of previous reasoning. Previous reasoning is documented here: https://github.com/opstrace/opstrace/blob/main/docs/references/cluster-configuration.md#required-parameters

> You should explicitly choose the number of machines your Opstrace cluster is comprised of, which is why this is a required parameter for now.

The cluster configuration interface is one of the most important user-facing interfaces in our software engineering universe. The most important part of a user-facing interface is its documentation. The _reference documentation_. It _specifies_ the behavior. The implementation then needs to follow the specification. It's therefore surprising that we didn't change the specification in sync with this implementation change. That should be caught by review!

Also, such a change should be tested by CI. The current CI config still specifies `node_count`.

Consistency matters! Prior state was consistent: parameter was required, was documented / specified as required, and explicitly tested like that. An important triangle.

Current state is not consistent anymore.

triclambert:
I knew you'd review it after you returned from vacation.  Fixing now.  Thanks.

triclambert:
Regarding the reasoning for the change:  I know it is exactly the opposite from the previous decision.  That is intentional.  There are several reasons for this, including the fact that we rarely (almost never) change this value.  At the moment, we are optimizing for an onboarding flow, and this parameter does not provide any value during that flow.  Additionally, one might argue that it leaks implementation details; in the future, we probably care to plan for capacity in some manner other than raw node counts.

I'd be happy to discuss this in more detail synchronously.

PR #549, opened 2021-04-10
clean up after #544/#545, make some additional edits along the way

triclambert:
@jgehrcke rightly identified that I forgot to update all of the documentation in #545.  Fixing `cluster-configuration.md` here.  

Since I'm in the documentation, I make some additional changes for grammar and content.

### DIFF
--- a/docs/references/cluster-configuration.md
+++ b/docs/references/cluster-configuration.md
@@ -4,7 +4,7 @@ description: Parameter reference
 
 # Cluster configuration
 
-For setting up an Opstrace cluster, one has to provide a corresponding configuration document in YAML format.
+When installing Opstrace, one has to provide a corresponding configuration document in YAML format.
 A minimal variant might look like this:
 
 ```yaml
@@ -13,11 +13,11 @@ tenants:
     - ci
 ```
 
-The following sections list individual configuration parameters and aim to provide detailed specification for their purpose and meaning.
+The following sections list individual configuration parameters and aim to provide a detailed specification for their purpose and meaning.
 
 At a high level, there are two different types of parameters:
 
-* required ones (you have to set those for creating an Opstrace cluster)
+* required ones
 * optional ones (but please do not assume sane defaults yet :-))
 
 **Note:**  as of today, the cluster configuration cannot be updated at runtime.
@@ -73,8 +73,7 @@ aws:
         - AWSReservedSSO_AdministratorAccess_133333abc3333337
 ```
 
-Use the `eks_admin_roles` parameter (expected to be an enumeration of strings) to define individual AWS IAM roles that should have administrator access to the underlying EKS cluster, via e.g. the [EKS console](https://aws.amazon.com/blogs/containers/introducing-the-new-amazon-eks-console/).
-
+Use the `eks_admin_roles` parameter (an enumeration of strings) to define individual AWS IAM roles that should have administrator access to the underlying EKS cluster, via e.g. the [EKS console](https://aws.amazon.com/blogs/containers/introducing-the-new-amazon-eks-console/).
 
 **Note:**
 we develop and test mainly with `region` set to `us-west-2`.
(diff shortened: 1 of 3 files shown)


docs/references/cluster-configuration.md
```diff
@@ -40,9 +39,11 @@ tenants:
     - ci
 ```
 
+## Optional parameters
+
 ### `node_count`
 
-The number of machines this Opstrace cluster is supposed to be comprised of.
+The number of underlying instances to use.
```
jgehrcke, line 46: Let's revert that part of the change that removes the notion of "machines". 

"instance" is a highly generic term, and not everyone immediately associates it with the number of VMs/nodes/machines.

Sorted by clarity:
1. machine count / node count
2. instance count

docs/references/cluster-configuration.md
```diff
@@ -98,13 +94,8 @@ gcp:
     zone_suffix: a
 ```
 
-
 Note: the example above shows the defaults.
 
-**Warning 😈:**
-we develop and test with region `us-west2` and have not validated other regions yet.
```
jgehrcke, line 105: maybe let's still keep this as a "we don't regularly test other regions"?

docs/references/cluster-configuration.md
```diff
@@ -127,18 +118,18 @@ The label name will be `env`, and the value will be what you provide here.
 
 *Value type:* boolean
 
-By default, authentication proof is required when accessing the HTTP API for pushing or querying logs/metrics data.
-This flag can be used to disable said authentication mechanism, allowing for unauthenticated clients to write or read data.
+By default, authentication proof is required when accessing the HTTP API for pushing or querying data.
+This flag disables said authentication mechanism, allowing for unauthenticated clients to write or read data.
```
jgehrcke, line 122: Let's revert this: this is still a boolean parameter, so it can be _used to disable_, but whether or not it really disables depends on the choice of the value.

jgehrcke: (`data_api_authentication_disabled: false` is the default)

docs/references/cluster-configuration.md
triclambert: Aligning this with the AWS text.

docs/references/cluster-configuration.md
triclambert: I'm trying to eliminate the use of the passive voice.  PTAL.

docs/references/cluster-configuration.md
jgehrcke: As long as we pick from (1) we're good, I thought I had said that :D

docs/references/cluster-configuration.md
```diff
@@ -127,18 +122,18 @@ The label name will be `env`, and the value will be what you provide here.
 
 *Value type:* boolean
 
-By default, authentication proof is required when accessing the HTTP API for pushing or querying logs/metrics data.
-This flag can be used to disable said authentication mechanism, allowing for unauthenticated clients to write or read data.
+By default, authentication proof is required when accessing the HTTP API for pushing or querying data.
+This flag enables or disables said authentication mechanism, allowing for unauthenticated clients to write or read data.
```
jgehrcke, line 126: thanks

jgehrcke: Whereas the logical consistency between the first part of the sentence and the second part is now disrupted:

> allowing for unauthenticated clients to write or read data.

jgehrcke: The passive voice improvement here should not come with logical regressions : ). You can get this right!

triclambert: Whoops.  Fixing by shortening the sentence.  Alternatively, we can abandon any changes here.  It's not critical.

triclambert: Whoops.  I'll just revert the change—it's not that important.

PR #551, opened 2021-04-12
controller: Fix deletion of the last credential or exporter (empty array case)

nickbp:
I thought this was working fine before, maybe I just didn't notice that the last one wasn't being deleted in practice.

# Have you...

### DIFF
--- a/packages/controller/src/reducers/graphql/credentials.ts
+++ b/packages/controller/src/reducers/graphql/credentials.ts
@@ -127,7 +127,7 @@ export function startInformer(
       })
       .subscribe({
         next: res => {
-          if (res.data?.credential && res.data.credential.length > 0) {
+          if (res.data?.credential) {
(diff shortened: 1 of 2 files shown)


nickbp:
The failing gcp test was the cluster failing to come up, meanwhile the equivalent AWS test passed. Given that this is a small bug fix (that was encountered and verified via manual testing) I think it's okay to merge as-is.

nickbp:
To be clear what the actual issue was: I was trying out the new blackbox exporter support (in #550) and noticed that when I sent the http delete command to delete the last exporter, the exporter was being deleted from graphql but the controller wasn't deleting the corresponding Deployment from the cluster. If the controller was restarted then it would delete the Deployment. At first I thought there was an issue with the GraphQL subscription client itself before finally noticing this size check. After removing the check I saw that the Deployment was being deleted correctly.

PR #552, opened 2021-04-12
ci: upgrades - add path to cluster config

sreis:
Fix missing flag in upgrade command.


sreis:
Upgrade test is green https://buildkite.com/opstrace/upgrade-tests/builds/390#ef6c427b-0528-4ad0-97cc-6f6f1522b167

PR #557, opened 2021-04-14
chore: handle failed certificate

sreis:
Been running CI off this branch for a while now in the hopes of catching a certificate in a failed state (aka issue #151) but have been unlucky so far.

It should be safe to merge, so let's do it (if CI is green) since the issue seems to occur in waves, and I'm always too late. The added logs should be enough to help us manage this long-standing issue.

### DIFF
--- a/packages/installer/src/certificates.ts
+++ b/packages/installer/src/certificates.ts
@@ -1,0 +1,67 @@
+/**
+ * Copyright 2021 Opstrace, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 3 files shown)


sreis:
Prediction: now that I've opened a PR it'll catch a failed certificate :joy:

packages/installer/src/certificates.ts
```diff
@@ -0,0 +1,75 @@
...
+} from "@opstrace/kubernetes";
+import { log } from "@opstrace/utils";
+
+//
+// Note: workaround for https://github.com/opstrace/opstrace/issue/151 until
+// https://github.com/jetstack/cert-manager/issues/3594 is resolved.
+//
+// If the https-cert certificate is in the failed state then delete it. The
+// controller will recreate the resource thereby triggering a restart of the
+// certificate request process.
```
jgehrcke, line 29: Yeeeeess, thank you for this code comment. Now it took me 10 seconds to understand what the idea here is. Love that.

jgehrcke (commented):
Thank you @sreis -- was about time for building our 'manual workaround' into the product, after all of those CI /devprod annoyances. This is great.

MatApple:
🎉 thanks @sreis

PR #561, opened 2021-04-14
chore: exit early if cluster name is already taken

sreis:
Exit installer early if cluster name is already taken and you don't own it.

### DIFF
--- a/lib/dns/src/opstrace.ts
+++ b/lib/dns/src/opstrace.ts
@@ -359,6 +359,11 @@ export class DNSClient {
             die("DNS setup failed with a permanent error (403 HTTP response).");
           }
 
+          // exit if we're trying to use a cluster name that is already taken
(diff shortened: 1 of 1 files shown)


sreis:
GCP run hit #151

lib/dns/src/opstrace.ts
```diff
@@ -359,6 +359,11 @@ export class DNSClient {
             die("DNS setup failed with a permanent error (403 HTTP response).");
           }
 
+          // exit if we're trying to use a cluster name that is already taken
+          if (e.response?.statusCode === 409 && method === "POST") {
+            die("DNS setup failed with a permanent error, please choose a different cluster name");
```
jgehrcke, line 364: let's maybe indicate the conflict/"name already in use" error in the error message. Of course the solution is good to have as in "choose a different one", but exposing the actual problem is also good UX in this case. :)

PR #567, opened 2021-04-16
ui: disable extra caching to see if causing login issues after cluster upgrade

terrcin:


packages/app/src/client/index.tsx
```diff
@@ -47,4 +46,8 @@ if (process.env.NODE_ENV !== "production") {
   );
 }
 
-serviceWorker.register();
+// 2021.04.16 NTW: this service worker does additional caching for progressive asset loading. After each cluster
+// upgrade users can't login and much clear their caches to get the login working again. Current theory is that
+// this serviceWorker might be creating a bad cache situation, so disabling it to test that.
+// Also not really providing any benefit at the moment.
```
jgehrcke, line 52: Thank you for a comment like this. I'm always a huge fan of these -- this will make things easier for at least one of us down the road : )

PR #573, opened 2021-04-19
fix: handle auto scaling group upgrade without recent activity

### LINKED ISSUE #564: upgrades: handle auto scaling group with no recent activity
During an upgrade [the AWS auto scaling group check ](https://github.com/opstrace/opstrace/blob/main/lib/aws/src/autoScalingGroup.ts#L126-L129) failed because the ASG didn't have any activity recently.

Check how we can handle this during an upgrade.

sreis:
Close #564 

I checked [Terraform code](https://github.com/hashicorp/terraform-provider-aws/blob/068ebd8bb5a703e92342ba69406746c993444033/aws/resource_aws_autoscaling_group_waiting.go#L22) to see how they handle this scenario. They [compare state](https://github.com/hashicorp/terraform-provider-aws/blob/068ebd8bb5a703e92342ba69406746c993444033/aws/resource_aws_autoscaling_group.go#L1057-L1060) using their internal data structures instead of relying on cloud provider state. This way they only trigger an AutoScaling group update if there are any desired changes.

I ended up settling on exit if the AutoScaling group activities are empty but the instance count is at the desired capacity.

### DIFF
--- a/lib/aws/src/autoScalingGroup.ts
+++ b/lib/aws/src/autoScalingGroup.ts
@@ -124,6 +124,12 @@ class ASGRes extends AWSResource<
     );
 
     const scalingacts: AutoScaling.Activities = await this.getScalingActivities();
+
+    if ((scalingacts.length === 0) && (instanceCount === this.expectedInstanceCount)) {
+      log.info("%s setup: nothing to do, no scaling activities seen recently and already at desired capacity", this.rname);
+      return asg;
+    }
+
     if (scalingacts.length === 0) {
       log.info("%s setup: no scaling activities seen yet", this.rname);
       return false;


lib/aws/src/autoScalingGroup.ts
```diff
@@ -124,6 +124,12 @@ class ASGRes extends AWSResource<
     );
 
     const scalingacts: AutoScaling.Activities = await this.getScalingActivities();
+
+    if ((scalingacts.length === 0) && (instanceCount === this.expectedInstanceCount)) {
+      log.info("%s setup: nothing to do, no scaling activities seen recently and already at desired capacity", this.rname);
+      return asg;
+    }
+
```
jgehrcke, line 132: There is a comment down below which says

```
// Desired state: ASG status `successful`, and expected instance count.
```

this needs updating, because it's not true anymore. 

Seemingly, the new definition of the desired state is simply "seeing expected instance count".

In theory we don't ever need to look at the last scaling activity.

It's probably nice to leave the code in here, for debugging.

sreis: > In theory we don't ever need to look at the last scaling activity.

We need it to cover the initial `create` scenario and check if the ASG reached the desired instance count. It's also how Terraform is doing it, see [this](https://github.com/hashicorp/terraform-provider-aws/blob/068ebd8bb5a703e92342ba69406746c993444033/aws/resource_aws_autoscaling_group_waiting.go#L81-L96).

> this needs updating, because it's not true anymore.

I don't think it needs updating since it covers the scenario when the ASG is created.

jgehrcke: The maybe not so obvious part... it's noteworthy that `scalingacts.length === 0` is hit only when six weeks past since initial create. As documented, scaling activities for the last six weeks are returned -- that is why this new code path can safely be considered to take effect only for upgrades. And that is why you're right, the code path and success criterion for the initial create did _not_ change. Can maybe tune the code comments later to make this more obvious.

PR #584, opened 2021-04-22
ci: pin base cluster version for upgrade test

sreis:
Start testing the upgrade path from a pinned version. The base version is the one being deployed to our collection cluster and is updated every other week. Will continue to automate this process.

Required env vars are set in https://buildkite.com/opstrace/upgrade-tests/steps.


sreis:
This only affects the upgrade test pipeline. I ran the upgrade test build separately, but one of the tests failed with a flake. Merging.

Issue #556, opened 2021-04-14
Azure metrics exporter


nickbp:
Look into implementing an Azure exporter, similar to the current AWS/GCP ones.

Need to look into credential requirements, the exporter has different instructions based on whether "managed identities" are being used. The instructions seem to imply that metrics need to be fetched from within an Azure-hosted VM or "app"? Hopefully there's a way to use credentials to query into metrics from the outside... See also: https://github.com/RobustPerception/azure_metrics_exporter#azure-account-requirements

Exporter itself would be:
- https://github.com/RobustPerception/azure_metrics_exporter
- https://hub.docker.com/r/robustperception/azure_metrics_exporter/tags - only has a single `latest` tag, may want to pin to the exact SHA just in case.

nickbp:
From reading the azure exporter readme linked above again, it sounds like getting a credential involves creating an "App" and give the App permission to access metrics. So in principle we should be good there.

See also:
- Azure docs: https://docs.microsoft.com/en-us/azure/active-directory/develop/howto-create-service-principal-portal#create-a-new-application-secret
- Instructions for another monitoring service that appears to support extracting Azure metrics (and faces the same need for credentials to get in): https://www.appdynamics.com/community/exchange/extension/azure-monitoring-extension/

nickbp:
FYI: There's also https://github.com/sylr/prometheus-azure-exporter which is recently active but only has one (human) contributor, whereas the RobustPerception one linked earlier has been less active recently but has had more contributors. So it's not totally cut and dry but I think it makes sense to try the RobustPerception one first and see how it goes.

nickbp:
Okay, one big count against the RobustPerception exporter is that it wants you to just put your Azure credentials directly into your scrape config file, with no provision for supplying them via a separate secret channel: https://github.com/RobustPerception/azure_metrics_exporter/blob/40606b2ae091e57e08376c0214427e7ea8f71bdc/config/config.go#L17

A workaround for this could be to inject credentials into the yaml configuration when starting the exporter, via e.g. an initContainer. Not a huge deal but still not confidence inspiring.

I checked the sylr exporter and it does have built-in env secrets support so it may end up being the better route after all: https://github.com/sylr/prometheus-azure-exporter/blob/master/pkg/config/config.go#L46

nickbp:
[...]

PR #590, opened 2021-04-23
test-remote: Add test for azure exporter (#556)

nickbp:
The Azure exporter fails to start when provided with bogus credentials, so we just check its logs to validate that it was deployed and that it failed to query Azure as expected. In comparison, when other exporters are faced with invalid credentials, they tend to instead expose a metric with a nonzero error count.

# Have you...

### DIFF
--- a/test/test-remote/test_alerts.ts
+++ b/test/test-remote/test_alerts.ts
@@ -63,8 +63,6 @@ import {
   mtimeDeadlineInSeconds,
   rndstring,
   sleep,
-  waitForCortexMetricResult,
-  waitForPrometheusTarget,
   CLUSTER_BASE_URL,
   CORTEX_API_TLS_VERIFY,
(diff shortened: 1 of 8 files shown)


nickbp:
Looks like the `system` tenant azure exporter test is timing out on seeing the expected azure exporter log. I was able to get a repro of the failure in a manual run as well. I am able to see the expected log entry in `kubectl logs` for the pod, so it may just be something with the log query itself.

nickbp:
The log query turned out to be fine, instead it seems like it just consistently takes a bit longer for the logs to appear on the system tenant, vs the default tenant. I assume the system tenant is configured with something like a longer poll interval that in turn results in the logs taking a bit longer to appear in Loki. Not a big deal, was able to fix it by just increasing the maximum wait in the test.

PR #594, opened 2021-04-24
chore: handle older controller config schema

sreis:
Between the use of ControllerConfig V1 and V2 schemas there was a cluster launched that had the tenant_api_authenticator_pubkey_set_json field but before logRetention and metricRetention were renamed.


sreis:
I also ran the upgrade test for this PR, and it came back green. 

https://buildkite.com/opstrace/upgrade-tests/builds/478

PR #595, opened 2021-04-26
wip: test bare urls

triclambert:


triclambert:
Testing whether or not we can update links in the quickstart with our `opstrace_name` substitution.  Viewing a dev build against this branch, I'm reminded of the state of things: the link text will update, but the underlying target will not be updated.  Will do some more digging.

Issue #611, opened 2021-04-28
Metric exporter tests: Azure exporter: RequestError: Timeout awaiting 'request' for 60000ms


sreis:
**Describe the bug**

```
2021-04-28T04:54:44Z] 2021-04-28T04:54:44.710Z info: Got exporter log for query={k8s_namespace_name="default-tenant",k8s_container_name="exporter",k8s_pod_name=~"exporter-testexporters-zcqxb-azure-.+"} |= "Tenant 'testtenantshouldfail' not found": 1619585681289224468,log="2021/04/28 04:54:41 Failed to get token: Did not get status code 200, got: 400 with body: {"error":"invalid_request","error_description":"AADSTS90002: Tenant 'testtenantshouldfail' not found. This may happen if there are no active subscriptions for the tenant. Check to make sure you have the correct tenant ID. Check with your subscription administrator.\r\nTrace ID: 87c301b3-f3f9-4d64-93ec-7e91125a1101\r\nCorrelation ID: e22c0ab9-6cf6-4bd9-8534-8ffd0e61bedd\r\nTimestamp: 2021-04-28 04:54:41Z","error_codes":[90002],"timestamp":"2021-04-28 04:54:41Z","trace_id":"87c301b3-f3f9-4d64-93ec-7e91125a1101","correlation_id":"e22c0ab9-6cf6-4bd9-8534-8ffd0e61bedd","error_uri":"https://login.microsoftonline.com/error?code=90002"}
```
https://buildkite.com/opstrace/scheduled-main-builds/builds/2186#b5d02062-85da-4276-84e1-815f2da4e050/2796-6613

nickbp:
This Azure exporter test had just recently been merged, so it looks like there's still a flake or two in there.

nickbp:
Looks like the `default-tenant` check had succeeded while the `system-tenant` check failed a few lines down: https://buildkite.com/opstrace/scheduled-main-builds/builds/2186#b5d02062-85da-4276-84e1-815f2da4e050/2796-6699
```
[2021-04-28T04:54:45Z] 2021-04-28T04:54:45.366Z [32minfo[39m: Waiting for exporter log: {k8s_namespace_name="system-tenant",k8s_container_name="exporter",k8s_pod_name=~"exporter-testexporters-zcqxb-azure-.+"} |= "Tenant 'testtenantshouldfail' not found"
[...]

PR #612, opened 2021-04-28
test-remote: Adjust loki query options for handling spurious timeouts (#611)

nickbp:
In the context of a poll loop, we can set the timeout to be fairly short so that we can try again quickly in the event of a spurious request timeout.

# Have you...

### DIFF
--- a/test/test-remote/testutils/logs.ts
+++ b/test/test-remote/testutils/logs.ts
@@ -40,7 +40,7 @@ export interface LogRecord {
   labels: LogRecordLabels;
 }
 
-export async function queryLoki(baseUrl: string, queryParams: URLSearchParams) {
+async function queryLoki(baseUrl: string, queryParams: URLSearchParams) {
   /* Notes, in no particular order:
 
   - test deprecated /api/prom/query endpoint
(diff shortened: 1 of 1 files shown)


jgehrcke (commented):
Thanks, Nick!

test/test-remote/testutils/logs.ts
```diff
@@ -40,7 +40,7 @@ export interface LogRecord {
   labels: LogRecordLabels;
 }
 
-export async function queryLoki(baseUrl: string, queryParams: URLSearchParams) {
+async function queryLoki(baseUrl: string, queryParams: URLSearchParams) {
```
jgehrcke, line 43: I loved seeing you removing this `export` -- I was wondering if this function is used in different contexts, and this part of the diff was providing the answer :).

test/test-remote/testutils/logs.ts
```diff
@@ -66,10 +66,16 @@ export async function queryLoki(baseUrl: string, queryParams: URLSearchParams) {
...
+    // Allow up to two retries in the event of spurious timeouts or similar errors.
+    retry: 2,
     throwHttpErrors: false,
     searchParams: queryParams,
-    timeout: httpTimeoutSettings,
+    // Use a 5s request timeout rather than the configured 60s default.
+    // We want the poll loop to retry quickly if there is a timeout.
+    timeout: {
+      connect: httpTimeoutSettings.connect, // inherit default
+      request: 5000,
```
jgehrcke, line 77: I want to say 5 seconds are super tight.

For some kinds of queries this might simply be too low. Say, response generation for a specific query is expected to take 10 seconds then we will need to change this again. When the time comes :).

I'm good with the change for now.

test/test-remote/testutils/logs.ts
```diff
@@ -66,10 +66,16 @@ export async function queryLoki(baseUrl: string, queryParams: URLSearchParams) {
   const headers = enrichHeadersWithAuthToken(url, {});
 
   const options = {
-    retry: 0,
+    // Allow up to two retries in the event of spurious timeouts or similar errors.
+    retry: 2,
```
jgehrcke, line 70: It's fine, but I am not sure about the logging quality around the internal retries right now. Maybe we should let said loop body take care of the retries (the rational behind putting `retry: 0`) so that the build log contains all detail of what happened. Of course, with `got`-internal retrying we can still achieve great logging. I tried to work towards that here: https://github.com/opstrace/opstrace/blob/90906e24da0b5a677186d6e53ba0070184eafdf4/lib/utils/src/httpclient.ts#L43 -- but in the context of test-remote, we don't have that kind of control yet. I am saying: we might not see in the build log when this internal retrying had to hit in.

PR #616, opened 2021-04-29
test-remote: handle Got timeout errs in query loops

jgehrcke:


test/test-remote/testutils/index.ts
```diff
@@ -905,25 +907,43 @@ export async function waitForQueryResult<T>(
...
   while (true) {
     if (mtime() > deadline) {
       log.error("Reached %ss deadline, giving up", maxWaitSeconds);
       break;
     }
 
-    const data = await queryFunc();
+    let data: any;
+    try {
+      data = await queryFunc();
```
jgehrcke, line 928: CC @nickbp 

So, `queryFunc()` is expected to throw errors of type `got.RequestError` upon transient issues such as DNS resolution or timeout errors. I have added this to the docstring. It's an important property of this func all other functions that carry a `got....()` HTTP client call that is _not_ wrapped in a corresponding err handler.

Here, in this query loop, it of course makes sense to not crash the loop when this happens, but to go into the next loop iteration. 

What's important:
- the log message, so this does not happen silently
- to only catch errors emitted by the HTTP client, not other errors such as programming errors -- these are re-thrown.

test/test-remote/testutils/index.ts
```diff
@@ -905,25 +907,43 @@ export async function waitForQueryResult<T>(
...
+  testFunc: { (queryResponse: T): any | null },
   maxWaitSeconds = 30,
   logQueryResponse = false
 ) {
   const deadline = mtimeDeadlineInSeconds(maxWaitSeconds);
-  log.info("Waiting for query to return a match, deadline in %ss", maxWaitSeconds);
+  log.info(
+    "Waiting for query to return a match, deadline in %ss",
+    maxWaitSeconds
+  );
```
jgehrcke, line 918: auto-format :shrug: -- let's sync editor configs to prevent this

test/test-remote/testutils/logs.ts
```diff
@@ -85,6 +88,7 @@ async function queryLoki(baseUrl: string, queryParams: URLSearchParams) {
   // in the `options` object above and then use
   // `response.body.toString("utf-8")` below but
   // https://github.com/sindresorhus/got/issues/1079
+  // Note: this may throw got.RequestError for request timeout errors.
```
jgehrcke, line 91: (I should have clarified this comment better -- that this depends on `throwHttpErrors: false` and that this is about any transient issue -- next time).

PR #657, opened 2021-04-30
chore: add a dependabot config for github actions only

triclambert:
Let's avoid falling behind for the GH Actions that we use.


spahl (approved):

triclambert:
It appears to be working...

image

Issue #580, opened 2021-04-21
proposal: method to allow Cortex and Loki configuration tuning/override


sreis:
We need a practical way to change Cortex and Loki configuration settings in a running cluster.

Currently, it requires changing the Opstrace controller code, building and pushing a new image, and then upgrading the cluster. This means any parameter tunning becomes the default setting going forward for all Opstrace clusters.

# Proposal 1

Add a new field to the cluster config file to allow overriding configuration.

This is how it could look like to tune [`max_label_names_per_series` in Cortex](https://cortexmetrics.io/docs/configuration/configuration-file/#limits_config). 

```
tenants:
  - default
env_label: opstrace-ci
node_count: 3
cert_issuer: letsencrypt-staging

advanced_config:
	cortex:
		limits_config:
			max_label_names_per_series: 45
```

The Opstrace CLI, in the create and upgrade commands, adds the `advanced_config` section to the controller configuration as JSON (keeps the current behavior).

The controller is responsible for merging the default Opstrace settings with the overrides when it handles the [Cortex](https://github.com/opstrace/opstrace/blob/main/packages/controller/src/resources/cortex/index.ts#L248) and [Loki](https://github.com/opstrace/opstrace/blob/main/packages/controller/src/resources/loki/index.ts#L151) deployments.

# Proposal 2

Create a Kubernetes ConfigMap named, for example, `opstrace-controller-config-overrides` in the default namespace (where Opstrace controller ConfigMap is stored).

The ConfigMap would have a field where we add the Cortex and Loki config overrides as a YAML string. 

This is how it could look like to tune [`max_label_names_per_series` in Cortex](https://cortexmetrics.io/docs/configuration/configuration-file/#limits_config). 

```
apiVersion: v1
kind: ConfigMap
metadata:
  name: opstrace-controller-config-overrides
  namespace: default
data:
  cortex: |
	limits_config:
		max_label_names_per_series: 45  
```

The controller is responsible for merging the config defaults with the overrides when it handles the [Cortex](https://github.com/opstrace/opstrace/blob/main/packages/controller/src/resources/cortex/index.ts#L248) and [Loki](https://github.com/opstrace/opstrace/blob/main/packages/controller/src/resources/loki/index.ts#L151) deployments.

Deployment of the configuration overrides, in this scenario, is done via `kubectl`.

sreis:
[...]

PR #675, opened 2021-05-04
wip: automate json schema definition of some config structs

sreis:
Cross-linking to #580.

Uses github.com/alecthomas/jsonschema to convert golang structs that define Cortex, Loki and Alertmanager configuration to JSON schema. The JSON schema will be used to validate user input.

Currently only prints stuff to stdout.

In the process learned that Loki does not support golang modules properly and we need to pin to specific commits with `go get github.com/grafana/loki@COMMIT_HASH`.


go/tools/cortex-jsonschema-exporter/alertmanager-config.go
```diff
@@ -0,0 +1,33 @@
...
+
+	"github.com/alecthomas/jsonschema"
+	"github.com/cortexproject/cortex/pkg/alertmanager"
+)
+
+//
+// Prints the json schema definition of the Cortex configuration to stdout.
+//
+func main() {
+	schema := jsonschema.Reflect(&alertmanager.Config{})
```
sreis, line 29: @terrcin Can you tell me which struct you looked at when you created the schema in https://github.com/opstrace/opstrace/blob/main/packages/app/src/client/validation/alertmanager/config/types.ts ?

terrcin: @sreis I looked at the alertmanager config documentation itself and manually created those types based on that. Did it at the same time as the Yup validations.

sreis: Ok. I'll try to find the corresponding golang struct.

sreis: Changed to use the one that looked the most similar. 

@terrcin Any ideias how we can compare the autogenerated JSON schema with https://github.com/opstrace/opstrace/blob/main/packages/app/src/client/validation/alertmanager/config/schema.json ?

terrcin: @sreis maybe by eye at first. Do you have an example of the output somewhere I can look it? Also could do over zoom when we're next both online.

terrcin:
@sreis I've taken a quick look at the generated alertmanager json schema compared to the one I manually created. The main differences are the generated schema additionally specifies which fields are required while also being an older version of the spec format "draft-04" vs "draft-07".

sreis:
Closing since there's ongoing work to support runtime config API.

PR #679, opened 2021-05-04
wip(kubernetes): replace isDeepStrictEqual with fast-deep-equal

### LINKED ISSUE #588: proposal: replace isDeepStrictEqual with something else (probably fast-deep-equal)
# Problem 

The controller reconciliation loop checks if the resource spec (or some other attributes) changed and requires an update. 

Currently, we are using `isDeepStrictEqual,` but it has issues comparing an array of deeply nested objects which means some Kubernetes resources equality checks have to be written manually to compare the objects in an array. Most of them require this sort of manual equality check, and sometimes it doesn't cover all the fields in the resource. This lack of coverage leads to hard to track down bugs, like #585, because resources are not updated. 

# Proposed solution

Replace all the Kubernetes resource equality checks in the lib/kubernetes package with the functions in [`fast-deep-equal`](https://www.npmjs.com/package/fast-deep-equal) package.

MoSattler:
closes #588

### DIFF
--- a/lib/kubernetes/src/equality/AlertManager.spec.ts
+++ b/lib/kubernetes/src/equality/AlertManager.spec.ts
@@ -1,0 +1,62 @@
+/**
+ * Copyright 2020 Opstrace, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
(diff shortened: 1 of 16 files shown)


MoSattler:
closed in favour of #685

Issue #615, opened 2021-04-29
ci instability: * exporter: expectation not fulfilled within 300 s


jgehrcke:
failure pattern:
```
[2021-04-29T09:03:33Z]   22 passing (25m)
[2021-04-29T09:03:33Z]   4 failing
[2021-04-29T09:03:33Z]
[2021-04-29T09:03:33Z]   1) Metric exporter tests
[2021-04-29T09:03:33Z]        Cloudwatch exporter:
[2021-04-29T09:03:33Z]      Error: Expectation not fulfilled within 300 s
[2021-04-29T09:03:33Z]       at Object.waitForQueryResult (testutils/index.ts:933:9)
[2021-04-29T09:03:33Z]       at Object.waitForPrometheusTarget (testutils/metrics.ts:82:5)
[2021-04-29T09:03:33Z]       at testExporterMetric (test_exporters.ts:181:3)
[2021-04-29T09:03:33Z]       at Context.<anonymous> (test_exporters.ts:259:5)
[2021-04-29T09:03:33Z] 
[2021-04-29T09:03:33Z]   2) Metric exporter tests
[2021-04-29T09:03:33Z]        Stackdriver exporter:
[2021-04-29T09:03:33Z]      Error: Expectation not fulfilled within 300 s
[2021-04-29T09:03:33Z]       at Object.waitForQueryResult (testutils/index.ts:933:9)
[2021-04-29T09:03:33Z]       at Object.waitForPrometheusTarget (testutils/metrics.ts:82:5)
[2021-04-29T09:03:33Z]       at testExporterMetric (test_exporters.ts:181:3)
[2021-04-29T09:03:33Z]       at Context.<anonymous> (test_exporters.ts:312:5)
[2021-04-29T09:03:33Z] 
[2021-04-29T09:03:33Z]   3) Metric exporter tests
[2021-04-29T09:03:33Z]        Blackbox exporter:
[2021-04-29T09:03:33Z]      Error: Expectation not fulfilled within 300 s
[2021-04-29T09:03:33Z]       at Object.waitForQueryResult (testutils/index.ts:933:9)
[2021-04-29T09:03:33Z]       at Object.waitForPrometheusTarget (testutils/metrics.ts:82:5)
[2021-04-29T09:03:33Z]       at testExporterMetric (test_exporters.ts:181:3)
[2021-04-29T09:03:33Z]       at Context.<anonymous> (test_exporters.ts:368:5)
[2021-04-29T09:03:33Z] 
[...]

PR #684, opened 2021-05-05
bump: memcached 1.6.6 -> 1.6.9

jgehrcke:
This also changes the resource requests a little bit, no magic to be expected from that change.

It's important to recognize the difference between resource requests and resource limits in k8s: https://kubernetes.io/docs/concepts/configuration/manage-resources-containers/#requests-and-limits


jgehrcke:
main-test-gcp failed with known instability https://github.com/opstrace/opstrace/issues/615

PR #685, opened 2021-05-05
wip(kubernetes): replace isDeepStrictEqual with fast-deep-equal

MoSattler:
Signed-off-by: Mo Sattler <hi@mosattler.com>

Relates to #588

In this PR I merely replace the current `isDeepStrictEqual` with `fast-deep-equal`, and add some tests for it. 

TODO:
- [ ] Replace manually written array checks mentioned in #588 with `fast-deep-equal` and add some tests for those cases.
- [ ] Investigate if the comparing code which is specific to resource could be replaced by a more generic comparer function (see @sreis [comment](https://github.com/opstrace/opstrace/issues/588#issuecomment-831988684))

# Have you...

### DIFF
--- a/lib/kubernetes/src/equality/AlertManager.spec.ts
+++ b/lib/kubernetes/src/equality/AlertManager.spec.ts
@@ -1,0 +1,101 @@
+/**
+ * Copyright 2020 Opstrace, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { V1Alertmanager } from "..";
+import { isAlertManagerEqual } from "./AlertManager";
+
+// mock logger
+jest.mock("@opstrace/utils", () => ({
+  log: {
+    debug: jest.fn
+  }
+}));
+
+// return an empty certificate for testing
+function generateAlertManager(
+  template: Partial<V1Alertmanager> = {}
+): V1Alertmanager {
+  return {
+    metadata: {
+      /* start default metadata */
+      generation: 1,
+      resourceVersion: "1234",
+      selfLink: "/random/string",
+      uid: "randomstring",
+      /* end default metadata */
+      annotations: {
+        some: "annotation"
+      },
+      labels: {
+        some: "label"
+      }
+    },
+    spec: {
+      baseImage: "my/image"
+    },
+    ...template
+  };
+}
+
+test("should return true when spec matches and default metatada is set", () => {
+  const desired = generateAlertManager({
+    metadata: {
+      generation: 1,
+      resourceVersion: "1234",
+      selfLink: "/random/string",
+      uid: "randomstring"
+    }
+  });
+  const existing = generateAlertManager({
+    metadata: {
+      generation: 2,
+      resourceVersion: "5678",
+      selfLink: "/even/more/random/string",
+      uid: "evenmorerandomstring"
+    }
+  });
+  expect(isAlertManagerEqual(desired, existing)).toBe(true);
+});
+
+test("should return false when spec does not match", () => {
+  const desired = generateAlertManager({
+    spec: {
+      baseImage: "foo"
+    }
+  });
+  const existing = generateAlertManager({
+    spec: {
+      baseImage: "bar"
+    }
+  });
+
+  expect(isAlertManagerEqual(desired, existing)).toBe(false);
+});
+
+test("should return true when spec matches", () => {
+  const desired = generateAlertManager({
+    spec: {
+      baseImage: "foo"
+    }
+  });
+  const existing = generateAlertManager({
+    spec: {
+      baseImage: "foo"
+    }
+  });
+
+  expect(isAlertManagerEqual(desired, existing)).toBe(true);
+});
--- a/lib/kubernetes/src/equality/AlertManager.spec.ts
+++ b/lib/kubernetes/src/equality/AlertManager.spec.ts
@@ -70,6 +70,42 @@ test("should return true when spec matches and default metatada is set", () => {
   expect(isAlertManagerEqual(desired, existing)).toBe(true);
 });
 
+test("should return false when metatada.annotations changed", () => {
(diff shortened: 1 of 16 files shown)


lib/kubernetes/src/equality/index.ts
```diff
@@ -368,15 +352,15 @@ export const hasClusterRoleChanged = (
   return false;
 };
 
-// isDeepStrictEquals doesn't handle the CRD schema very well and always reports
+// equals doesn't handle the CRD schema very well and always reports
```
MoSattler, line 355: I guess this still holds true for `fast-deep-equal`, right @sreis ?

sreis: Yes, this can be 1M+ object we can use `fast-deep-equal` to make it consistent but let's keep the workaround with the annotations.

MoSattler:
@MatApple regarding the [env var filtering](https://github.com/opstrace/opstrace/blob/03f91eb8a57ef9b10f152d87a070df97a64d2b30/lib/kubernetes/src/equality/Pod.ts#L31)

> Yes, but unfortunately that isn’t clear in the code! That is an env var that we inject with the controller, so it’ll always be undefined on the desired resource, but could be defined on the existing object. We use ENV_HASH_NAME env var  to automatically roll any deployments etc that have a configmap or secret that changed value (automatic rollout when dependencies change).
> At the end of the reconcile loop (after we’ve checked for equality), we call this function that will update that env var and add those resources to the toUpdate collection https://github.com/opstrace/opstrace/blob/e9438b6c4a1a367962b7c299a6f3fc1f4c4d4075/lib/kubernetes/src/reconciliation/reconcile.ts#L472
> Adding something along these lines as a code comment would be nice!

lib/kubernetes/src/equality/ServiceMonitor.ts
```diff
@@ -14,63 +14,22 @@
...
 ): boolean => {
-  if (!isDeepStrictEqual(desired.spec.selector, existing.spec.selector)) {
+  if (!equal(desired.spec.selector, existing.spec.selector)) {
     return false;
   }
   if (desired.spec.jobLabel !== existing.spec.jobLabel) {
     return false;
   }
-  if (!areEndpointsEqual(desired, existing)) {
+  if (!equal(desired, existing)) {
```
sreis, line 30:
```suggestion
  if (!equal(desired.spec.endpoints, existing.spec.endpoints)) {
```

Otherwise, the controller will always try to update this resource because the equality check always fails. The check doesn't account for the fields Kubernetes sets in the metadata when the resource is created. 

Here's an example of how the resource looks like when it's created in Kubernetes.
```
> k get servicemonitors.monitoring.coreos.com -n system-tenant prometheus -o yaml
apiVersion: monitoring.coreos.com/v1
kind: ServiceMonitor
metadata:
  annotations:
    opstrace: owned
  creationTimestamp: "2021-05-07T11:19:46Z"
  generation: 1
  labels:
    app: prometheus
    prometheus: system-prometheus
    tenant: system
  name: prometheus
  namespace: system-tenant
  resourceVersion: "2795"
  selfLink: /apis/monitoring.coreos.com/v1/namespaces/system-tenant/servicemonitors/prometheus
  uid: 9d6a8309-2584-49e9-8090-329e0715cc0d
spec:
  endpoints:
  - interval: 30s
    path: /prometheus/metrics
    port: web
  jobLabel: prometheus
  selector:
    matchLabels:
      app: prometheus
```

sreis (changes requested):
Overall looks good but we need to fix the ServiceMonitor endpoints equality check.

Also, please add an extra test to all the resources you're adding that checks the equality checks are true when some default metadata fields are set, something like:

```
test("should return true when spec matches and default metatada is set", () => {
  const desired = generateAlertManager();
  const existing = generateAlertManager();

  desired.spec = {
    baseImage: "foo"
  };
  existing.spec = {
    baseImage: "foo"
  };
  existing.metadata["generation"] = 1;
  existing.metadata["resourceVersion"] = "1234";
  existing.metadata["selfLink"] = "/random/string";
  existing.metadata["uid] = "randomstring";

  expect(isAlertManagerEqual(desired, existing)).toBe(true);
});
```

lib/kubernetes/src/equality/ServiceMonitor.ts
MoSattler: Good catch!

MoSattler: Committed your suggestion, and also [added a test](https://github.com/opstrace/opstrace/pull/685/files#diff-b639d9f58c083bd9922e86c84af6a4577d0dbdf2c0c8e6b0f2f0024dd51d5abdR72-R81) for it

lib/kubernetes/src/equality/AlertManager.spec.ts
```diff
@@ -44,6 +44,57 @@ function generateAlertManager(
...
+    }
+  });
+  expect(isAlertManagerEqual(desired, existing)).toBe(false);
+});
+
+test("should return false when metatada.labels changed", () => {
+  const desired = generateAlertManager({
+    metadata: {
+      labels: {
+        my: "old-annotation"
```
sreis, line 84: label

lib/kubernetes/src/equality/AlertManager.spec.ts
```diff
@@ -44,6 +44,57 @@ function generateAlertManager(
...
+      }
+    }
+  });
+  const existing = generateAlertManager({
+    metadata: {
+      labels: {
+        my: "new-annotation"
```
sreis, line 91: old-label

sreis:
Unit tests are failing with:
```
[2021-05-07T15:23:38Z] FAIL src/equality/Certificate.spec.ts
[2021-05-07T15:23:38Z]   ● should return true if certificate dnsNames order changed
[2021-05-07T15:23:38Z]
[2021-05-07T15:23:38Z]     expect(received).toBe(expected) // Object.is equality
[2021-05-07T15:23:38Z]
[2021-05-07T15:23:38Z]     Expected: true
[2021-05-07T15:23:38Z]     Received: false
[2021-05-07T15:23:38Z]
[2021-05-07T15:23:38Z]       132 |   existing.spec.dnsNames = ["bar", "foo"];
[2021-05-07T15:23:38Z]       133 |
[2021-05-07T15:23:38Z]     > 134 |   expect(isCertificateEqual(desired, existing)).toBe(true);
[2021-05-07T15:23:38Z]           |                                                 ^
[2021-05-07T15:23:38Z]       135 | });
[2021-05-07T15:23:38Z]       136 |
[2021-05-07T15:23:38Z]       137 | test("should not be equal when certificate dnsNames does not match", () => {
[2021-05-07T15:23:38Z]
[2021-05-07T15:23:38Z]       at Object.<anonymous> (src/equality/Certificate.spec.ts:134:49)
```

You can run them locally with `make ts-unit-tests`.

PR #698, opened 2021-05-07
feat: add comparer function that disregards order

MoSattler:
In order to address the problem of nested arrays when comparing Kubernetes resources (see [here](https://github.com/opstrace/opstrace/issues/588#issuecomment-833642578)), I facilitated lodash's deepEqual function, that provides an escape hatch with `deepEqualWith`. 
Whenever an array needs to be compared, the code takes over control from lodash, and compares the arrays with disregard of item order. 

### DIFF
--- a/lib/kubernetes/src/equality/isDeepEqualInAnyOrder.test.ts
+++ b/lib/kubernetes/src/equality/isDeepEqualInAnyOrder.test.ts
@@ -1,0 +1,95 @@
+/**
+ * Copyright 2020 Opstrace, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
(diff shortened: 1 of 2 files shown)


MatApple (commented):
Hey @MoSattler thanks for proposing this! I really like the idea of creating an abstraction for our Array checks. They are currently hard to read and are repeated a lot. I think we can extract that difficult syntax into a reusable function with some clear commentary on how it works.

I'm also very open to having a full deepEqual object check if we can find a way to work around the defaults that Kubernetes injects into the `existing` object. @sreis any ideas here?

sreis:
> I'm also very open to having a full deepEqual object check if we can find a way to work around the defaults that Kubernetes injects into the existing object. @sreis any ideas here?

Nothing practical comes to mind. [Typescript classes](https://www.typescriptlang.org/docs/handbook/2/classes.html) can have fields with default values but we don't have those defined [upstream (example field in PodSpec with a default](https://github.com/kubernetes-client/javascript/blob/master/src/gen/model/v1PodSpec.ts#L44-L47) so I think we would have to PR upstream. Another option could be to go back and go over our codebase and add all the default values in all the resources we create.

PR #708, opened 2021-05-11
ui: bugfix for "can't save alertmanager configuration changes made in the UI"

terrcin:
The are two independent fixes in this PR:

1. Recent changes removing alertmanager template editing also ment those templates were not being included when saving configuration changes, what was resulting in a server side validation error. This fix simply includes the default set of templates when saving which is the same as the prior behaviour if the user didn't edit the templates.

2. As the user was updating the alertmanager configuration validation checks were performed to dynamically enable/disable the "save" button but those changes were not additionally being kept for use when the "save" button was clicked. This change now keeps the user's changes in `dataRef`. FYI: the underlying issue here is that we can't directly query the monaco editor for it's current value when the user hits save so we have to constantly record it as the user types.

### DIFF
--- a/packages/app/src/state/tenant/sagas/index.ts
+++ b/packages/app/src/state/tenant/sagas/index.ts
@@ -98,11 +98,11 @@ function* getAlertmanagerListener() {
 
 function* getAlertmanager(action: ReturnType<typeof actions.getAlertmanager>) {
   try {
-    const response: AsyncReturnType<typeof graphqlClient.GetAlertmanager> = yield graphqlClient.GetAlertmanager(
-      {
-        tenant_id: action.payload
-      }
-    );
+    const response: AsyncReturnType<
+      typeof graphqlClient.GetAlertmanager
+    > = yield graphqlClient.GetAlertmanager({
+      tenant_id: action.payload
+    });
 
     if (response.data?.getAlertmanager?.config) {
       const cortexConfig = yamlParser.load(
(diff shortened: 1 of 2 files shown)


packages/app/src/state/tenant/sagas/index.ts
```diff
@@ -98,11 +98,11 @@ function* getAlertmanagerListener() {
 
 function* getAlertmanager(action: ReturnType<typeof actions.getAlertmanager>) {
   try {
-    const response: AsyncReturnType<typeof graphqlClient.GetAlertmanager> = yield graphqlClient.GetAlertmanager(
-      {
-        tenant_id: action.payload
-      }
-    );
+    const response: AsyncReturnType<
```
terrcin, line 101: note: my editor reformatted this to match or prettierer settings

packages/app/src/state/tenant/sagas/index.ts
```diff
@@ -151,19 +157,22 @@ function* updateAlertmanager(
...
       }
     );
 
-    const response: AsyncReturnType<typeof graphqlClient.UpdateAlertmanager> = yield graphqlClient.UpdateAlertmanager(
-      {
-        tenant_id: action.payload.tenantName,
-        input: { config }
-      }
-    );
+    const response: AsyncReturnType<
```
terrcin, line 170: note: my editor reformatted this to match or prettierer settings

PR #695, opened 2021-05-06
test-remote: cover tenant creation flow, authn token usage derived from custom key pair (WIP, part 2)

jgehrcke:

PR #717, opened 2021-05-12
go: authenticator: fix tenant name prefix detection

jgehrcke:
```
For tenant names containing at least one dash,
the len(ssplits) is larger than two.

Generally, using more robust and expressive string
manipulation methods would have prevented this bug.
So, I am using HasPrefix() and TrimPrefix() now.

Also, if the error message is ever hit again:
log the actual claims.Subject, it's likely to reveal
the problem.

Signed-off-by: Dr. Jan-Philip Gehrcke <jp@opstrace.com>
```


jgehrcke:
It's great that this passed independently -- I have however merged the same commit via https://github.com/opstrace/opstrace/pull/695. Closing.

PR #719, opened 2021-05-12
build and update all Opstrace docker images to latest

MatApple:
While CI builds and updates the docker-images.json on every build, we need to also commit the updates to docker-images.json for local development. Only Opstrace org users have access to push the docker images, so for contributor's sake, we should keep this file as up-to-date as possible.

Note: @sreis @jgehrcke I couldn't find where we commit this file in CI, so assume we don't. Does this mean that when we run a CLI build, the git SHA that we reference as the CLI version will not help us to see the _updated_ docker-images.json (in GitHub) that the build was run with?

What do you think about doing this in CI?

Signed-off-by: Mat Appelman <mat@opstrace.com>


sreis:
> What do you think about doing this in CI?

When I last looked at this (before public launch), I couldn't figure out a way to add a commit during the build without triggering a new Buildkite CI build when you pushed the new commit with changes.

PR #724, opened 2021-05-13
ui: added a "check status" button to show page of integration

terrcin:
It directly checks with grafana if it's received metrics from the installed integration yet.


MatApple (approved):
Very cool!

PR #729, opened 2021-05-13
ui: new integration - k8s-logs

terrcin:
This is the matching pair for the `k8s-metrics` integration that we launched with. Uses @nickbp's promtail work


nickbp:
screenshot

nickbp:
At this point the logs/loki integration is more or less at parity with the metrics integration, so probably a good time to get it into main - I'm planning on merging tonight once the PR has turned green :crossed_fingers: 

**NOTE:** The implementation is incomplete. It should have a user-facing option to select whether dockerd or cri/containerd logs are being used. At the moment it's hardcoded to cri/containerd, see `TODO` in `k8sLogs/Show.tsx`. The option could be included in the integration creation form, or alternatively we could show the each option in the form of two different integration "types" in the parent list of integration types/cards?

PR #731, opened 2021-05-14
ui: add copy i9n commands to clipboard icon

terrcin:


terrcin:
note: the laytout is a bit unfun at the moment with the icon below the text being copied

PR #737, opened 2021-05-17
ui: link up the "install dashboard" functionality to the UI

terrcin:


nickbp:
I've tested the PR manually via `make start-remote-dev` + `yarn server:remote` + `yarn client:start`. I was able to create a metrics integration against an onprem cluster, which was successfully detected as `active`. I then successfully installed the dashboards into grafana via the magic button.

So the new dashboard functionality looks like it's working. I did catch a couple caveats:
- The metrics status `Check` button was observed to reuse the same timestamp across presses. The timestamp seemed to be from when the page was loaded. I would've expected it to query `now-1h to now`. I tried manually editing `CheckStatusBtn.tsx` to not cache the query URL, but then this somehow led to the check being run continuously in a loop so I refrained from committing that fix...
- The preinstalled dashboards themselves could likely be improved, but this is really just a proof-of-concept where the dashboards had been copied from the existing system-tenant preinstalled ones.

Summary: looks good to merge so doing so

nickbp:
screenshot

PR #761, opened 2021-05-19
More integrations styling for K8sMetrics (need to apply to other integrations)

MatApple:
Screen Shot 2021-05-19 at 11 49 07 AM


packages/app/src/client/viewsBasic/tenantIntegrations/tabs/Installed.tsx
```diff
@@ -47,6 +48,7 @@ const useStyles = makeStyles(theme => ({
 
 export const InstalledIntegrations = withIntegrationListFromParams(
   ({ integrationList }: IntegrationListProps) => {
+    // This integrationList needs to be mapped to integrationDefs (at least to access the Status component, and possibly Edit button)
```
MatApple, line 51: @terrcin

terrcin: @MatApple a nice way to do it would be that whenever integrations are loaded their respective definition is added to the object so you could easily go `integration.def...`.

At the moment you can import `integrationDefRecords` from `viewsBasic/integrationDefs/index.ts` and get it with `integrationDefRecords[integration.kind]`

terrcin: I can do that as part of updating the k8s-logs layout/stying

MatApple: Awesome sounds good!

PR #772, opened 2021-05-21
ui: update the i9n config file viewer to render properly and syntax highlight the yaml content

terrcin:


terrcin:
Currently have a display issue I need some help on. The Dialog is auto sizing based on the content and the YamlEditor autosizes based on the parent - ie the Dialog. So they fight and nobody sizes makeing it small. To get the content to show I'm hard [coding the height and width](https://github.com/opstrace/opstrace/pull/772/files#diff-cb4cb0949acd3a8e9cdb0d80a771135703404ad77fc8e6d786452aed493db471R61).

MatApple:
> Currently have a display issue I need some help on.
@terrcin since this is a Modal, how about we do something like the following:
```
const maxWidth = 1000;
const width = window.innerWidth > maxWidth ? maxWidth : window.innerWidth;
const height = window.innerHeight;
```
A better option would be to use https://usehooks.com/useWindowSize

terrcin:
@MatApple that works thanks.

I have a an additional scrollbar from the modal still :-/ So need to figure that one out next.

image

Hard to see in the screenshot as the scollbar fades away before I can take the screenshot, but to the left of the modal scrollbar is the monaco scrollbar area showing other occurances of the selected text in the whole document.

MatApple:
@terrcin I just tested this on various screen sizes and it resolves the double scrollbar. I'm sure there is a more idiomatic way of doing this, but my suggestion would be to roll with something like this here https://github.com/opstrace/opstrace/commit/c96be9ef7466d91f73c184ec3266e38f38be05e7

Issue #765, opened 2021-05-20
cli: list/destroy: AWS region lookup may fail for cn-north-1 / cn-northwest-1 with TCP connect timeouts


jgehrcke:
prior discussion: https://github.com/opstrace/opstrace/issues/511#issuecomment-803967724

> Educated speculation: based on past experience, this happens when AWS API endpoints in and around Asia are not TCP-reachable.

> Do not support all AWS regions. The ones that make problems here are often in Asia. Of course we also want users in Asia. And not supporting those has a weird taste -- feels like discrimination.

```
[2021-05-20T13:16:05Z] 2021-05-20T13:16:05.799Z info: Found Opstrace clusters in AWS us-west-2: prs-bk-4266-4a2-a, prs-bk-4267-b87-a, tracy, prs-bk-4265-a60-a, prs-bk-4268-5d3-a, schedul-bk-2320-826-a, schedul-bk-2321-5c5-a, upgrade-bk-632-703-a, chris6, jpload-1621421101, matdemo, prs-bk-4245-d39-a
[2021-05-20T13:16:09Z] 2021-05-20T13:16:09.750Z info: AWS API request failed (attempt 0): Socket timed out without establishing a connection (retry in 1.55 seconds)
[2021-05-20T13:16:09Z] 2021-05-20T13:16:09.750Z info: AWS API request failed (attempt 0): Socket timed out without establishing a connection (retry in 1.07 seconds)
[2021-05-20T13:16:16Z] 2021-05-20T13:16:16.830Z info: AWS API request failed (attempt 1): Socket timed out without establishing a connection (retry in 2.99 seconds)
[2021-05-20T13:16:17Z] 2021-05-20T13:16:17.302Z info: AWS API request failed (attempt 1): Socket timed out without establishing a connection (retry in 1.81 seconds)
[2021-05-20T13:16:25Z] 2021-05-20T13:16:25.124Z info: AWS API request failed (attempt 2): Socket timed out without establishing a connection (retry in 3.21 seconds)
[2021-05-20T13:16:25Z] 2021-05-20T13:16:25.824Z info: AWS API request failed (attempt 2): Socket timed out without establishing a connection (retry in 3.22 seconds)
[2021-05-20T13:16:34Z] 2021-05-20T13:16:34.340Z info: AWS API request failed (attempt 3): Socket timed out without establishing a connection (retry in 7.87 seconds)
[...]

PR #773, opened 2021-05-21
ci: retry teardown in AWS region if initial command failed

sreis:
The destroy command searches for the cluster in all the available AWS regions and sometimes it can timeout and fail to cleanup resources. We don't return an error code to signal this timeout error so the workaround is to retry the teardown in the known region when the initial command fails. The CI run will fail to signal the initial teardown error to ensure we look at the failure.
 
The upgrade test and command share the same issue so I just hardcoded the region to prevent any more resource leakage.


jgehrcke:
This approach is great; high impact with almost 0 effort.

> The CI run will fail to signal the initial teardown error to ensure we look at the failure.

That is important; thanks.

jgehrcke:
Will address most of the fallout of https://github.com/opstrace/opstrace/issues/765 on CI resource leakage.

PR #784, opened 2021-05-25
feat: display ring health 

MoSattler:
Todo:

- [ ] add FE tests
- [x] loading states 
- [x] error handling
- [x] display tokens
- [x] [move the sidebar tab for this under the "Health" tab](https://github.com/opstrace/opstrace/pull/784#issuecomment-849177243)

# Have you...


MatApple:
Hey @MoSattler, check out this code on the branch I'm currently working on. You might want to cherry-pick this commit to your branch and then add the ring endpoints here for Cortex. Then we can copy this file and do the same for Loki.

https://github.com/opstrace/opstrace/commit/fe1501b081d7cf95f46ace323c979facf7c649ed

MatApple:
@MoSattler you don't need to worry about the tenant in this case when talking to Cortex for the ring health.

Something like the following should work:
```
// For Ruler
cortex.use("/ruler/ring", (req, res) => {
    proxyTo(`http://ruler.cortex.svc.cluster.local/ruler/ring`, req, res);
  });

// For Ingester
cortex.use("/ingester/ring", (req, res) => {
    proxyTo(`http://ingester.cortex.svc.cluster.local/ingester/ring`, req, res);
  });
```

packages/app/src/server/routes/api/ringHealth.ts
```diff
@@ -0,0 +1,35 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+import axios from "axios";
+import express from "express";
+
+const getCortexClient = () =>
+  axios.create({
+    baseURL: "https://cortex.internet" // TODO DO NOT MERGE!!!!!!
+    // baseURL: 'https://cortex.${tenantName}.${clusterHost}', // TODO DO NOT MERGE!!!!!!
```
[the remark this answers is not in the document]
sreis: From the package folder name, I assume this code runs server-side (meaning, inside the Kubernetes cluster). If that is the case it should use the Kubernetes DNS name: `http://cortex.${namespace}.svc.cluster.local/` where `namespace` is `${tenantName}-tenant.`

packages/app/src/server/routes/api/cortex.ts
```diff
@@ -0,0 +1,143 @@
...
+
+  const contentType = proxyReq.getHeader("Content-Type");
+  const writeBody = (bodyData: string) => {
+    proxyReq.setHeader("Content-Length", Buffer.byteLength(bodyData));
+    proxyReq.write(bodyData);
+  };
+
+  if (contentType === "application/json") {
+    // @ts-ignore
+    writeBody(JSON.stringify(req.body));
```
jgehrcke, line 53: This is wild :-) :fire:  Just to summarize, the HTTP proxy architecture we use here
- reads the request bytes, deserializes as JSON (this is what bodyParser is doing)
- reserializes to a byte sequence again and only then passes the request on to the proxy target

Instead, it should transparently forward-stream the body byte sequence.

I am optimistic that there is a better solution; using canonical tools from the NodeJS/express ecosystem. What about

https://github.com/chimurai/http-proxy-middleware which uses https://github.com/http-party/node-http-proxy under the hood (the same library used here via `import httpProxy from "http-proxy";`).

There is a comment in https://github.com/chimurai/http-proxy-middleware#intercept-and-manipulate-requests which suggests that this middleware should be applied before the bodyParser middleware:

> ... which is used to fix proxied POST requests when bodyParser is applied before this middleware.

This makes sense -- then the proxy middleware has access to the "raw" request body in a streaming fashion.

The same pointer, but way more explicit, is provided in 
https://www.npmjs.com/package/express-http-proxy#middleware-mixing

> Proxy requests and user responses are piped/streamed/chunked by default.

> If you use 'https://www.npmjs.com/package/body-parser' you should declare it AFTER the proxy configuration, otherwise original 'POST' body could be modified and not proxied correctly.

MoSattler: cc @MatApple

packages/app/src/client/views/ring-health/RingTable.tsx
```diff
@@ -0,0 +1,92 @@
...
+            <TableCell>ID</TableCell>
+            <TableCell>State</TableCell>
+            <TableCell>last Heartbeat</TableCell>
+            <TableCell>Zone</TableCell>
+            <TableCell>Address</TableCell>
+            <TableCell>Tokens</TableCell>
+          </TableRow>
+        </TableHead>
+        <TableBody>
+          {shards?.map(shard => (
```
MoSattler, line 74: It'd probably be better to show some spinner. Do we already have something like this - to keep things consistent?

packages/app/src/client/views/ring-health/RingTable.tsx
```diff
@@ -0,0 +1,92 @@
...
+type Props = {
+  ringEndpoint: string;
+};
+
+const RingTable = ({ ringEndpoint }: Props) => {
+  const [shards, setShards] = useState<Array<Shard>>();
+  useEffect(() => {
+    async function fetchShards() {
+      const response = await axios.get<Payload>(ringEndpoint);
+      setShards(response.data.shards);
```
MoSattler, line 55: I need to add error handling. Do we already have error boundaries setup somewhere? How should errors be handled?

MoSattler (commented):
Todo:

- [ ] add FE tests
- [x] loading states 
- [ ] error handling
- [ ] display tokens
- [ ] move the sidebar tab for this under the "Health" tab

Current state:
Screenshot 2021-05-26 at 17 44 22

packages/app/src/server/routes/api/cortex.ts
MatApple: This is fine for a first pass, @jgehrcke and I talked about this offline. We can move the order of middleware and fix this problem - I'll take a look at this after we land this PR

packages/app/src/client/views/ring-health/RingTable.tsx
MatApple: We use the <Skeleton /> component for loading. In this case, instead of showing the table during loading, you can show 4 horizontal skeletons that looks like the following: https://material-ui.com/components/skeleton/#animations

packages/app/src/client/views/ring-health/RingTable.tsx
MatApple: We do have a top-level ErrorBoundary but in this case, let's just use https://material-ui.com/components/alert/#alert and show a red Alert in place of the table, with the error as the message

MatApple:
@MoSattler let's move the sidebar tab for this under the "Health" tab. The existing page that renders for the Health page can be renamed to "System". Then we can add a new tab for this ring page that goes under "Metrics". We'll then need to repeat all this for "Logs" (separate PR).

New nav under **Cluster Admin** then is:

**Health**
  *  **System** -> [existing "Health" page](https://github.com/opstrace/opstrace/blob/main/packages/app/src/client/views/cluster-health/index.tsx)
  *  **Metrics** -> this is your page for Cortex ring tables. Note: you'll have to use a different component like a [ButtonGroup](https://material-ui.com/components/button-group/) for switching between the components (Ingester, Ruler, etc). It's fine that these buttons are not routed for now.

MoSattler:
Added a loading state: 
Screenshot 2021-05-27 at 15 04 28

packages/app/src/client/views/ring-health/TokenDialog.tsx
```diff
@@ -0,0 +1,41 @@
...
+    <Dialog open={!!tokens} onClose={onClose} aria-labelledby="simple-dialog-title">
+      <DialogTitle id="simple-dialog-title">Tokens</DialogTitle>
+      <Box pl={3} width={300} height={500}>
+        {tokens?.map(token => (
+          <div>
+            {token}
+          </div>
+        ))}
+      </Box>
+    </Dialog>
```
MoSattler, line 37: The Dialog component doesn't seem to be particularly responsive, so right now it just shows a one column line of tokens. I can add manually responsiveness - unless we already have a responsive dialog.

MoSattler (commented):
It's still missing tests, but as this PR is quite big, I think already getting some reviews would be beneficial, as well as collecting feedback regarding the dialog question,

packages/app/src/client/views/ring-health/TokenDialog.tsx
MatApple: @MoSattler check out this example on how to set the Dialog to be responsive with a maxSize https://next.material-ui.com/components/dialogs/#optional-sizes

cc/ @terrcin (I know you've also been working with the Dialog and had responsive challenges)

packages/app/src/client/views/ring-health/index.tsx
```diff
@@ -0,0 +1,106 @@
...
+};
+
+const RingHealth = ({ baseUrl }: Props) => {
+  const location = useLocation();
+  const classes = useStyles();
+
+  const tabs = [
+    {
+      title: "Ingester",
+      path: `${baseUrl}/ingester`,
```
MatApple, line 46: 👏

packages/app/src/server/routes/api/cortex.ts
```diff
@@ -0,0 +1,125 @@
...
+import { log } from "@opstrace/utils/lib/log";
+
+export const cortexProxy = httpProxy.createProxyServer({ ignorePath: true });
+/**
+ * bodyParser middleware (which we use earlier in chain to parse POST body content), doesn't play
+ * well with the http-proxy, since the body is parsed and altered. It results in POSTs hanging until the connection times out.
+ *
+ * This is a workaround to restream the already parsed body, for the proxy target to consume.
+ */
+cortexProxy.on("proxyReq", function onProxyReq(proxyReq, req, res) {
```
MatApple, line 29: We can import {onProxyReq} from "server/utils" now and use that here instead of redeclaring the function again.

MatApple:
@MoSattler can we make the notificationService a little more pretty? It's in need of some color adjustments (backgrounds and text) and we should add some border-radius to it (standard to our theme `theme.shape.borderRadius`).

Check the styles on dark mode too by using the command+k hotkeys to open the command palette and turn dark mode on/off.
Screen Shot 2021-05-27 at 6 41 12 PM
Screen Shot 2021-05-27 at 6 43 21 PM

packages/app/src/client/views/ring-health/RingTable.tsx
```diff
@@ -0,0 +1,146 @@
...
+  return (
+    <>
+      <TokenDialog tokens={tokensToDisplay} onClose={() => history.push(baseUrl)} />
+      <TableContainer component={Card}>
+        <Table aria-label="tenants">
+          <TableHead>
+            <TableRow>
+              <TableCell>ID</TableCell>
+              <TableCell>State</TableCell>
+              <TableCell>last Heartbeat</TableCell>
```
MatApple, line 106: Let's capitalize the L

packages/app/src/client/views/ring-health/RingTable.tsx
```diff
@@ -0,0 +1,146 @@
...
+              {shards.map(shard => (
+                <TableRow key={shard.id}>
+                  <TableCell component="th" scope="row">
+                    {shard.id}
+                  </TableCell>
+                  <TableCell>{shard.state}</TableCell>
+                  <TableCell>
+                    {format(new Date(shard.timestamp), "Pppp")}
+                  </TableCell>
+                  <TableCell>{shard.zone}</TableCell>
```
MatApple, line 123: if zone is nullish, render "-"

packages/app/src/client/views/ring-health/RingTable.tsx
```diff
@@ -0,0 +1,146 @@
...
+          {shards ? (
+            <TableBody>
+              {shards.map(shard => (
+                <TableRow key={shard.id}>
+                  <TableCell component="th" scope="row">
+                    {shard.id}
+                  </TableCell>
+                  <TableCell>{shard.state}</TableCell>
+                  <TableCell>
+                    {format(new Date(shard.timestamp), "Pppp")}
```
MatApple, line 121: For this one, can we use date-fns to show it as "<time> ago", e.g. "2 mins ago"?

MatApple:
@MoSattler this is great work! Especially like the routed token links too! 👏 Feel free to merge after addressing the comments above and CI is green. We can continue to iterate once merged.

MatApple (approved):
Approving based on existing comments being addressed.

MoSattler:
> can we make the notificationService a little more pretty? It's in need of some color adjustments (backgrounds and text) and we should add some border-radius to it (standard to our theme theme.shape.borderRadius).

Sure thing. Keep in mind though, I am using the notification service - meaning any changes I make will be global to that service and affect these alerts everywhere.

MoSattler:
btw, I am unable to merge PRs
Screenshot 2021-05-28 at 11 52 38

packages/app/src/client/views/ring-health/RingTable.tsx
MoSattler: it's possible, but it means we need to start account for timezones (server vs client timezone), which can be a bit tricky in JS.

MatApple:
> Sure thing. Keep in mind though, I am using the notification service - meaning any changes I make will be global to that service and affect these alerts everywhere.

You're the first one to use it :-)

PR #785, opened 2021-05-25
docs: update docs with fixes and other basic improvements

triclambert:
Several of the docs are out of date and/or just in need of some polish.


MatApple (approved):
🎉

PR #787, opened 2021-05-25
ci: add a buildkite pipeline to upgrade a long running cluster

sreis:
Add a Buildkite pipeline to upgrade a long running cluster to the latest version. The cluster is ingesting 2.5M metrics. I'll set up the pipeline to run once a day.

Also, fixes a bug I noticed when upgrading the cluster to the same version. The CLI should exit early if a controller rollout is not required.


sreis:
Stopped PR tests since this introduces a new Buildkite pipeline and it's already running https://buildkite.com/opstrace/continuous-deployment-aws/builds/12#d8b28861-035b-4c40-9a95-6529f15d8aeb.

PR #794, opened 2021-05-26
ui: refactor the old "cloud metrics" Exporters as Integrations

terrcin:
We're also simplify things by rolling up "credentials" into each exporter (now an integration). So for each exporter the user needs to enter appropriate credentials at the same time rather than create them elsewhere and pick them from a list.

Current progress.

Screen Shot 2021-05-27 at 11 32 29 AM

Screen Shot 2021-05-27 at 11 34 45 AM


MatApple:
@terrcin this is great! Can we change the copy on the two exporter descriptions to (and name of the Google one, since Stackdriver was the old name):

**Amazon CloudWatch**
Pipe any of your metrics from CloudWatch into Opstrace. You can select metrics from any of the AWS Services such as RDS or Load Balancers, as long as you've enabled CloudWatch monitoring on the service in the AWS console.

**Google Cloud Monitoring**
Pipe any of your metrics from Google Cloud Monitoring into Opstrace. You can select metrics from any of the Google Cloud Services such as BigTable, or Load Balancers as long as you've enabled monitoring on the service in the Google Cloud Console.

packages/app/migrations/1622003137738_alter_table_public_integration_add_column_key/up.sql
```diff
@@ -0,0 +1,7 @@
+ALTER TABLE "public"."integration" ADD COLUMN "key" text NULL;
+ALTER TABLE "public"."integration" ADD CONSTRAINT "integration_key_key" UNIQUE ("key");
+
+-- There are already DNS records using the integration name, so need to keep those working
```
MatApple, line 4: Which DNS records are you referring to?

MatApple: Don't we have a UUID PK for integrations?

packages/controller/src/reducers/graphql/integrations.ts
```diff
@@ -24,46 +24,53 @@ import { ResourceCache } from "./util";
...
+  tenant_id: string;
+  // This data is defined on a per-integration basis, see *IntegrationData types.
+  // Some integration kinds do not involve the controller at all, so we allow this to be any.
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  data: any;
 }
-export type Exporters = Exporter[];
+export type Integrations = Integration[];
 
 export const actions = {
```
MatApple, line 43: This file feels like its a duplicate now of the state we already load for Integrations https://github.com/opstrace/opstrace/blob/main/packages/app/src/state/integration/reducer.ts

MatApple: my bad, this is the controller! Carry on :-D

packages/app/migrations/1622003137738_alter_table_public_integration_add_column_key/up.sql
terrcin: ah my bad with the comment. this is for use in k8s pod names rather than using the UUID. Like the same named and trigger populated field on Tenant it's not a PK.

MatApple: Gotcha, because we need to scrape a consistent metrics endpoint on the Exporters 👍 makes sense

terrcin:
FYI: these new integrations are being wrapped in the `EARLY_RELEASE` for this PR. The outstanding piece of work to remove that is to do end-to-end testing as the status checks likely need a bit of fine tuning due to being a bit more complicated now. Previously the exporters would only show if they were erroring or not, now we have a status field that also needs to show if the integration is actually active too.

MatApple:
Sounds great @terrcin 🎉

terrcin:
Very much enjoying the new styling

Screen Shot 2021-05-27 at 4 11 39 PM

terrcin:
note: It's getting to late in my Friday afternoon to comfortably merge this to main and then go offline for a couple of days. Plan is to merge first thing on my Monday.

terrcin:
Have successfully manually tested the migration upgrade process

Issue #733, opened 2021-05-14
Adding a tenant via UI: API never comes up, no error message shown


jgehrcke:
When you add a tenant via UI and use capital letters, the tenant's API will never come available. The exact technical problem is described in https://github.com/opstrace/opstrace/issues/710.

jgehrcke:
To make this a tiny little less severe, in the blog post we wrote:

> Warning: we do not yet do strict tenant name validation. To make sure things work, please keep the name lower case [a-z] for now.

https://opstrace.com/blog/introducing-dynamic-tenant-addition-and-token-management/

terrcin:
I've updated the validation regex used in https://github.com/opstrace/opstrace/pull/994 to exclude uppercase.

The main problem here is that the interface we use for getting the tenant name from users doesn't allow for validation feedback, so if they put in some invalid we accept it from them then do nothing.

jgehrcke:
> The main problem here is that the interface we use for getting the tenant name from users doesn't allow for validation feedback, so if they put in some invalid we accept it from them then do nothing.

Ouch. That is of really limited value; and almost as annoying UX-wise as the previous state. 

Let's accelerate the work on the API-based validation seriously; and then make sure that the API-based validation feedback is displayed to the user.

jgehrcke:
This is fixed now that the API rejects invalid tenant names.

PR #798, opened 2021-05-28
docs: cleanup and catchup (i.e., add docs that are missing)

triclambert:
Clean up existing docs and add missing docs.

- [x] Add an Adding and Managing Tenants guide
- [x] Add an Adding and Managing Users guide
- [x] Update Writing guide
- [x] Reduce reliance on the word "cluster" in modified docs


triclambert:
@MatApple what do you think about the drop-shadow option?  See [docs/assets/tenants-guide-delete-2.gif](https://github.com/opstrace/opstrace/blob/9e02d650d9401533cfcddd4aac573241252892ce/docs/assets/tenants-guide-delete-2.gif)

docs/guides/administrator/adding-and-managing-tenants.md
```diff
@@ -0,0 +1,144 @@
+# Adding and Managing Tenants Guide
+
+This guide will show you how to:
+
+* Add tenants during the initial create operation
+* Add new tenant to a running cluster (with public and private keys)
```
jgehrcke, line 6: missing "a"

jgehrcke: remove "(with public and private keys)"

docs/guides/administrator/adding-and-managing-tenants.md
```diff
@@ -0,0 +1,144 @@
...
+* Delete a tenant
+
+## Introduction
+
+Opstrace instances support multiple, secured tenants to logically isolate concerns on otherwise shared infrastructure.
```
jgehrcke, line 11: Prefer singular: "An Opstrace instance supports ..."

docs/guides/administrator/adding-and-managing-tenants.md
```diff
@@ -0,0 +1,144 @@
...
+If you are not familiar with what a "tenant" is, see our [tenant key concepts](../../references/concepts.md#tenants) documentation.
+
+Tenants are presented neatly in the Cluster Admin section of our UI:
```
jgehrcke, line 14: remove "neatly", let the reader judge :D

docs/guides/administrator/adding-and-managing-tenants.md
```diff
@@ -0,0 +1,144 @@
...
+
+![tenant overview page](../../assets/tenants-guide-overview-1.png)
+
+Let's walk through an example of adding a new tenant named `newtenant` to a running Opstrace instance named `showdown`.
+
+If you’re coming from the [quick start](../../quickstart.md), and haven’t yet sent data to one of your tenants, stay tuned for our forthcoming integrations guide to make that process easy.
```
jgehrcke, line 20: > stay tuned for our forthcoming integrations guide to make that process easy.

confuses me as a reader... does this refer to content below? does this refer to content being published in the future? how will I find this content? and why am I really interested in reading it?

docs/guides/administrator/adding-and-managing-tenants.md
```diff
@@ -0,0 +1,144 @@
...
+## Add a New Tenant in the UI
+
+Visit `https://showdown.opstrace.io/cluster/tenants` and press the `Add Tenant` button.
+Type the name of the new tenant (here: `newtenant`).
+
+![add tenant gif](../../assets/tenants-guide-add.gif)
+
+**Warning**: we do not yet do strict tenant name validation.
+To make sure things
+work, please keep the name lower case `[a-z]` for now.
```
jgehrcke, line 94: CC @MatApple @terrcin and the corresponding bug ticket: https://github.com/opstrace/opstrace/issues/733 -- let's address this soon and then remove this bit from this documentation.

docs/guides/administrator/adding-and-managing-tenants.md
```diff
@@ -0,0 +1,144 @@
...
+
+Getting a `200` response (and not a `401` response) means: the authentication token provided in the request was accepted.
+The so-called _authenticator_ in the Opstrace instance extracted the public key ID from the token's header section, found a corresponding public key in its trust store (think: "set of public keys that I am configured to trust"), and then performed a cryptographic verification using that public key.
+It also confirmed that the tenant name encoded in the token matches the tenant associated with the API endpoint.
+
+You could now go ahead and take this authentication token and configure serious API clients with it, such as a Prometheus instance to be able to `remote_write` to `https://cortex.newtenant.showdown.opstrace.io/api/v1/push`.
+
+## Send Data to a new Tenant with an Integration
+
+Opstrace provides integrations that simplify collection of data from various.
```
jgehrcke, line 134: from various ...

docs/guides/administrator/adding-and-managing-tenants.md
```diff
@@ -0,0 +1,144 @@
...
+This includes automating the configuration of the API tokens.
```
jgehrcke, line 135: > This includes automating the configuration of the API tokens

The meaning of this sentence is not clear. Let's try to reword.

_issuing_ authentication tokens always has to be done manually.

_using_ them has to be done kind of manually.

Which kind of automation do we refer to here?

jgehrcke:
In the admin guide for adding a tenant, we should make it very clear that

- As an admin, one can issue an _arbitrary number_ of authentication tokens from the _same_ RSA private key.
- As an admin, one is _responsible_ for managing the private key.

I think this is one of the most important UX aspects to clarify prominently here; because right now the content is misleading: was already a weakness in the blog post -- the content kind of suggests that the intended way to do things is to create a new RSA key pair for _each tenant to be added_.

That is really not what should happen.

There may be reasons to create more than one RSA key pair; and these reasons will always have to do with security management: for example; if there is more than one admin persona, each of them may want to create an RSA keypair for themselves.

docs/guides/administrator/adding-and-managing-tenants.md
```diff
@@ -0,0 +1,144 @@
+# Adding and Managing Tenants Guide
```
jgehrcke, line 1: Adding and Managing?

"Adding" is part of "managing" I suppose.

Maybe let's do "Adding and Removing Tenants" or just "Managing Tenants".

docs/guides/administrator/adding-and-managing-tenants.md
```diff
@@ -0,0 +1,144 @@
...
+{"status":"success","data":[]}
+```
+
+Getting a `200` response (and not a `401` response) means: the authentication token provided in the request was accepted.
+The so-called _authenticator_ in the Opstrace instance extracted the public key ID from the token's header section, found a corresponding public key in its trust store (think: "set of public keys that I am configured to trust"), and then performed a cryptographic verification using that public key.
+It also confirmed that the tenant name encoded in the token matches the tenant associated with the API endpoint.
+
+You could now go ahead and take this authentication token and configure serious API clients with it, such as a Prometheus instance to be able to `remote_write` to `https://cortex.newtenant.showdown.opstrace.io/api/v1/push`.
+
+## Send Data to a new Tenant with an Integration
```
jgehrcke, line 132: Side note, not too relevant for this review. Language-wise / expression-wise we have to improve on this: "to a tenant" is not really good in systems thinking. Maybe "to the tenant ingest API". The tenant is not just "in the cluster/instance", but it is more than that: the _sending_ side of things belongs to it, too.

"with an Integration" is a difficult language construct, doesn't sound right; too bloaty/abstract.

docs/guides/administrator/adding-and-managing-tenants.md
```diff
@@ -0,0 +1,144 @@
...
+
+Opstrace provides integrations that simplify collection of data from various.
+This includes automating the configuration of the API tokens.
+For example, from a Kubernetes cluster.
+
+For more information, visit our forthcoming Integrations Guide.
+
+## Delete a Tenant
+
+Deleting a tenant is simple, and permanent.
```
jgehrcke, line 142: Why is "permanent" important? It sounds super important because the word is here, and it leaves the reader wondering.. what does it mean, really? Sounds like the same tenant (same name) cannot be re-added later. Is that what it means to say?

docs/guides/administrator/adding-and-managing-users.md
```diff
@@ -0,0 +1,24 @@
+# Adding and Managing Users
+
+## Introduction
+
+An administrator can provision and manage users in Opstrace.
+Users allow authentication against our web UI and the per-tenant Grafana UI.
```
jgehrcke, line 6: Let's maybe reword "Users allow authentication"

Sounds like there is a group of people that allow me to do something :-).

A key difference between tenants and users that we should point out somewhere:

- A tenant's identity is meant to be used by machines/scripts.
- A user's identity is meant to be used by human users sitting in front of the browser, never by scripts.

docs/guides/administrator/adding-and-managing-users.md
```diff
@@ -0,0 +1,24 @@
...
+An administrator can then* provide fine-grained permissions for what users can and cannot do.
```
jgehrcke, line 7: Can the admin really do that as of today?

docs/guides/administrator/adding-and-managing-users.md
```diff
@@ -0,0 +1,24 @@
...
+
+![users management UI overview](../../assets/users-guide-overview-1.jpg)
+
+\* _This is a roadmap item; all users are currently granted the ADMIN role_
```
jgehrcke, line 11: oh I see

docs/guides/administrator/adding-and-managing-users.md
```diff
@@ -0,0 +1,24 @@
...
+## How to add a user
+
+Adding a user is easy.
+
+![adding a user gif](../../assets/users-guide-add.gif)
+
+## How to remove a user
+
+Removing a user can be done.
+However, as you might expect, **this operation cannot be undone**.
```
jgehrcke, line 22: > **this operation cannot be undone**.

Let's clarify what this means. For example, can't I re-add the same user (as identified by the same email address as before)?

Do we mean to say that any associated relational data will be deleted, too? (such as permissions, which... we don't offer yet)

docs/references/concepts.md
```diff
@@ -4,16 +4,35 @@ description: Fundamental user-facing concepts
 
 # Key Concepts
 
+## Instance
+
+An "instance" of Opstrace runs inside your cloud account.
+It is installed with the CLI `opstrace create` command.
+This name must be _globally_ unique and limited to 23 characters (`[a-z0-9-_]`) because it used to provide a conveient DNS name:  `<instance_name>.opstrace.io`.
```
jgehrcke, line 11: > conveient

jgehrcke: Personally, I think we should say a few things about that an Opstrace instance is comprised of multiple machines/nodes,   a Kubernetes cluster, and a bunch of cloud resources. Just to architecturally make it _clear_ that the instances entails _everything_ (or -- does it? important architectural question to be answered in the mid term)

docs/references/concepts.md
```diff
@@ -4,16 +4,35 @@ description: Fundamental user-facing concepts
...
+Once chosen, an instance name cannot be changed.
+
 ## Tenants
```
jgehrcke, line 14: Didn't we want to create a new page for that?

docs/references/concepts.md
```diff
@@ -4,16 +4,35 @@ description: Fundamental user-facing concepts
...
 
-The concept of multitenancy allows for isolating environments within the same Opstrace cluster.
-Tenants are logically isolated, but otherwise share the same \(physical\) base system.
+A "tenant" is a unit of isolation.
+[Multi-tenant systems](https://en.wikipedia.org/wiki/Multitenancy) allow for logical separation of concerns, while otherwise sharing the same underlying system.
+
+Opstrace instances support multiple, secured tenants.
```
jgehrcke, line 19: I have always wondered what the value is of saying ", secured" here. To me, this is marketing speech. 

Rationally, the following sentence has good and _important_ information content:

> A single Opstrace instance supports multiple tenants

The main task of this sentence is to point out the one-to-many relationship here.

To make this message unambiguous, it is important to use `Opstrace instance" in singular.

If we want to say something about the relationship between tenants and security let's do this in its own sentence or paragraph!

docs/references/concepts.md
```diff
@@ -4,16 +4,35 @@ description: Fundamental user-facing concepts
...
+The tenant primitive extended from the underlying projects—Cortex and Loki.
```
jgehrcke, line 20: Is this sentence correct? "extended" vs "is extended"? Maybe "builds on top of .."?

Is the primary purpose of this sentence to make a shout-out/ an attribution?

Maybe something like 

"The tenant primitive in Opstrace walks on the shoulders of giants; it is built on top of the multi-tenancy capabilities in Cortex and Loki"

?

docs/references/concepts.md
```diff
@@ -4,16 +4,35 @@ description: Fundamental user-facing concepts
...
+Tenants isolate isolate data, dashboards, and API endpoints.
```
jgehrcke, line 21: isolate isolate

docs/references/concepts.md
```diff
@@ -4,16 +4,35 @@ description: Fundamental user-facing concepts
...
+For example, isolate by:
```
jgehrcke, line 22: Maybe

"For example, you can use tenants to achieve separation by .."

docs/references/concepts.md
```diff
@@ -4,16 +4,35 @@ description: Fundamental user-facing concepts
...
+
+* Teams (e.g., `team-revenue` vs. `team-fraud`)
+* Environment (e.g., `prod`, `staging`, `dev`)
```
jgehrcke, line 25: Team and Environment should both be singular

docs/references/concepts.md
```diff
@@ -4,16 +4,35 @@ description: Fundamental user-facing concepts
...
+* A combination of these things, or anything else important to you.
```
jgehrcke, line 26: I think this is not a third example, and should therefore not be in the list, but below it as a new sentence.

docs/references/concepts.md
```diff
@@ -4,16 +4,35 @@ description: Fundamental user-facing concepts
...
 
-Tenants can be used for logical separation of concerns \(e.g., `prod`. vs. `ci`\), for isolating teams and people, or for any kind of categorization that makes sense in your specific use case.
+Sometimes this isolation is used to address security concerns, but more generally, it helps keep things well-organized: because, after all, separation of concerns is a valuable best practice to keep in mind.
 
-Each Opstrace cluster comes with a "system tenant" which ingests cluster-internal system logs and metrics.
+Opstrace comes with a default "system tenant" which ingests internal system logs and metrics.
+The system tenant is useful for the administrator to monitor the overall health of Opstrace, across tenants.
+It cannot be deleted.
```
jgehrcke, line 32: > It cannot be deleted.

I hope so :D. Do we know? haha.

docs/references/configuration.md
```diff
@@ -0,0 +1,192 @@
+---
+description: Parameter reference
+---
+
+# Configuration
+
+When installing Opstrace, one has to provide a corresponding configuration document in YAML format.
```
jgehrcke, line 7: > When installing Opstrace

I know this is old content, buuut WDYT about 

"When creating a new Opstrace instance, ..."

docs/references/configuration.md
```diff
@@ -0,0 +1,192 @@
...
+*Value type:* list of strings
+
+*Example:*
+
+```yaml
+tenants:
+    - prod
+    - ci
+```
+
```
jgehrcke, line 39: Note: here, let's say something like "note: if you wish to add tenants later, ..." and link to the "Adding tenants" guide :)

docs/references/configuration.md
```diff
@@ -0,0 +1,192 @@
...
+*Example:*
+
+```yaml
+env_label: ci
+```
+
+The label name will be `env`, and the value will be what you provide here.
+
+### `data_api_authentication_disabled`
```
jgehrcke, line 117: note to self: rename here or in code or in both

jgehrcke (changes requested):
Thanks @triclambert -- this is a lot of great work that, as discussed, goes into the right direction.

While going through this I noticed a couple of things; some are quite important, others are not. I just typed things down quickly. I don't expect all 'requested changes' to be done before merging.

docs/references/concepts.md
MatApple: Yup, cannot delete the tenant in Postgres: https://github.com/opstrace/opstrace/blob/275cf649e30bda84493f6c6ad9a6ced3d35a4d0b/packages/app/metadata/tables.yaml#L76-L81

Deleting the tenant elsewhere (for example, deleting the namespace), will just result in the controller recreating everything that was deleted to bring the tenant back

docs/guides/administrator/adding-and-managing-tenants.md
triclambert: Leaving out the word "instance" entirely as it is not strictly necessary.

docs/guides/administrator/adding-and-managing-tenants.md
triclambert: This was intended as a placeholder for linking to future content.  Instead, I will add a TODO in a comment.

docs/guides/administrator/adding-and-managing-tenants.md
triclambert: Using them.

In other words, the integration "code" handles the otherwise manual effort of setting the tokens in the correct place.  I will try to adjust the sentence.

docs/guides/administrator/adding-and-managing-tenants.md
triclambert: Good points.  Let's dig into how we want to express these concepts and update the docs in a separate PR.  Let's also consider defining them in the key concepts or a new glossary page.

> "with an Integration" is a difficult language construct, doesn't sound right; too bloaty/abstract.

I don't follow.  An "integration" is a feature of Opstrace; perhaps we should define it in the key concepts?  Perhaps "using an integration"?

docs/guides/administrator/adding-and-managing-tenants.md
triclambert: @MatApple updated the text with a version of what you wrote.  PTAL, but I am going to merge this, but we can update it in a future PR.

triclambert:
> In the admin guide for adding a tenant, we should make it very clear that
> 
> 1. As an admin, one can issue an _arbitrary number_ of authentication tokens from the _same_ RSA private key.
> 2. As an admin, one is _responsible_ for managing the private key.
> 
> About 1: I think this is one of the most important UX aspects to clarify prominently here; because right now the content is misleading: was already a weakness in the blog post -- the content kind of suggests that the intended way to do things is to create a new RSA key pair for _each tenant to be added_. That is really not what should happen.
> 
> About 2: I think that content is covered in the blog post and can be re-used.
> 
> There may be reasons to create more than one RSA key pair; and these reasons will always have to do with security management: for example; if there is more than one admin persona, each of them may want to create an RSA keypair for themselves.

Great points, however, I am going to move this to a separate issue: #839.

docs/guides/administrator/adding-and-managing-users.md
triclambert: > any associated relational data will be deleted

This ☝️ 

However, to err on the side of simplicity, I have removed the sentence entirely.

docs/guides/administrator/adding-and-managing-users.md
triclambert: > we should point out somewhere

Good point.  Perhaps we should move it to the key concepts?  I'll do that in a separate PR.

In the meantime I will modify the sentence with your suggestions.

docs/references/configuration.md
triclambert: This is a longer conversation, but no, I don't think that we just s/cluster/instance/.  The word "instance" is not necessary in many cases.

docs/references/concepts.md
triclambert: Let's defer the architecture bits for now.  I think it's a good idea, but I think that is a larger conversation than can be had here at the moment.

docs/references/concepts.md
triclambert: Yes, just as we discussed offline.  This PR is already rather large and I want to get it merged.

docs/references/concepts.md
triclambert: For now, I will remove the ", secured".

A general term such as "secure" highlights our value proposition without being wordy and slowing the reader down.

docs/references/concepts.md
triclambert: > Is the primary purpose of this sentence to make a shout-out/ an attribution?

I do not think a shout-out is the same thing as attribution.  The purpose of this sentence is to explain where the concept comes from.  In other words, we are clarifying that we did not create it.

docs/references/configuration.md
triclambert: Great idea.

jgehrcke:
@triclambert just noticing that we should have done a `git mv cluster-configuration.md configuration.md` here. The old doc is still on `main`, and we lost history on the new doc.

triclambert:
Whoops, sorry about that.

PR #800, opened 2021-05-30
Feature: View and Edit Cortex Runtime Overrides

MatApple:
This feature allows users to view and edit per-tenant limits and configuration for Cortex.
It includes:
* validation of runtime config to ensure valid config only
* allows to reset changed config back to default cortex config
* shows description of what each config is
* number fields are masked with thousands-commas to make large numbers easier to parse by the human eye
* reports validation and api errors

Example api error:
Screen Shot 2021-05-29 at 5 46 40 PM


test/test-remote/test_ui.ts
```diff
@@ -285,7 +285,7 @@ suite("test_ui_with_headless_browser", function () {
 
     log.info("cookie header value: %s", cookie_header_value);
 
-    const url = `${CLUSTER_BASE_URL}/_/config/cortex/runtime`;
+    const url = `${CLUSTER_BASE_URL}/_/cortex/runtime_config`;
```
MatApple, line 288: @jgehrcke note I changed this to align with the other cortex api handlers

packages/app/src/server/routes/api/cortex/cortexRuntimeConfig.ts
```diff
@@ -105,6 +104,14 @@ export default async function setCortexRuntimeConfigHandler(
   res: Response,
   next: NextFunction
 ) {
+  if (!req.is("text/plain")) {
```
MatApple, line 107: @jgehrcke I enforced the content type too!

packages/app/src/state/cortex-config/sagas/cortexConfigSubscription.ts
```diff
@@ -113,7 +114,12 @@ export function cortexConfigSubscriptionEventChannel(): EventChannel<Actions> {
             emitter(actions.setCortexRuntimeConfig(res.data));
           })
           .catch(err => {
-            emitter(actions.setCortexConfigError(err.toString()));
+            if (ServerError.isInstance(err.response.data)) {
```
MatApple, line 117: client handling of errors: 1) check if the error came from the server and extract the detailed message, 2) fallback to a generic error, usually in the case of a network connection problem.

packages/app/src/server/routes/api/cortex/cortexRuntimeConfig.ts
```diff
@@ -105,6 +104,14 @@ export default async function setCortexRuntimeConfigHandler(
...
+    return next(
+      new GeneralServerError(
```
MatApple, line 109: @jgehrcke by using this error handling technique we can present a consistent error handling mechanism to the client, like this: https://github.com/opstrace/opstrace/pull/800/commits/fa9959acaa1f72957630c19744afd2573b288781#r642102300

packages/app/src/server/routes/api/cortex/cortexRuntimeConfig.ts
```diff
@@ -0,0 +1,235 @@
...
+ * the "observed" runtime config.
+ */
+export async function readCortexRuntimeConfigHandler(
+  req: Request,
+  res: Response,
+  next: NextFunction
+) {
+  if (KUBECONFIG === undefined) {
+    // This will be logged with a stack trace and handled correctly by our error middleware,
+    // issueing a 500 back to the user with the error message below in the resonse body
```
jgehrcke, line 193: Thanks for that comment -- sure that "kubeconfig not set" does end up being written to the HTTP client? In the outest and most general error handler in a web framework one usually does not want to leak error detail to the client, but only a generic error message like "internal server error"; also for security reasons.

packages/app/src/server/routes/api/cortex/index.ts
```diff
@@ -94,6 +97,30 @@ export default function createCortexHandler(): express.Router {
...
+  //     res
+  //   );
+  // });
+
+  // Being able to access this implies having access to cortex config for
+  // all tenants, i.e. this is a privileged / superuser action.
+  cortex.post("/runtime_config", setCortexRuntimeConfigHandler);
+
+  // This is the runtime config as set in the config map
+  cortex.get("/runtime_config_file", readCortexRuntimeConfigHandler);
```
jgehrcke, line 111: `runtime_config_file` suggests we get the contents of the runtime config file; however at least one difference is that we get a JSON document instead of a YAML document -- is that the only difference?

packages/app/src/state/cortex-config/types.ts
```diff
@@ -0,0 +1,287 @@
...
+const isDuration = {
+  name: "is-duration",
+  message: "Must be a valid duration: [number][s|m|h|d|w], e.g. 2d for 2 days.",
+  test: (value: any) =>
+    typeof value === "undefined" || /^\d+[s|m|h|d|w]{1}$/.test(value || "") // allow this field to be undefined or be a valid duration
+};
+
+export const cortexLimitsSchema = yup.object({
+  ingestion_rate: yup.number().integer().min(0, "Cannot be negative").meta({
+    description: "Ingestion rate limit in samples per second for this tenant."
```
jgehrcke, line 28: For making the meaning of this parameter unambiguous, we need to inform about the rate limit strategy being `local`:

> f.StringVar(&l.IngestionRateStrategy, "distributor.ingestion-rate-limit-strategy", "local", "Whether the ingestion rate limit should be applied individually to each distributor instance (local), or evenly shared across the cluster (global).")

We explicitly choose this to be `local`, see 

https://github.com/opstrace/opstrace/blob/17b792ddf72e29f941a33b6e0bba0acaf5c58874/packages/controller/src/resources/cortex/index.ts#L305

I.e. this is a per-ingester limit.

packages/app/src/state/cortex-config/types.ts
```diff
@@ -0,0 +1,287 @@
...
+      description:
+        "This tenant's shard size. A value of 0 disables shuffle sharding for the tenant."
+    }),
+  compactor_blocks_retention_period: yup
+    .string()
+    .meta({
+      description:
+        "Delete blocks containing samples older than the specified retention period. 0 to disable."
+    })
+    .test(isDuration)
```
jgehrcke, line 268: good choice with the `.test(isDuration)`! :)

jgehrcke:
I see that we expose "all available parameters", and leave responsibility for correct usage to the user. That's a fair approach!

But with that approach, there are still some important questions:

- Are we sure all parameters actually take effect? Is a parameter like `enforce_metric_name` really configurable on a per-tenant basis?
- There are 'conflicting' parameters -- do we want to expose those? Example: `max_series_per_user` and `max_global_series_per_user`.
- There are parameters that need additional information to become useful/unambiguous, explained for `ingestion_rate` in a review comment above.


I think one way to introduce these config knobs would have been to sanity-check them one by one, think them through a bit and then document them. But we can also take the reverse direction, happy to help filling the 'description' fields :-)

Regarding potentially helpful Cortex code to understand things, for the record I want to link to 
https://github.com/cortexproject/cortex/blob/94b0137e7133652f32a6ca641510b8c4c861791e/pkg/cortex/runtime_config.go#L25
and 
https://github.com/cortexproject/cortex/blob/94b0137e7133652f32a6ca641510b8c4c861791e/pkg/util/validation/limits.go#L34
which suggests that all of these can really be defined on a per-tenant basis, so this probably answers my question. But seeing it with our own eyes would also be great :-).

jgehrcke:
We should also expose the global ingester limit config via UI: https://github.com/opstrace/opstrace/issues/804

packages/app/src/state/cortex-config/types.ts
MatApple: Yes, good point, thanks!

packages/app/src/server/routes/api/cortex/index.ts
MatApple: I don't think that `runtime_config_file` necessarily implies that the client would get yaml. We could add a content type if a client wanted to get yaml, but I don't see the use case there. Because that file can be edited using kubectl (or numerous other ways), or if we've deployed a newer version of Cortex to an existing cluster that has a different limits schema, I want to catch the source of error early and provide that feedback in CI (during upgrades for example) and also to users. Without validating the read path too, we'd have no guarantee of the data shape on the client, while at the same time relying on a strict data shape and therefore we open the possibility to unknown or unexpected bugs.

MatApple: > may this validation ever remove or add keys transparently?
It could if we wanted it to

packages/app/src/server/routes/api/cortex/cortexRuntimeConfig.ts
MatApple: Yes, this will be written to the HTTP client! In theory, we should never hit this issue because we load the KUBECONFIG at process startup, and we exit(1) if it fails, so the pod would crash loop.

I wasn't concerned with this particular error leaking to the client, quite the opposite - as a System Admin, I'd like to see the error quickly, especially when all this code is open and visible anyway. What would the security concern be in this case?

PR #803, opened 2021-05-31
dashboards: TO cortex: add panel with active metrics count

jgehrcke:
Screenshot from 2021-05-31 14-19-19
Not particularly beautiful, but useful.

### DIFF
--- a/packages/controller/src/resources/monitoring/system/dashboards/tenant-overview-cortex.json
+++ b/packages/controller/src/resources/monitoring/system/dashboards/tenant-overview-cortex.json
@@ -17,8 +17,7 @@
   "editable": true,
   "gnetId": null,
   "graphTooltip": 1,
-  "id": 32,
-  "iteration": 1621599508771,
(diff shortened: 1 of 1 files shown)


MatApple:
@jgehrcke do we need the timestamp in there? Also, as another suggestion, could we use a gauge component with the actual limit as the upper bound (similar to this) but with the absolute value instead of percent:
Screen Shot 2021-05-31 at 11 20 06 AM

jgehrcke:
> do we need the timestamp in there? 

Wanted the timestamp to see the freshness of the value (it's not always "now"). Can change this later if we feel it's needless. In terms of cosmetics: all this looks a bit messy anyway right now.

> could we use a gauge component with the actual limit as the upper bound (similar to this) but with the absolute value instead of percent:

Sure we can try to also show the 'upper bound' (i.e. limit), and then we should also try and see if they use the same counting method.

Merging for now, let's change things on top of this.

PR #806, opened 2021-05-31
ui(test-remote): add test case for ring health

MoSattler:
Adding E2E tests for ring health. 

- Navigates to the health metrics, opens each tab (`Ingester`, `Ruler`, etc), and checks if the table loads.
- It assumes that when at least one row with valid content for the "Last heartbeat" cell is rendered, that the table loaded correctly, and the test passes.
- Excluding Alert Manager for the moment, as there is a problem that prevents it from working right now, see [here](https://opstrace-community.slack.com/archives/C01KZS8JRNC/p1623664061091000).

# Have you...

### DIFF
--- a/test/test-remote/test_ui.ts
+++ b/test/test-remote/test_ui.ts
@@ -363,4 +363,27 @@ suite("test_ui_with_headless_browser", function () {
       await sleep(10);
     }
   });
+
+  test("view_ring_health", async function () {
+    // Same consideration as above: this test is here in this module for now
+    // just because it's easy to use the authentication state after actual
+    // UI-based login.
+    assert(COOKIES_AFTER_LOGIN, "Auth cookies are present");
+    const page = await BROWSER.contexts()[0].newPage();
+    await page.goto(CLUSTER_BASE_URL);
+    await page.waitForNavigation();
+
+    await page.click("text=Health");
+    await page.click("text=Metrics");
+    assert(await page.isVisible("text=Cortex Ring Health"), "page loaded");
+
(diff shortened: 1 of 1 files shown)


test/test-remote/test_ui.ts
```diff
@@ -363,4 +363,34 @@ suite("test_ui_with_headless_browser", function () {
...
+
+  test("view_ring_health", async function () {
+    // Same consideration as above: this test is here in this module for now
+    // just because it's easy to use the authentication state after actual
+    // UI-based login.
+    assert(COOKIES_AFTER_LOGIN);
+    const page = await BROWSER.contexts()[0].newPage();
+    
+    await page.goto(CLUSTER_BASE_URL + "/cluster/health/metrics/ingester");
+    assert(await page.isVisible("text=Cortex Ring Health"));
```
MoSattler, line 375: there might be better ways to assert that this integration is working - but I cannot use the IDs, as some of them are UUIDs. 
Any suggestion, or is this okay for now?

jgehrcke (approved):
Thanks @MoSattler!

I looked at the build log and found
```
...
[2021-06-14T12:22:54Z]     ✓ view_ring_health (13611ms)
```
which is all I need to see for now :) Let's get this in.

edit: oh and maybe next time use a commit message that is prefixed with `test-remote: `.

About `feat(..)`, I want to quote myself from https://github.com/opstrace/opstrace/pull/813#issuecomment-852912445:

>  Also, feat is kind of only meaningful in the context of semantic versioning -- and in there, it would describe the addition/change of a user-facing feature.

PR #808, opened 2021-05-31
ui: use grafana as the source of truth for integration metadata

terrcin:
Currently when we install dashboards for an integration we put them in a unique folder for the integaration, creating it if it's not already present. When we create the folder we store it's ID and path in postgres against the integration. We never subsequently "sync" whats in Grafana with Postgres, ie the user could go and delete the folder and we wouldn't know. 

This PR aims to stop using postgres for storing Grafana metadata and instead just query Grafana directly for that information.


terrcin:
Now that the `integrations` section in redux is storing data from multiple sources this PR is also about being careful with how that's updated - ie so that Postgres updates don't clear Grafata data causing that to be reloaded and then clearing the Postgres data etc..... make a loop. This is why storing the data as id/name indexed objects rather than an array is important as, unlike an array, we can merge partial data easy without effecting unrelated data.

terrcin:
Pretty sure this is good to go, just need to double check something and merge first thing my morning (rather than late at my night)

MatApple:
> This is why storing the data as id/name indexed objects rather than an array is important as, unlike an array, we can merge partial data easy without effecting unrelated data.

You could also store two arrays and keep data for each source separate, then only merge when/if needed in any UI markup

PR #809, opened 2021-06-01
ui: remove EARLY_PREVIEW flag from Export Integrations 

terrcin:
TODO: verify that the status checks still work - tweak as needed, For this I need to book time with @nickbp for testing this with AWS, GCP, and Azure as he's done this before and it'd be waaaay faster than me fumbling about.


terrcin:
This is good to merge once CI passes. I have some further tweaks to do, but they don't need to be in this PR

terrcin:
note: All the exporters stop outputting to logs once they are up and running happily so I had to increase the query time substantially to figure out they were active. A second, and better, pass at this has the status check also checking for active metrics.

terrcin:
image

PR #820, opened 2021-06-02
ui: global redux actions for tenant change

terrcin:
This creates a global action which gets dispatched when the currently selected Tenant is changed by the user - any reducer can respond to it, in this case the Integration reducer resets it's state so as to not containt Integrations for the wrong Tenant

### DIFF
--- a/packages/app/src/client/services/Picker/types.ts
+++ b/packages/app/src/client/services/Picker/types.ts
@@ -18,6 +18,7 @@ import { ReactNode } from "react";
 export type PickerOption = {
   id: string;
   text: string;
+  data?: any;
 };
 
 export type PickerListProps = {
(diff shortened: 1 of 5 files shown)


packages/app/src/client/services/Picker/types.ts
```diff
@@ -18,6 +18,7 @@ import { ReactNode } from "react";
 export type PickerOption = {
   id: string;
   text: string;
+  data?: any;
```
MatApple, line 21: @terrcin  Not important to block this PR, but in general we should try to avoid the `any` type because anything that relies on it downstream will not be type-checked by the compiler for correctness. We lose a lot of static analysis value at build time. Perhaps a generic could be used here

MatApple (approved):

PR #821, opened 2021-06-02
controller: Configure ingesters to not block on readiness (of other ingesters)

nickbp:
The ingester readiness check is based on the ring status of all ingesters.
As a result you can end up with e.g. ingester-1 refusing to deploy because ingester-5 is unhappy.
The StatefulSet will then refuse to deploy ingester-5 because it wants ingester-1 to be ready first.

This breaks the dependency cycle by allowing ingesters to deploy (and/or scale out) in parallel. Note that it does not affect the rollout of config changes, which will remain sequential.

# Have you...

### DIFF
--- a/packages/controller/src/resources/cortex/index.ts
+++ b/packages/controller/src/resources/cortex/index.ts
@@ -1221,7 +1221,9 @@ export function CortexResources(
         spec: {
           serviceName: "ingester",
           replicas: config.ingester.replicas,
-          podManagementPolicy: "OrderedReady",
+          // Avoid circular dependency between ingester-X failing readiness checks due to ingester-Y being down.
+          // In this situation, the StatefulSet will not deploy ingester-Y if the podManagementPolicy is OrderedReady.
+          podManagementPolicy: "Parallel",
           selector: {
             matchLabels: {
               name: "ingester"


nickbp:
@sreis I've been running this change successfully when benchmarking things, including frequently tearing down/restarting ingesters. However I haven't tried it in a fresh install - only against a cluster that was already running with pods that had existing state. Mat mentioned earlier that there might be reasons to keep it sequential? Another solution could be to fix the readiness check somehow to make it not depend on the other ingester nodes all being happy/JOINED.

sreis:
On one hand, this behavior (the readiness check)  seems to be [by design](https://github.com/cortexproject/cortex/blob/4c56545d7d88976627bd9b42b94557ffdace3c34/pkg/ring/lifecycler.go#L204-L207) because:

> CheckReady is used to rate limit the number of ingesters that can be coming or going at any one time,

On the other hand, it works in a fresh install since the tests are green.

What happened when you did a rollout restart of the ingesters? All of them restarted in parallel? Were you sending traffic to the cluster **while** tearing down/restarting the ingesters? 

What is the behavior when you scale up the number of replicas by two or more?

nickbp:
With this change:
- When deploying the ingesters "from scratch", or when increasing or decreasing the `scale` value, they will deploy and tear down in parallel/all at once.
- When deploying an edit to the StatefulSet (e.g. config or args change), the change will be rolled out sequentially/one at a time, starting at `ingester-N` and going to `ingester-0`. It turns out K8s/StatefulSet currently doesn't support deploying edits in parallel at all right now, see [kubernetes#68397](https://github.com/kubernetes/kubernetes/issues/68397). So this doesn't completely avoid the problem in the case of a rollout, where e.g. ingester-6 might block on its readiness check because ingester-3 is sad.
- I've been sending traffic to the ingesters while doing rollouts and it seems like the distributors are ring-aware. If only one ingester is down at a time errors were staying zero but if two ingesters are down at the same time the distributors can complain about not finding an available ingester.

nickbp:
TBH the real fix might be in Cortex - where ideally the readiness check for ingester-X shouldn't depend on whether ingester-Y is happy, only whether ingester-X is happy. Maybe there's another reason for that?

MatApple:
@nickbp @sreis can we take action on this PR first thing next week? FWIW, Grafana's production config for Loki also uses Parallel pod management on the statefulset https://github.com/grafana/loki/blob/main/production/ksonnet/loki/ingester.libsonnet#L58

nickbp:
If they're doing the same thing I'd lean towards merge, any preference @sreis?

nickbp:
Meanwhile it looks like the cortex k8s example has the ingester as a Deployment, which is surprising - I thought ingesters needed storage for WAL? Or did that go away entirely with block storage?: https://github.com/cortexproject/cortex/blob/master/k8s/ingester-dep.yaml

That said, the Deployment they're using would launch all the ingesters in parallel, with `maxUnavailable: 1` for rollouts. `podManagementPolicy: Parallel` would mirror this with new deployment and scale out in parallel, rollouts in sequence.

nickbp:
Out of curiosity I checked what our ingesters were doing and it looks like they indeed are making use of a local volume. So not clear why the cortex example doesn't have a volume at all - maybe it's just for demo purposes or something.

Example from `nick-18nodes` test cluster:
```
~ # du -h /cortex/tsdb/
[...]
14.2G	/cortex/tsdb/default
[...]
289.0M	/cortex/tsdb/system
14.5G	/cortex/tsdb/
```

sreis:
Let's merge it then and I'll add it to cortex-operator.

PR #847, opened 2021-06-07
chore: refresh and other nits for the README

triclambert:
Now that we have releases, we could highlight that the top of the readme with a calver badge:  

badge

### DIFF
--- a/README.md
+++ b/README.md
@@ -99,6 +99,8 @@ Don't forget to clean up if you're done kicking the tires:
 opstrace destroy aws <choose_a_name>
 ```
 
+Note: add the `--region` flag to ensure the CLI can locate any dangling resources, should you run into any sort of trouble cleaning up (for example, after a failed install).
+
 ## Community
 
 Authentic collaboration in a community setting is important to us.
--- a/README.md
+++ b/README.md
@@ -62,15 +62,15 @@ info: Opstrace cluster upgrade done for tracy (aws)
 Alertmanager can be difficult to configure, especially [for Cortex](https://cortexmetrics.io/docs/architecture/#alertmanager) where they correspond with tenants you have created.
 Our goal is to improve this with editors that provide realtime validation feedback:
 
-![alertmanager configuration ui](https://p-aa3d33.f1.n0.cdn.getcloudapp.com/items/RBuYDxD8/a7f04563-71f0-4d18-a5e5-d8ca86663a36.jpg?v=699f44b3f89f096a6bde61e3310169f4)
+![alertmanager configuration ui](https://p95.p4.n0.cdn.getcloudapp.com/items/eDuy9lnR/1a7a1030-4b27-4dfc-bf52-774f9f61d365.jpg?v=d0e8968befa6e0e3e1922594e61f9189)
 
 ### Tenant Management
 
-Tenants provide isolation for logically separate entities.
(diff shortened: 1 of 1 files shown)


triclambert:
FYI, specifically not adding any GIFs or anything to the readme—trying to keep it super basic.

README.md
```diff
@@ -2,8 +2,9 @@
 <!-- markdownlint-disable MD033 -->
 
 ![Build status](https://badge.buildkite.com/df9e995b3a5e4b0bebce8b432b0bf48b092fd261b7017b65c1.svg)
-[![License](https://img.shields.io/github/license/opstrace/opstrace)](https://github.com/opstrace/opstrace/blob/main/LICENSE)
+[![License](https://img.shields.io/github/license/opstrace/opstrace)](LICENSE)
```
MatApple, line 5: does this correctly link to our license still?

README.md
```diff
@@ -2,8 +2,9 @@
...
 [![Contributor Covenant](https://img.shields.io/badge/Contributor%20Covenant-v2.0%20adopted-ff69b4.svg)](CODE_OF_CONDUCT.md)
+[![Calver YYYY.0M.0D](https://img.shields.io/badge/calver-YYYY.0M.0D-22bfda.svg)](https://calver.org)
```
MatApple, line 7: How do we keep the value of the badge up to date?

README.md
```diff
@@ -99,6 +100,8 @@ Don't forget to clean up if you're done kicking the tires:
 opstrace destroy aws <choose_a_name>
 ```
 
+Note: if you run into trouble cleaning up after a failed install, add the `--region` flag to ensure the CLI can locate any dangling resources.
```
MatApple, line 103: Maybe we should remove the part about `if you run into trouble cleaning up after a failed install` since this could also apply if a user exited out of the destroy command after the k8s cluster was destroyed, but before completion.

I'd actually advocate for adding the --region <region> to the default command in these docs (however one is supposed to document optional parameters):
```
opstrace destroy aws <choose_a_name> [--region <region>]
```

README.md
triclambert: Yes.  It just links to the license file on the particular branch, which I think is the more correct behavior.  It matches how we do it with the code of conduct below.

README.md
triclambert: I can remove it. My thinking was to just add some additional help for folks the way things work _right now_.  I'll remove it for now and we can discuss how to handle this offline.

MatApple: I think having that additional help is good - I'd just make it even more prominent (i.e. with less text before you get to the "add the --region flag" part. We already know that people don't read that far (because it's printed in the CLI and still gets missed)

README.md
triclambert: Manually.  The intent of this PR was to propose we see this badge as a way to direct people towards releases.  We could leave the pattern we've chosen there, or we could update it with the latest release.

If we can't decide what to do here, we can certainly remove it for now.

MatApple: Unless this badge is automated, we should remove it - it'll be too easy to forget about updating that little part, besides the release shows up on the right sidebar in Github

README.md
triclambert: Ok, I've updated it—PTAL.

MatApple (approved):
Changes are great - especially happy to have the --region note here! Thanks @triclambert!

PR #829, opened 2021-06-03
cli: change 'cluster name' to 'instance name' in user-facing areas, and also in code here and there

jgehrcke:

PR #849, opened 2021-06-07
chore: update additional user-facing text removing "cluster"

triclambert:
@jgehrcke already updated much of the user-facing CLI output in #829.  While I understand and agree that the log message output is not written for humans, I would like to update the very last lines that come out of the CLI to make them compliant with our new guidelines.  

UPDATE:  aside from making these consistent for our users, it's important to make these consistent for screenshots, gifs, etc.

### DIFF
--- a/packages/cli/src/util.ts
+++ b/packages/cli/src/util.ts
@@ -200,9 +200,7 @@ export function timestampForFilenames(ts: ZonedDateTime): string {
  * - use what was provided via --region on cmdline
  * - error out if --region was not provided
  */
-export async function awsGetClusterRegionWithCmdlineFallback(): Promise<
-  string
-> {
+export async function awsGetClusterRegionWithCmdlineFallback(): Promise<string> {
   if (cli.CLIARGS.region !== "") {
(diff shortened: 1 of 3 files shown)


packages/cli/src/util.ts
```diff
@@ -200,9 +200,7 @@ export function timestampForFilenames(ts: ZonedDateTime): string {
  * - use what was provided via --region on cmdline
  * - error out if --region was not provided
  */
-export async function awsGetClusterRegionWithCmdlineFallback(): Promise<
-  string
-> {
+export async function awsGetClusterRegionWithCmdlineFallback(): Promise<string> {
```
triclambert, line 203: Sorry for the noise, but my editor did this automatically.  Am I missing a linewrap setting or something?

MatApple (approved):

jgehrcke:
@triclambert this patch is great.

One note for the future: please use the commit message prefix `cli: ` for commits like this so that it is obvious from the commit message which part of the code base changed -- thanks : ).

triclambert:
Ah, will do.

PR #850, opened 2021-06-08
[WIP] ui: playwright tests are Chromium only, so only install that version of playwright

terrcin:
Sometimes we're getting CI failures due to installation of playwright failing to download all the browsers, but we're at this stage only using Chromium meaning we're getting failures about things we're not going to use. Yes we need to be doing multi-browser testing, but I have made this temp change to see if it has a positive effect on our CI run time and intermittent failures.

Also added env var for debugging mode, `OPSTRACE_PLAYWRIGHT_DEBUG`, it turns off headless browser mode. Better than editing the code and trying not to commit it.


terrcin:
This passed CI, good to know. In the mean time I figured out this [PR](https://github.com/opstrace/opstrace/pull/851) which is a better work around.

terrcin:
The other PR has been merged, so don't need this one.

Issue #709, opened 2021-05-11
ci instability: golangci-lint: failed to load packages: timed out to load packages


jgehrcke:
```
[2021-05-10T16:42:11Z] (cd go/ && golangci-lint run --allow-parallel-runners)
[2021-05-10T16:43:12Z] ERRO Running error: context loading failed: failed to load packages: timed out to load packages: context deadline exceeded
[2021-05-10T16:43:12Z] ERRO Timeout exceeded: try increasing it by passing --timeout option
[2021-05-10T16:43:12Z] make[1]: *** [Makefile:236: lint-codebase.go] Error 4
[2021-05-10T16:43:12Z] make[1]: Leaving directory '/build'
[2021-05-10T16:43:12Z] make: *** [Makefile:729: preamble] Error 2
```
https://buildkite.com/opstrace/prs/builds/4123#87bc0576-5b96-41f7-9af9-5f9a6433c3b8/952-1109

jgehrcke:
@MatApple saw another occurrence here: https://github.com/opstrace/opstrace/issues/166#issuecomment-850474052

jgehrcke:
```
[2021-06-14T09:48:45Z] # Also see https://github.com/opstrace/opstrace/issues/166
[2021-06-14T09:48:45Z] (cd go/ && golangci-lint run --timeout 3m --allow-parallel-runners)
[2021-06-14T09:52:53Z] ERRO Timeout exceeded: try increasing it by passing --timeout option
```
https://buildkite.com/opstrace/prs/builds/4618#abd5240f-c292-477f-bd1b-700eee750e5a/964-1151

jgehrcke:
https://buildkite.com/opstrace/prs/builds/4620#c9c95727-b394-4a46-96b3-c46e7627de3e/979-1165

jgehrcke:
https://buildkite.com/opstrace/prs/builds/4675#d06e49e1-516f-4e7f-8407-9f67683e289e/978-1171

jgehrcke:
https://buildkite.com/opstrace/prs/builds/4733#db5a74a1-6660-469f-abda-ad6e0b0ba685/979-1177

jgehrcke:
https://buildkite.com/opstrace/prs/builds/4773#3b8ac0b8-10ba-4f46-924b-376faa646666/986-1183

PR #870, opened 2021-06-12
Use Grafana 8.0 and enable Grafana Alerting for Cortex and Loki

MatApple:
This PR:
* Bumps Loki to 2.2.1
* Bumps Grafana to 8.0
* Deploys the Loki Ruler so that rules and alerts can now be evaluated against logs
* Enables the new Grafana Alerting feature, and alerts can be managed for metrics and logs
* Enables Grafana Alerting with the multitenant alertmanager from Cortex so that alerting routes and receivers can be managed in Grafana per tenant
* Adds a default alertmanager configuration for each tenant when the tenant is created

### New alerting via Grafana showing a firing alert against Loki and one against Cortex:
Screen Shot 2021-06-11 at 5 10 35 PM

### Same firing alerts in Alertmanager UI:
Screen Shot 2021-06-11 at 5 10 46 PM

### Alertmanager config management:
Screen Shot 2021-06-11 at 5 06 21 PM


jgehrcke:
Pushed a commit to fix
```
[2021-06-12T00:32:28Z]  | /build/packages/controller/src/resources/monitoring/tenants/grafanaDatasources.ts
[2021-06-12T00:32:28Z]  |   41:9  error  'datasourcesToDelete' is never reassigned. Use 'const' instead  prefer-const
```

packages/controller/src/tasks/syncTenants.ts
```diff
@@ -21,7 +21,58 @@ import { KubeConfig } from "@kubernetes/client-node";
...
+
+  if (res.statusCode > 499) {
+    // Network error, don't potentially overwrite existing config
+    log.error(
+      `could not read alertmanager config for tenant: ${tenant}, got resp: ${res.statusCode}, ${res.statusMessage}`
+    );
+    return;
+  }
+
+  if (res.body.length) {
```
jgehrcke, line 47: if the response is a 404 Not Found one that this might be a false positive -- why not explicitly check for a 2xx response?

packages/controller/src/tasks/syncTenants.ts
```diff
@@ -21,7 +21,58 @@ import { KubeConfig } from "@kubernetes/client-node";
...
+      method: "GET",
+      headers: {
+        "X-Scope-OrgID": tenant
+      },
+      throwHttpErrors: false
+    }
+  );
+
+  if (res.statusCode > 499) {
+    // Network error, don't potentially overwrite existing config
```
jgehrcke, line 40: When we land here we did send the request and got a response. So, "network error" is an unexpected explanation.

packages/controller/src/tasks/syncTenants.ts
```diff
@@ -21,7 +21,58 @@ import { KubeConfig } from "@kubernetes/client-node";
...
+    log.error(
+      `could not read alertmanager config for tenant: ${tenant}, got resp: ${res.statusCode}, ${res.statusMessage}`
+    );
```
jgehrcke, line 43: The condition `if (res.statusCode > 499)` means "handle 5xx responses". What we typically want to do in those cases is to log the response body in a responsible fashion; to see the error detail emitted by the remote end.

I think `${res.statusMessage}` is just a hard-coded translation of the status code, and not a peek into the response body. Maybe log `res.body.slice(0, 500)`.

jgehrcke:
next build failed as of yarnpkg problem (https://github.com/opstrace/opstrace/issues/873), retry triggered

jgehrcke:
next build failed as of another known ci instability: https://github.com/opstrace/opstrace/issues/709#issuecomment-860618334

jgehrcke:
again https://github.com/opstrace/opstrace/issues/873#issuecomment-860637443, retried.

jgehrcke:
CI failed with
```
[2021-06-14T12:58:49Z] 2021-06-14T12:58:49.912Z warning: DNS service client:POST failed: Response code 429 (Too Many Requests)
[2021-06-14T12:58:49Z] 2021-06-14T12:58:49.913Z debug: saw HighLevelRetry error: DNS service client:POST failed
[2021-06-14T12:58:49Z] 2021-06-14T12:58:49.914Z error: 3 attempt(s) failed. Stop retrying. Exit.
[2021-06-14T12:58:49Z] 2021-06-14T12:58:49.914Z debug: shut down logger, then exit with code 1
```

I am now manually cleaning up the CI cluster resource leakage in the DNS service.

packages/controller/src/tasks/syncTenants.ts
MatApple: The multi-tenant Alertmanager in Cortex responds with a 404 for the endpoint if it hasn't been configured yet, I.e. if it hasn't received a POST with a valid config before. 😞

packages/controller/src/tasks/syncTenants.ts
MatApple: I set the `httpclient` to not throw on error, so we can control the flow easier for the case above where a 404 will be returned if alertmanager config the tenant hasn't been set yet.

packages/controller/src/tasks/syncTenants.ts
MatApple: Yes, good point and great suggestion!

MatApple:
Thanks @jgehrcke for the review and helping to get CI passing! 💯

packages/controller/src/tasks/syncTenants.ts
jgehrcke: I am not sure if I got my point across, and I am also not sure if I understand your reply.

Trying my point again: this code block contains the comment `// Already has config set`. Hence, I am assuming that it really is meant to handle a success response, indicating that a config is set for this tenant. Subsequently I assume that this is always wrapped in a 2xx response, probably a 200 one. Why do we not explicitly check for that expected status code?

The criterion implemented right now is `res.body.length` being non-zero. This is fulfilled for any HTTP response having a non-empty body. This is usually the case for all error responses.

As of the `if (res.statusCode > 499)` above, this condition is met for _all responses_ with a non-empty body and a status code `<= 499`, including most 4xx responses. This still seems wrong to me :). If it is correct, it kind of needs better explanation.

That is why I assume that this condition is met for cases where `// Already has config set` is _not_ true.

packages/controller/src/tasks/syncTenants.ts
jgehrcke: Hm, I am not sure if I got my point across.

Again, when we are in this code block it means that we got a response, i.e. no network error happened. Yet, in the code block it says ` // Network error, ...` and I cannot understand what is meant by that.

> I set the httpclient to not throw on error

I had seen that. It just means that an HTTP error response will not result in an error thrown. However, all errors resulting in the request not being sent or in no response being received (transient network errors, leading to DNS resolution errors or TCP connect()/recv() timeout etc) will nevertheless result in an error thrown.

My confusion is resolved when we remove the comment `// Network error, don't potentially overwrite existing config`.

Maybe a better explanation would be

```
// Handle 5xx responses: ambiguous result, we do not know if
// a config has already been set for this tenant or not. Back out.
```

packages/controller/src/tasks/syncTenants.ts
jgehrcke: Maybe we want something like this:

```
if (res.statusCode === 200) {
    // definite signal: a config for this tenant has been set
    return;
}

if (res.statusCode !== 404 {
  // Handle 5xx responses: ambiguous result, we do not know if
  // a config has already been set for this tenant or not. Back out.   
  // Also handle all 4xx responses except for 404; they are unexpected
  // at this point and should be logged properly.
  log.error(...)
  return
}

// When we are here, we know the response had status code 404.
// definite signal that no config for this tenant has been set yet
...do the POST request...

Comment (MatApple) [reply packages/controller/src/tasks/syncTenants.ts:49]: PR with suggested changes: https://github.com/opstrace/opstrace/pull/902

File: packages/controller/src/tasks/syncTenants.ts (hunk 2 above)
Comment (MatApple) [reply packages/controller/src/tasks/syncTenants.ts:40]: PR with suggested changes: https://github.com/opstrace/opstrace/pull/902

File: packages/uninstaller/src/aws.ts [hunk 4]
```diff
@@ -71,6 +71,10 @@ export function* destroyAWSInfra(): Generator<
     clusterName: destroyConfig.clusterName,
     suffix: "loki"
   });
+  const lokiConfigBucketName = getBucketName({
+    clusterName: destroyConfig.clusterName,
+    suffix: "loki-config"
+  });
```
jgehrcke, line 77: ok, good, but we after all forget to trigger the lifecycle-based teardown below where we have

```
  taskGroup4.push(
    yield fork([
      new S3BucketRes(destroyConfig.clusterName, lokiBucketName),
      "teardown"
    ])
  );
```

PR #876, opened 2021-06-14
ui(ring-health): handle more errors

MoSattler:
# Have you...

### DIFF
--- a/packages/app/src/client/views/ring-health/RingTable.tsx
+++ b/packages/app/src/client/views/ring-health/RingTable.tsx
@@ -62,10 +62,30 @@ type Payload = {
   now: string;
 };
 
+type JSONResponse = {
+  headers: {
+    ["content-type"]: "application/json";
+  };
+  data: Payload;
+};
+
+type HTMLResponse = {
+  headers: {
+    ["content-type"]: "text/html; charset=utf-8";
+  };
+  data: string;
+};
+
+type Response = JSONResponse | HTMLResponse;
+
 type Props = {
   ringEndpoint: string;
   baseUrl: string;
 };
+
+function isJSONResponse(resp: Response): resp is JSONResponse {
+  return resp.headers["content-type"] === "application/json";
+}
 
 function getRandomInt() {
   return Math.floor(Math.random() * Math.floor(100000));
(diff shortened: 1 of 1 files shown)


packages/app/src/client/views/ring-health/RingTable.tsx
```diff
@@ -123,13 +124,19 @@ const RingTable = ({ ringEndpoint, baseUrl }: Props) => {
   const fetchShards = async () => {
     try {
       const response = await axios.get<Payload>(ringEndpoint);
+      if (response.headers["content-type"] !== "application/json") {
+        // some error messages are returned via HTML
+        // So we throw an error here to trigger an error toast
```
MoSattler, line 129: For example [here](https://opstrace-community.slack.com/archives/C01KZS8JRNC/p1623664061091000)

packages/app/src/client/views/ring-health/RingTable.tsx
```diff
@@ -78,6 +78,7 @@ const RingTable = ({ ringEndpoint, baseUrl }: Props) => {
     registerNotification,
     unregisterNotification
   } = useNotificationService();
+  const [keepPolling, setKeepPolling] = useState(true);
```
MoSattler, line 81: I added this to interrupt the polling when errors happen. 

Otherwise we just add a bunch of notifications to the UI for each subsequent failure of fetching.

packages/app/src/client/views/ring-health/RingTable.tsx
MatApple: We should at least populate the Error with an appropriate message

MoSattler: It would fall back to the current generic message [here](https://github.com/opstrace/opstrace/pull/876/files#diff-4d85fd5df3e23c5aa25f39c3f608b4cd7e4baf23c713929e440dc7e1701468d0R155).
Should we add a different message for this use case?

MatApple: I think we should display the error message to the user

MatApple: Let's do that in a follow up PR though and get this one landed!

MatApple (approved):

PR #880, opened 2021-06-14
chore: allow for more ingesters when there are more nodes

triclambert:
When scale testing, we explicitly want more ingesters when we add more nodes.  This is a bit of a specific fix for that use case, rather than a well-thought-out strategic change.

### DIFF
--- a/packages/controller/src/resources/cortex/index.ts
+++ b/packages/controller/src/resources/cortex/index.ts
@@ -195,7 +195,13 @@ export function CortexResources(
           choose: 9
         },
         {
+          "<=": 20,
+          // TODO does rounding down to odd matter?
+          choose: roundDownToOdd(getNodeCount(state))
+        },
+        {
           "<=": Infinity,
+          // TODO reconsider whether dividing by 2 is the right thing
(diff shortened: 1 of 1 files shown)


nickbp:
Background:
- On an 18-node test cluster, we found better performance when it ran more ingesters, rather than the current controller value of 9 ingesters.
- In the test we had run 18 ingesters, whereas this would run 17. But going by the existing selection logic, it looks like running an odd number of ingesters is preferable for some reason.

Overall, AFAICT having more ingesters results in better scaling - it behaves as though there's a CPU or mutex (or similar) bottleneck in the ingesters at the moment.

nickbp (approved):

MatApple:
@nickbp there was a preference on running an odd number for quorum purposes. I don't think it's important at that scale.

MatApple:
Also, one thing to note, once you add read load, then you'll have to divide resources in a way where there is room for other components. The scaling function will have to evolve to account for Loki as well.

triclambert:
Build successful; merging before complete CI passes because it shouldn't actually test this change.

PR #888, opened 2021-06-15
ui(ring-health): fix polling

MoSattler:
Superset of #876 

As we poll the shards every few seconds, we occassionaly run into the situation where in between request and response the component unmounts. To prevent errors and warnings like 
> Can't perform a React state update on an unmounted component

we check for the mounted state and cancel any further operation if the component in fact did unmount.

This also fixes previously broken tests, which I reactivated. 

# Have you...


packages/app/package.json
```diff
@@ -111,6 +110,7 @@
     "react-router-dom": "^5.2.0",
     "react-scrollbars-custom": "^4.0.25",
     "react-split-pane": "^0.1.92",
+    "react-use": "^17.2.4",
```
MoSattler, line 113: I replaced the useInterval library with a general hook library [react-use](https://github.com/streamich/react-use), which also includes the `useMounted` hook facilitated in this PR.

MatApple (approved):

PR #892, opened 2021-06-16
Enable dockerhub creds to be applied to all image pulls to avoid anonymous user rate limits

MatApple:
Dockerhub has a pretty aggressive rate limit for unauthenticated image pulling. Large Opstrace installations will require dockerhub credentials for the user to avoid image pull errors due to rate limiting.

https://www.docker.com/increase-rate-limits

During `install` and `upgrade`, the CLI will look for specific env variables:

`OPSTRACE_DOCKERHUB_USERNAME`
`OPSTRACE_DOCKERHUB_TOKEN`

If present, it will use these to create/update the image pull secret.

The controller will then check for the presence of this secret on each reconcile loop and if it exists, update all deployments and statefulsets to use the image pull secret. If the secret does not exist, the controller will not add the image pull secret ref to deployments/statefulsets.

If these env vars do not exist during `create` or `update`, then the secret will not be created/updated.

To delete the image pull secret in an Opstrace instance that has one previously configured, kubectl can be used to delete the secret directly. The controller will then notice the absence of the secret and remove the ref from all deployments/statefulsets.

**When creating an Opstrace instance greater than 5 nodes, the CLI will exit and direct the user to set the appropriate env vars for the DockerHub credentials, if they haven't been set**


packages/cli/src/create.ts
```diff
@@ -40,6 +40,10 @@ import * as ucc from "./ucc";
 import * as util from "./util";
 import * as schemas from "./schemas";
 import { BUILD_INFO } from "@opstrace/buildinfo";
+import {
+  OPSTRACE_DOCKERHUB_TOKEN,
+  OPSTRACE_DOCKERHUB_USERNAME
```
jgehrcke, line 45: confused that this is imported here but seemingly not used in this module. why does that compile?

jgehrcke: oh wow, `OPSTRACE_DOCKERHUB_USERNAME` is the name of the environment variable holding the `OPSTRACE_DOCKERHUB_USERNAME`. Can we maybe work on naming here? :D

packages/controller-config/src/resources/controller.ts
```diff
@@ -41,6 +42,10 @@ export function ControllerResources({
 
   const controllerCmdlineArgs = [`${opstraceClusterName}`];
 
+  // create dockerhub credentials secret if we have the credentials set
+  // as environment variables
+  collection.add(dockerHubCredsSecret(kubeConfig));
```
jgehrcke, line 47: Is this how the installer after all injects this secret into the k8s cluster?

packages/controller-config/src/resources/dockerhub.ts
```diff
@@ -0,0 +1,135 @@
...
+  if (!(username && token)) {
+    log.info(
+      "OPSTRACE_DOCKERHUB_USERNAME, OPSTRACE_DOCKERHUB_TOKEN not present, skip image pull secrets"
+    );
+    return collection;
+  }
+
+  const dockerconfig = `{"auths":{"https://index.docker.io/v2/":{"auth":"${Buffer.from(
+    `${username}:${token}`
+  ).toString("base64")}"}}}`;
```
jgehrcke, line 45: Let's info-log that these credentials are now being communicated into the Opstrace instance via a k8s secret. The log line should contain the Docker Hub username, and the first and last character of the token as a bonus; for debuggability.

packages/controller-config/src/resources/dockerhub.ts
```diff
@@ -0,0 +1,135 @@
...
+const DOCKERHUB_CREDS_NAMESPACE = "kube-system";
+
+export function dockerHubCredsSecret(kubeConfig: KubeConfig) {
+  const collection = new ResourceCollection();
+  const username = process.env[OPSTRACE_DOCKERHUB_USERNAME];
+  const token = process.env[OPSTRACE_DOCKERHUB_TOKEN];
+
+  if (!(username && token)) {
+    log.info(
+      "OPSTRACE_DOCKERHUB_USERNAME, OPSTRACE_DOCKERHUB_TOKEN not present, skip image pull secrets"
```
jgehrcke, line 38: maybe "skip creating image pull secret"?

packages/controller-config/src/resources/dockerhub.ts
jgehrcke: Can we maybe make `const dockerconfig` be a string holding the actual config verbatim?

Can then do the translation of that string into gibberish further below. 

Right now it appears as if we are base64-encoding twice. If that is correct it maybe needs a code comment!

jgehrcke: simpler, clearer, probably more correct:
```typescript
const jsonstring = `{"auths":{"https://index.docker.io/v2/":{"auth":"${username}:${token}"}}}`
```

and then
```
".dockerconfigjson": Buffer.from(jsonstring).toString("base64")
```

jgehrcke:
In the BK build artifacts we find that image pulls failed with `ImagePullBackOff`, e.g.

```
  Warning  Failed   8m11s (x461 over 112m)  kubelet  Error: ImagePullBackOff
  Normal   BackOff  3m10s (x483 over 112m)  kubelet  Back-off pulling image "opstrace/cortex-api:f88f81513f6268e74641c4a7c899f46a9e5ca33f"
```

jgehrcke:
I suppose an important choice in here is that the Docker Hub credentials are
- not passed as part of the user-given Opstrace instance config doc
- not passed as part of the controller config doc

Is that because we want to keep both non-sensitive? I mean, that has always been my goal; would be great to state this in this PR if _that_ is the reason.

As an alternative to passing this in via env vars to `opstrace create ...` we could add a CLI arg to point to a credentials file. Then we do not need to start documenting environment variables yet :).

When I first skimmed through the code I did not find the part in the CLI/installer that actually writes the credentials into the k8s cluster. I then realized that you added this as part of 'controller resources', which the installer deploys into the k8s cluster.

When the installer writes k8s secrets into the k8s cluster I think we should generally aim for doing most/all of that in the same part of code in the same fashion. Here is another example:
https://github.com/opstrace/opstrace/blob/9ae7b7904195a5dd285ff0dfa6c630edc976413e/packages/installer/src/secrets.ts#L32
https://github.com/opstrace/opstrace/blob/9ae7b7904195a5dd285ff0dfa6c630edc976413e/packages/installer/src/index.ts#L283

(just trying to collect instances where we do this; no action needed).

packages/controller-config/src/resources/controller.ts
MatApple: yip, same time as it creates the controller secrets, roles, rolebindings, deployment etc

packages/controller-config/src/resources/dockerhub.ts
MatApple: :+1:

packages/controller-config/src/resources/dockerhub.ts
MatApple: :+1:

MatApple:
> Is that because we want to keep both non-sensitive? I mean, that has always been my goal; would be great to state this in this PR if that is the reason.

Yes, this was the goal! Also wanted to do it in a non-intrusive way as much as possible for now, since we (as devs) of Opstrace are mostly the ones blocked by this.

> As an alternative to passing this in via env vars to opstrace create ... we could add a CLI arg to point to a credentials file. Then we do not need to start documenting environment variables yet :).

This is a good point to raise! We could think about doing this in some follow-up work perhaps? Would love to unblock us on the dev front asap on this :-)

> When the installer writes k8s secrets into the k8s cluster I think we should generally aim for doing most/all of that in the same part of code in the same fashion.

I was thinking the same, but looking more closely now I see we do have two ways, both of which deploy a few things. The installer itself creates those secrets you referenced, and also in `ControllerResources` we deploy some secrets and other things needed by the controller. I chose to take the simplest route where the `installer` and `upgrader` both use this common function `ControllerResources`. I still consider this Secret a "Controller resource", since it's necessary for the controller to manage all other resources, but perhaps we'd benefit from aligning on the best place to be creating the Kubernetes resources during install/upgrade (especially for the reasons you pointed out - the function call was not where you expected it to be!)

packages/cli/src/create.ts
MatApple: Retyping the string literal is more prone to error though. I much prefer a `const` that can be reused!

MatApple: Can rename it if it helps

MatApple: actually, second thoughts, I think renaming it causes more confusion because then it's easy to mistake the name of the const for the name of the env var.

MatApple:
> In the BK build artifacts we find that image pulls failed with ImagePullBackOff, e.g.

Had accidentally added an additional char to the DockerHub token in CI! Token should now be correct :-)

jgehrcke:
> Had accidentally added an additional char to the DockerHub token in CI! Token should now be correct :-)

Glad it was so easy, glad you found that quickly.

>  We could think about doing this in some follow-up work perhaps? Would love to unblock us on the dev front asap on this :-)

Of course, you did the right thing here to just move ahead. It's just always important to keep in mind that this is the kind of interface that (well, at least once documented) is a brutal breaking change when it changes : ).

packages/cli/src/create.ts
jgehrcke: >  I think renaming it causes more confusion because then it's easy to mistake the name of the const for the name of the env var.

Not when using something expressive like `const DOCKERHUB_TOKEN_ENVVAR_NAME` :D (this is really what I was aiming for, should have said that).

well. We will see. :D It was certainly fine to not act on this.

PR #889, opened 2021-06-15
[WIP] run multi browser tests in a dedicated image

terrcin:
This is an evolution of PR https://github.com/opstrace/opstrace/pull/866 running the browser tests in image based on the official Playwright docker image. So no faffing around with getting everything installed correctly etc... Currently the new image works in my local dev env, this PR has been created to test the CI build changes.

PR #866, opened 2021-06-10
[WIP] ui: setup basic multi-browser tests with Playwright

terrcin:
Currently we don't do any browser testing with playwright, but rather only use playwright to get auth credentials to use in some direct API calls. Our current setup doesn't allow for multi-browser testing either which we need to have.

This PR moves to using the [Playwright Test runner](https://playwright.dev/docs/test-intro) which is designed for running multi-browser tests with. This means moving to a different test file structure and separating out Playwright configuration into a dedicated file. Initially there are two simple tests to show that the user can login and navigate the site and see the correct content that are run against Chromium, Firefox, and Webkit

The old placeholder API calling UI tests are there and run still

PR #895, opened 2021-06-16
[wip] ui: Multi-Browser test suite

terrcin:
Currently we don't do any browser testing with playwright, but rather only use playwright to get auth credentials to use in some direct API calls. Our current setup doesn't allow for multi-browser testing either which we need to have.

This PR moves to using the Playwright Test runner which is designed for running multi-browser tests with. This means moving to a different test file structure and separating out actual browser tests into a new section. Initially there are two simple tests to show that the user can login and navigate the site and see the correct content that are run against Chromium, Firefox, and Webkit

Additionally this PR runs the tests in a new docker image based on the official Playwright image that has all the browsers and their dependencies correctly installed. There are no dependencies on the rest of the Opstrace code base except for knowing the CI login credentials.

Todo:
- [ ] Currently Firefox doesn't work at all on CI, works locally
- [ ] Better abilty to debug, including saving of screenshots on failure
- [ ] Developer/contributor documentation
- [ ] Move across the new browser tests that were added to main while this was worked 
- [ ] Add mobile browser testing - simply a project specify user agent and handset
- [ ] Demonstrate good practice by not testing for content but rather dedicated data tags.

These will likely be seperate PRs

note: This PR is based up research through https://github.com/opstrace/opstrace/pull/889 and https://github.com/opstrace/opstrace/pull/866

### DIFF
--- a/ci/cd/run.sh
+++ b/ci/cd/run.sh
@@ -31,3 +31,4 @@ make ci-cd-upgrade-cluster
 make test-remote || true
 make test-remote-looker
 make test-remote-ui
+make test-browser
--- a/.gitignore
+++ b/.gitignore
@@ -58,6 +58,10 @@ packages/buildinfo/src/buildinfo.ts
 
 *.jwt
 
+# playwright output
+test/test-remote/uishot-*
+test/browser/test-results
+
 packages/app/.env
 packages/app/.env.client.development
 packages/app/.env.server.remote
--- a/Makefile
+++ b/Makefile
@@ -680,7 +680,7 @@ test-browser: kubectl-cluster-info
 		--dns $(shell ci/dns_cache.sh) \
 		--workdir /build/test/browser \
 		opstrace/test-browser:$(CHECKOUT_VERSION_STRING) \
-		yarn playwright test --project=Chromium
+		yarn playwright test --workers 1
 
 # Used by CI:
 # three outcomes:
--- a/Makefile
+++ b/Makefile
@@ -540,6 +540,9 @@ rebuild-testrunner-container-images:
 	docker build . --rm --force-rm \
 		-t opstrace/test-remote:$(CHECKOUT_VERSION_STRING) \
 		-f ./test/test-remote/nodejs-testrunner.Dockerfile
+	docker build ./test/browser --rm --force-rm \
+		-t opstrace/test-browser:$(CHECKOUT_VERSION_STRING) \
+		-f ./test/browser/Dockerfile
 	docker pull opstrace/systemlog-fluentd:fe6d0d84-dev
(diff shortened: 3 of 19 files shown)


ci/cd/run.sh
```diff
@@ -31,3 +31,4 @@ make ci-cd-upgrade-cluster
 make test-remote || true
 make test-remote-looker
 make test-remote-ui
+make test-browser
```
jgehrcke, line 34: ignorant question before understanding most of what's going on here : )

How would you describe the responsibility of `make test-remote-ui` vs `make test-browser`? Or is this really part of `[wip]` and should keep this question to myself until this is ready for review? :D

terrcin: @jgehrcke `make test-remote-ui` doesn't (well didn't when I started this work) do browser based testing. It uses playwright to get creds so that it can make api calls. `make test-browser` is actual browser testing, it has zero dependencies on the rest of the code base and is completely stand alone, it's also is testing against multiple browsers as it's using the Playwright Test runner which is designed for that - where as the mocha setup that test-remote-ui uses can't do multi-browser testing without getting hacky.

note: There have been some browser tests put into test-remote-ui in the last week which I'll be helping to move accross to test-browser - should be very straight forward and they get multi-browser testing for free. Also, in a previous version of this PR/change I had renamed `test-remote-ui` to `test-remote-ui-api` to make it more clear what it's for but dropped that for the time being as it got a bit out of scope.

jgehrcke:
> @jgehrcke make test-remote-ui doesn't (well didn't when I started this work) do browser based testing.

@terrcin I can see where you are coming from, but let me try to positively tweak your perspective on history :D There was a time when we _truly_ had **zero** browser-based tests. In that situation, which test would be the first one to build if you had to build one? :-) We really wanted test coverage for the login flow as the most important browser-based test and that is why we built `test-remote-ui`.

That is, `test-remote-ui` was born specifically to test the browser-based login flow.

You are right that we then started to use the outcome of that (authentication proof) to do "browser API" / "UI API" testing in the same module.

As part of your reply I see that you opted for not touching too much of this legacy stuff; which I think is a sane choice towards shipping/landing.

Great work here! And I agree that for pure browser-based testing it's probably super cool to go all in on the playwright-way of doing things.

Like you, I can see that in the future we will have

1) tests in the mochajs-based test runner that use the authentication state created via browser interaction to do "UI API"-testing directly (the 'fixture' work you did is great for that!)

2) tests in the playwright-based test runner doing exclusively browser-driven interaction

It's great that we seem to be aligned on that, and I like/appreciate the idea in

>  I had renamed test-remote-ui to test-remote-ui-api to make it more clear what it's for ...

PR #899, opened 2021-06-16
fix: handle service restarts in grafana datasource splitter

### LINKED ISSUE #896: Bad Gateway errors in Grafana: fixed after restarting grafana-datasource-splitter pods
**Describe the bug**
After e.g. rolling out a cortex change (which redeploys the cortex query-frontend pods), Grafana will start failing its dashboard and logs queries, showing `Bad Gateway` errors.
These errors go away on their own if the corresponding tenant `grafana-datasource-splitter` pod is restarted.
My suspicion is that these splitter pods are caching DNS results, and when cortex or loki pods move to a new pod IP, suddenly the splitter pod can't reach cortex/loki anymore.
It looks like the splitter pods are running nginx internally so this is probably a shortcoming with nginx (non-broken DNS caching might even be a paid feature).

Possible solutions:
- Use something other than nginx that reacts better to loki/cortex query pods changing IPs?
- Use a different kind of K8s Service with an IP that doesn't change?

Example logs when nginx is acting up:
```
$ kubectl logs -n system-tenant grafana-datasource-splitter-cortex-559d4dfdbd-dwj2d
2021/06/16 04:45:22 [error] 22#22: *2830 no live upstreams while connecting to upstream, client: 192.168.42.224, server: _, request: "POST /api/v1/query_range HTTP/1.1", upstream: "http://query-frontend.cortex.svc.cluster.local/api/v1/query_range", host: "grafana-datasource-splitter-cortex.system-tenant.svc.cluster.local"
192.168.42.224 - - [16/Jun/2021:04:45:22 +0000] "POST /api/v1/query_range HTTP/1.1" 502 157 "-" "Grafana/8.0.0"
```

sreis:
Fix #896.

Using a variable in proxy_pass forces re-resolution of the DNS names because NGINX treats variables differently to static configuration. For this, to work we need to also set a resolver but EKS and GKE use different ClusterIPs for the pod DNS upstream resolver so we have to account for that.

### DIFF
--- a/packages/controller/src/resources/monitoring/tenants/grafanaDatasources.ts
+++ b/packages/controller/src/resources/monitoring/tenants/grafanaDatasources.ts
@@ -23,7 +23,7 @@ import {
 } from "@opstrace/kubernetes";
 import { State } from "../../../reducer";
 import { Tenant } from "@opstrace/tenants";
-import { getTenantNamespace } from "../../../helpers";
+import { getTenantNamespace, getControllerConfig } from "../../../helpers";
 import { KubeConfig } from "@kubernetes/client-node";
 import { DockerImages } from "@opstrace/controller-config";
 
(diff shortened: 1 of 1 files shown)


sreis:
I deployed and tested this locally and after multiple restarts of the Cortex query-frontend, I could continue to see the graphs in Grafana and in the Opstrace UI.

MatApple:
Ahh yes, thank you @sreis! 🎉

PR #900, opened 2021-06-16
ui(ring-health): improve HTML error handling

MoSattler:
This PR addresses the problem mentioned [here](https://opstrace-community.slack.com/archives/C01KZS8JRNC/p1623664061091000), where Cortex is returning configuration "errors" via an HTML response. This was partially [addressed in the client](https://github.com/opstrace/opstrace/pull/876).

This client aims to abstract this behaviour away from the client and moves it to the proxy, while also extracting the error message to inform the user about the cause. 

# Have you...

### DIFF
--- a/packages/app/src/server/routes/api/cortex/index.ts
+++ b/packages/app/src/server/routes/api/cortex/index.ts
@@ -33,30 +33,52 @@ export const cortexProxy = httpProxy.createProxyServer({ ignorePath: true });
 cortexProxy.on("proxyReq", onProxyReq);
 
 cortexProxy.on("proxyRes", (proxyRes, req, res) => {
-  const isHTMLResponse = proxyRes.headers["content-type"]?.includes("text/html")
-  const body: Array<any> = [];
-  res.writeHead(proxyRes.statusCode!, proxyRes.headers);
+  // if it is an HTML response, it is most likely an error message
+  const isHTMLErrorResponse = proxyRes.headers["content-type"]?.includes(
+    "text/html"
+  );
+  if (isHTMLErrorResponse) {
+  } else {
+    res.writeHead(proxyRes.statusCode!, proxyRes.headers);
+  }
+
+  const chunks: Array<any> = [];
   proxyRes.on("data", function (chunk) {
-    if (!isHTMLResponse) {
-      res.write(chunk)
+    if (isHTMLErrorResponse) {
+      /* In case of HTML error, we do not want to stream the response directly
+       * to the client, but instead extract the error message in `proxyRes.on("end")`
+       * and send that.
+       */
+      chunks.push(chunk);
+    } else {
+      res.write(chunk);
     }
-    body.push(chunk)
   });
   proxyRes.on("end", function () {
-    if (isHTMLResponse) {
-      /* Some error messages are returned via HTML.
-       * We have mainly seen it with configuration errors, e.g. when sharding 
-       * is disabled. 
-       *
-       * For example HTML responses please check the tests.
+    if (isHTMLErrorResponse) {
+      /* Error messages are returned via HTML via a specific layout.
+       * See examples here:
+       * - https://github.com/cortexproject/cortex/blob/82b32ec65ed16920e6053ac7fb748c42e3cae452/pkg/storegateway/gateway_http.go#L16-L25
+       * - https://github.com/cortexproject/cortex/blob/82b32ec65ed16920e6053ac7fb748c42e3cae452/pkg/compactor/compactor_http.go#L16-L25
        */
+      const body = chunks.join();
       const parsedHTML = parse(body);
-      const data = parsedHTML.querySelector("p").text;
-      res.end(data);
+      const errorMessage = parsedHTML.querySelector("html body p")?.text;
+
+      /* setting the status code as 412, as usually HTML error responses
+       * of cortex are due to misconfigurations
+       */
+      res.writeHead(errorMessage ? 412 : 500, {
+        "content-type": "text/plain"
+      });
+      res.end(errorMessage ?? "An unknown error occured.");
     } else {
+      /* if response is not HTML, the chunks have already been written
+       * in `proxyRes.on("data")`, no need to add it to `res.end`
+       */
       res.end();
     }
-  })
+  });
 });
 
 const proxyTo = (target: string) => (
--- a/packages/app/src/server/routes/api/cortex/index.ts
+++ b/packages/app/src/server/routes/api/cortex/index.ts
@@ -21,6 +21,7 @@ import { log } from "@opstrace/utils/lib/log";
 import setCortexRuntimeConfigHandler, {
   readCortexRuntimeConfigHandler
 } from "./cortexRuntimeConfig";
+import { parse } from "node-html-parser";
 
 export const cortexProxy = httpProxy.createProxyServer({ ignorePath: true });
 /**
@@ -31,6 +32,33 @@ export const cortexProxy = httpProxy.createProxyServer({ ignorePath: true });
  */
 cortexProxy.on("proxyReq", onProxyReq);
 
+cortexProxy.on("proxyRes", (proxyRes, req, res) => {
+  const isHTMLResponse = proxyRes.headers["content-type"]?.includes("text/html")
+  const body: Array<any> = [];
+  res.writeHead(proxyRes.statusCode!, proxyRes.headers);
+  proxyRes.on("data", function (chunk) {
+    if (!isHTMLResponse) {
+      res.write(chunk)
+    }
+    body.push(chunk)
+  });
+  proxyRes.on("end", function () {
+    if (isHTMLResponse) {
+      /* Some error messages are returned via HTML.
+       * We have mainly seen it with configuration errors, e.g. when sharding 
+       * is disabled. 
+       *
+       * For example HTML responses please check the tests.
+       */
+      const parsedHTML = parse(body);
+      const data = parsedHTML.querySelector("p").text;
+      res.end(data);
+    } else {
+      res.end();
+    }
+  })
+});
+
 const proxyTo = (target: string) => (
   req: express.Request,
   res: express.Response
@@ -39,6 +67,7 @@ const proxyTo = (target: string) => (
     req,
     res,
     {
+      selfHandleResponse: true,
       target,
       headers: req.params.tenant
         ? {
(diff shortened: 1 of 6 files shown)


packages/app/src/server/routes/api/cortex/index.ts
```diff
@@ -31,6 +32,28 @@ export const cortexProxy = httpProxy.createProxyServer({ ignorePath: true });
...
+  });
+  proxyRes.on("end", function () {
+    if (proxyRes.headers["content-type"]?.includes("text/html")) {
+      /* some error messages are returned via HTML and status code 200 OK.
+       * We have mainly seen it with configuration errors,
+       * e.g. sharding is disabled. Hence we set the status code 
+       * to 412 Precondition Failed.
+       * 
+       * For example HTML responses please check the tests.
+       */
```
MoSattler, line 48: Cortex might return other errors via HTML, which would not fall into the category of misconfiguration. But I do not know.

packages/app/src/server/routes/api/cortex/index.ts
```diff
@@ -31,6 +32,28 @@ export const cortexProxy = httpProxy.createProxyServer({ ignorePath: true });
...
+      const parsedHTML = parse(data)
+      data = parsedHTML.querySelector("p").text
+      res.statusCode = 412
```
MoSattler, line 51: This assumes that *all* error messages are in the format [we have seen so far.](https://opstrace-community.slack.com/archives/C01KZS8JRNC/p1623664061091000)

packages/app/src/server/routes/api/cortex/index.ts
```diff
@@ -31,6 +32,34 @@ export const cortexProxy = httpProxy.createProxyServer({ ignorePath: true });
  */
 cortexProxy.on("proxyReq", onProxyReq);
 
+cortexProxy.on("proxyRes", (proxyRes, req, res) => {
+  const isHTMLResponse = proxyRes.headers["content-type"]?.includes("text/html")
+  let data: any = undefined;
+  const body: Array<any> = [];
+  res.writeHead(proxyRes.statusCode!, proxyRes.headers);
```
MoSattler, line 39: I am not 100% if just taking the headers from the response might have any unexpected side-effects. It worked fine during testing in this case, but if anyone knows more, please let me know!

packages/app/src/server/routes/api/cortex/index.ts
```diff
@@ -39,6 +68,7 @@ const proxyTo = (target: string) => (
     req,
     res,
     {
+      selfHandleResponse: true,
```
MoSattler, line 71: https://github.com/http-party/node-http-proxy#miscellaneous

packages/app/src/server/routes/api/cortex/index.ts
```diff
@@ -31,6 +32,34 @@ export const cortexProxy = httpProxy.createProxyServer({ ignorePath: true });
...
+cortexProxy.on("proxyRes", (proxyRes, req, res) => {
+  const isHTMLResponse = proxyRes.headers["content-type"]?.includes("text/html")
+  let data: any = undefined;
+  const body: Array<any> = [];
+  res.writeHead(proxyRes.statusCode!, proxyRes.headers);
+  proxyRes.on("data", function (chunk) {
+    if (!isHTMLResponse) {
+      res.write(chunk)
+    }
+    body.push(chunk)
```
MoSattler, line 44: If the response is HTML, it's not sent to the client, but rather collected in `body`, [to be later modified](https://github.com/opstrace/opstrace/pull/900/files#diff-4ba3f8112ae1c4d28b5397690668351095b7568556ad0bce0d6e2bf0b067d4dbR47-R57).

jgehrcke:
You asked on Slack for some eyes on this, and I'll give this a quick shot :)

It's worth looking into Cortex code so that we understand better The way Cortex generated the output is extremely basic and naive right now. Example: https://github.com/cortexproject/cortex/blob/82b32ec65ed16920e6053ac7fb748c42e3cae452/pkg/compactor/compactor_http.go#L14

Are we sure that there is no 'JSON API' or something like that to get to the same kind of error information? I guess the answer is 'yes'.

An architecturally more sane choice in Cortex would be to look into the request and inspect the `Accept` header -- which could be either `text/html` (sent by a browser) or `application/json` (sent by custom code).

It's OK to not overthink this and simply rather _precisely_ parse the current HTML error output. I would instead of
```
parsedHTML.querySelector("p").text;
```
extract `html->body->p` (just to enhance precision a bit), and in the code also link to the relevant places in the Cortex code base where these HTML templates are defined. I found two:

https://github.com/cortexproject/cortex/blob/82b32ec65ed16920e6053ac7fb748c42e3cae452/pkg/storegateway/gateway_http.go#L13
https://github.com/cortexproject/cortex/blob/82b32ec65ed16920e6053ac7fb748c42e3cae452/pkg/compactor/compactor_http.go#L13

I guess we're looking for a third one.

Given the simplicity of the HTML response I think we're also good with copying the response headers as they are. I don't see how this could be dangerous right now. However, if we do not have a specific reason to do this, then maybe just don't do it for now. I think we could also generate our own set of headers in the proxy. Because: once the proxy is sure that it got an HTML response containing an error message, it knows everything about the response it wants to generate for its own client. 

In the proxy you maybe want to prepare for the case where the response coming in from Cortex is unexpected, and then generate a 500 Internal Server Error response for our client -- with good error information in the body.

These are just a few quick thoughts, happy to look again later.

packages/app/src/server/routes/api/cortex/index.ts
```diff
@@ -31,6 +32,33 @@ export const cortexProxy = httpProxy.createProxyServer({ ignorePath: true });
  */
 cortexProxy.on("proxyReq", onProxyReq);
 
+cortexProxy.on("proxyRes", (proxyRes, req, res) => {
+  const isHTMLResponse = proxyRes.headers["content-type"]?.includes("text/html")
+  const body: Array<any> = [];
```
jgehrcke, line 37: maybe rename to `bodychunks` or so :)

packages/app/src/server/routes/api/cortex/index.ts
```diff
@@ -31,6 +32,33 @@ export const cortexProxy = httpProxy.createProxyServer({ ignorePath: true });
...
+       * We have mainly seen it with configuration errors, e.g. when sharding 
+       * is disabled. 
+       *
+       * For example HTML responses please check the tests.
+       */
+      const parsedHTML = parse(body);
+      const data = parsedHTML.querySelector("p").text;
+      res.end(data);
+    } else {
+      res.end();
```
jgehrcke, line 57: Does this mean we write a response with an empty body?

MoSattler: No, because in the case of `!isHTMLResponse` we already stream the response to the client in [proxyRes.on("data")](https://github.com/opstrace/opstrace/pull/900/files#diff-4ba3f8112ae1c4d28b5397690668351095b7568556ad0bce0d6e2bf0b067d4dbR41).

MoSattler:
In case we want to keep going with the current approach, I kept working on it. Should be good for another review!

MoSattler:
> Given the simplicity of the HTML response I think we're also good with copying the response headers as they are. I don't see how this could be dangerous right now. However, if we do not have a specific reason to do this, then maybe just don't do it for now. 
> 
> I think we could also generate our own set of headers in the proxy. Because: once the proxy is sure that it got an HTML response containing an error message, it knows everything about the response it wants to generate for its own client. 

The main place where we want to copy the headers are response which are not errors/html. There are many cases where we need to do it. E.g. we need to indicate to the browser if a response is compressed or not. Though for now I changed it a bit, and added manual headers which seems to work right now. This might break in the future though, if cortex has breaking changes to the API upstream.  But most of the potential breakage should be caught by our E2E tests.

MoSattler:
> In the proxy you maybe want to prepare for the case where the response coming in from Cortex is of an unexpected HTML format/structure, and then generate a 500 Internal Server Error response for our client -- with good error information in the body.

Can you elaborate what you mean by good error information in this case? I am not sure if we can really tell too much why we couldn't extract an error message besides "something went wrong".

packages/app/src/server/routes/api/cortex/index.ts
```diff
@@ -31,6 +32,55 @@ export const cortexProxy = httpProxy.createProxyServer({ ignorePath: true });
...
+       * of cortex are due to misconfigurations
+       */
+      res.writeHead(errorMessage ? 412 : 500, {
+        "content-type": "text/plain"
+      });
+      res.end(errorMessage ?? "An unknown error occured.");
+    } else {
+      /* if response is not HTML, the chunks have already been written
+       * in `proxyRes.on("data")`, no need to add it to `res.end`
+       */
```
jgehrcke, line 78: thanks for having added the comment here!

packages/app/src/server/routes/api/cortex/index.ts
```diff
@@ -31,6 +32,55 @@ export const cortexProxy = httpProxy.createProxyServer({ ignorePath: true });
...
+  proxyRes.on("end", function () {
+    if (isHTMLErrorResponse) {
+      /* Error messages are returned via HTML via a specific layout.
+       * See examples here:
+       * - https://github.com/cortexproject/cortex/blob/82b32ec65ed16920e6053ac7fb748c42e3cae452/pkg/storegateway/gateway_http.go#L16-L25
+       * - https://github.com/cortexproject/cortex/blob/82b32ec65ed16920e6053ac7fb748c42e3cae452/pkg/compactor/compactor_http.go#L16-L25
+       */
+      const body = chunks.join();
+      const parsedHTML = parse(body);
+      const errorMessage = parsedHTML.querySelector("html body p")?.text;
```
jgehrcke, line 66: thanks for the changes : ) when code reads like an English sentence that says what it's doing w/o having to think a lot: :rocket:

jgehrcke:
>  Can you elaborate what you mean by good error information in this case? I am not sure if we can really tell too much why we couldn't extract an error message besides "something went wrong".

Good question :). Imagine a user triggers this behavior and gets to see _something_ in the UI. Now imagine they will report that _something_ to _you_. Take it from there :) If "something went wrong" is going to quickly help you understand which scenario the user ran into: :+1: -- if you feel like there is more you want to add: better. For example, a "Error: Cortex responded with an unexpected HTML response, the first 500 bytes are ...:" -- I am really not thinking towards a specific solution right now, just want to give an impression of what I believe matters at the high level:
- optimize for debuggability while not making the UX a mess (I think that's a fair set of priorities at current product stage)
- imagine the likelihood for Cortex to respond with a special HTML response is non-zero, but practically zero. When this happens 'in the field' then this is an extremely valuable data point and we would like to find out as much as we can about that.

MoSattler:
After talking to @jgehrcke, we agreed to merge this PR, and address his comments in a separate PR!

PR #902, opened 2021-06-16
improve the setting of default Alertmanager config for tenants during tenant sync

MatApple:
Following up on comments left by @jgehrcke:

https://github.com/opstrace/opstrace/pull/870#discussion_r651805486
https://github.com/opstrace/opstrace/pull/870#discussion_r651795390

### DIFF
--- a/packages/controller/src/tasks/syncTenants.ts
+++ b/packages/controller/src/tasks/syncTenants.ts
@@ -34,12 +34,16 @@ function* setDefaultAlertmanagerConfigIfEmpty(tenant: string) {
       }
     });
 
-    if (res.data.length) {
+    if (res.status === 200) {
       // Already has config set
       return;
     }
   } catch (err) {
     if (!err.response) {
+      log.error(
+        `could not read alertmanager config for tenant: ${tenant}, got unknown error: %s`,
+        err
+      );
       return;
     }
     // Cortex will return a 404 if config hasn't been set for this tenant already, so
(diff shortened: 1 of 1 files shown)


packages/controller/src/tasks/syncTenants.ts
```diff
@@ -34,12 +34,16 @@ function* setDefaultAlertmanagerConfigIfEmpty(tenant: string) {
...
     });
 
-    if (res.data.length) {
+    if (res.status === 200) {
       // Already has config set
       return;
     }
   } catch (err) {
     if (!err.response) {
+      log.error(
```
jgehrcke, line 43: surprised that this is axios again, and not the `got`-based httpcl -- I see you switched here: https://github.com/opstrace/opstrace/pull/882. Don't want to consume too much time on this, but I think it's important to understand why you did the switch back to axios. Don't want to make all the points again; but a super quick reminder: with `httpcl` we get custom logging and retrying behavior that is _super_ valuable. Also, connect() and recv() timeout control that axios will never be able to provide, ...

MatApple: I really wanted to use the httpcl and tried to use it! But for some reason, I just couldn't get it to work and I never managed to get to the bottom of why. Opted to just move forward with what did work

jgehrcke: Oh, I didn't see this response before @MatApple. Thanks for the follow-up and we'll get there :).

PR #907, opened 2021-06-17
controller: increase default PVC size

triclambert:
Perhaps we should increase the PVC size to fit more of our workloads.  PTAL.

### DIFF
--- a/packages/controller/src/resources/cortex/index.ts
+++ b/packages/controller/src/resources/cortex/index.ts
@@ -1323,7 +1323,7 @@ export function CortexResources(
                 accessModes: ["ReadWriteOnce"],
                 resources: {
                   requests: {
-                    storage: "30Gi"
+                    storage: "100Gi"
                   }
                 }
               }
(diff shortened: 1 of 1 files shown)


jgehrcke:
Would love to see a super brief discussion of maybe upsides and downsides in the PR, and ideally also in the commit message. I trust that we then understand / kept record of the downsides, and we could then merge this.

To keep track of reasoning/decision-making, this is the previous commit touching this: https://github.com/opstrace/opstrace/commit/0b388e05ab0d8c1c6851f7b05f0621b8f7c8d33e and PR: https://github.com/opstrace/opstrace/pull/413

Let's think about which need the default should serve primarily (it can't serve all the needs).

And maybe let's think about if we instead want to expose this parameter nicely to users in the instance config file (instead of finding a 'magic formula'). We could then set a really small value for CI (to save cost, if that matters) and pick a larger default value.

triclambert:
Closing as overcome by events.  The default is sufficient for most workloads, and can be increased if needed.

PR #912, opened 2021-06-18
[WIP] ui: fix firefox browser tests

terrcin:
When setting up using the Playwright Test running the tests would timeout with Firefox, so I disabled them to get the main work out. This PR is about getting Firefox browser tests working.


terrcin:
Note: the timeout failures encountered on CI with Firefox I've not managed to reproduce locally on OSX, Ubuntu, or via my own build CI docker image. Good excuse to add it further debugging abilitiy.

terrcin:
pretty sure this is todo with the sidebar autohiding. I can't replicate the faliure with Firefox, but I can with iPhone-11. When in headed mode if I manually move the curosor over the sidebar making the items appear the test then passes.

terrcin:
I could reproduce these (well, same failure atleast) by changing the Firefox viewport to be only 500px wide.

I started getting problems with the `text=abcd` selectors not working - text was on screen so not sure what was with that. Should not be testing against content though (as tests shouldn't break when content changes) so switched it all to data attribute tests (this is also how you read entity IDs for later tests). Briefly frustrated by [playwright limitations around data attributes in selectors](https://playwright.dev/docs/selectors/#id-data-testid-data-test-id-data-test-selectors), but it's all working locally now including the hover to bring out the sidebar.

terrcin:
Was included as part of https://github.com/opstrace/opstrace/pull/949

Issue #908, opened 2021-06-17
Grafana promql queries fail after upgrade from 6cb91909-ci to latest main


sreis:
**Describe the bug**

All the promql queries are getting a 301 and a redirect to grafana/login and then fail. Yes, I've cleared the cache, cookies and browser local storage. 

![Screenshot from 2021-06-17 15-17-07](https://user-images.githubusercontent.com/495294/122426411-c2260580-cf7f-11eb-97f7-35cce688b4bb.png)

This doesn't happen with metrics although we get a 404 for the query_examplars endpoint.

![Screenshot from 2021-06-17 15-23-25](https://user-images.githubusercontent.com/495294/122426723-00bbc000-cf80-11eb-8eae-a81b7fc34f14.png)

**To Reproduce**

Upgrade Opstrace instance starting at 6cb91909-ci to latest main.

MatApple:
@sreis this shouldn't be happening but I'm guessing Grafana 8 has changed some config settings around authentication. We use Grafana in proxy mode but it appears the Grafana instance thinks it's "not in proxy mode" because it's redirecting to its own login. Perhaps there is an announcement in the Grafana 8 release notes that we missed.

Here are the config options for Grafana https://grafana.com/docs/grafana/latest/administration/configuration/

sreis:
@MatApple One strange thing is that the CI/CD cluster was upgraded this morning and does not have this problem.

sreis:
Answer to myself, the `query_examplars` that is returning a 404 is a [Prometheus endpoint](https://prometheus.io/docs/prometheus/latest/querying/api/#querying-exemplars) to correlate metrics with traces. AFAIK not yet available in Cortex.

jgehrcke:
In the very first screenshot for the very **first** 301 response it would be great to show more details:
- `Location` header of that response
- _all_ other aspects of that response so that we understand better (or _precisely_) which entity really emitted it

MatApple:
It should be noted that we turn `proxy_redirect` off, which is the same setting as nginx-ingress controller uses.

https://github.com/opstrace/opstrace/blob/5ed454ec96191dc4bea278dc019128b6848f2be6/packages/controller/src/resources/monitoring/tenants/grafanaDatasources.ts#L182

MatApple:
[...]

Issue #873, opened 2021-06-14
ci instability: yarn install exits with ESOCKETTIMEDOUT


jgehrcke:
```
install yarn workspace dependencies
...
[3/5] Fetching packages...
info There appears to be trouble with your network connection. Retrying...
info There appears to be trouble with your network connection. Retrying...
info There appears to be trouble with your network connection. Retrying...
info There appears to be trouble with your network connection. Retrying...
info There appears to be trouble with your network connection. Retrying...
error An unexpected error occurred: "https://registry.yarnpkg.com/@material-ui/icons/-/icons-4.11.2.tgz: ESOCKETTIMEDOUT".
```
https://buildkite.com/opstrace/prs/builds/4614#06d0a6e6-bcd5-4860-b4bf-a941dc7ef00f/916-927

jgehrcke:
https://buildkite.com/opstrace/prs/builds/4619#8fe084d0-4afc-410c-a4b0-5363b2a1c521/925-935

jgehrcke:
https://buildkite.com/opstrace/prs/builds/4620#8d93d91f-d05c-4a2b-9cb1-c6edcf9c45ca/946-961

jgehrcke:
A manual build retry has helped. That is, this does not happen _every_ time, but 'just' with elevated frequency. Let's hope that this rate drops again asap.

jgehrcke:
https://buildkite.com/opstrace/prs/builds/4631#7128a1e7-dd1f-4144-b60e-e924eb105b5e/924-934

jgehrcke:
https://buildkite.com/opstrace/prs/builds/4733#3a6bb9a9-3665-444e-8cce-e345e9f82c33/933-943

MoSattler:
https://buildkite.com/opstrace/prs/builds/4839#53fa0140-1a7e-42cb-96f1-1e48dc9a0c12/925-931

MoSattler:
https://buildkite.com/opstrace/prs/builds/4839#53fa0140-1a7e-42cb-96f1-1e48dc9a0c12/925-931

nickbp:
Lots of failures in mid-late June followed by no repros, feels like an upstream issue

PR #914, opened 2021-06-18
Try to fix nginx datasource splitter reverse proxy config

jgehrcke:
Let's look at
```
            location / {
        ...
              set $backend http://querier.loki.svc.cluster.local:1080/$request_uri;
              proxy_pass $backend;
            }
```

### Special case: `location /`

Let's distinguish four cases.

* This adds a double slash from Loki's point of view, because `$request_uri` always starts with a slash (probably a cause for #908, see https://github.com/opstrace/opstrace/issues/908#issuecomment-863960135):
```
set $backend http://querier.loki.svc.cluster.local:1080/$request_uri;
```

* This always presents just `/` to Loki, forgets about the incoming URL path:
```
set $backend http://querier.loki.svc.cluster.local:1080/;
```

* These two are correct and equivalent for the special case of `location /`:
```
set $backend http://querier.loki.svc.cluster.local:1080;
```
```
set $backend http://querier.loki.svc.cluster.local:1080$request_uri;
```

### Other cases: `location /foo/bar`

Let's look at

```
            location /prometheus/api/v1/rules/ {
              set $backend http://ruler.loki.svc.cluster.local:1080/prometheus/api/v1/rules$request_uri;
              proxy_pass $backend;
            }
```

`$request_uri` is documented as full original request URI (with arguments).

When there is a request incoming at `/prometheus/api/v1/rules/foo/bar` then this location handler handles the request and `$request_uri` is `/prometheus/api/v1/rules/foo/bar`.

The request is then passed on to Loki with the URL

`http://ruler.loki.svc.cluster.local:1080/prometheus/api/v1/rules/prometheus/api/v1/rules/foo/bar`;

Clearly bogus. If the goal is to rewrite parts of the URL and to use the DNS caching workaround one cannot rely on the proxy_pass URL rewriting behavior documented at http://nginx.org/en/docs/http/ngx_http_proxy_module.html#proxy_pass

That would be this:

> If the proxy_pass directive is specified with a URI, then when a request is passed to the server, the part of a normalized request URI matching the location is *replaced* by a URI specified in the directive:

(emphasis mine)

That replacement does not work when using variables in the proxy_pass target URL. The generic solution is to extract the relevant portion in the incoming URL with regex in the location directive and to re-use that in the proxy_pass statement. Example:
```
location ~ ^/system/v1/leader/mesos(?<url>.*)$ {
    ...
    set $system_v1_leader_mesos "http://leader.mesos/system/v1$url$is_args$query_string";
    proxy_pass $system_v1_leader_mesos;
}
```
(from DC/OS' Admin Router). See the `(?<url>.*)` bit in the first line. Reused that via `$url` in the proxy_pass target.

When one only wants to carry over query args, one can use `$is_args$query_string`

### DIFF
--- a/packages/controller/src/resources/monitoring/tenants/grafanaDatasources.ts
+++ b/packages/controller/src/resources/monitoring/tenants/grafanaDatasources.ts
@@ -251,6 +251,8 @@ http {
         data: {
           "nginx.conf": getNginxConfig(
             `
+
+            # Pass most requests on to the querier
             location / {
               ${generalProxyConfig}
 
@@ -259,36 +261,25 @@ http {
               proxy_pass $backend;
             }
 
+            # For specific routes, pass on to the ruler
             location /prometheus/api/v1/rules {
               ${generalProxyConfig}
-
               resolver ${dnsResolver};
-              set $backend http://ruler.loki.svc.cluster.local:1080/prometheus/api/v1/rules$is_args$query_string;
-              proxy_pass $backend;
-            }
-
-            location /prometheus/api/v1/rules/ {
-              ${generalProxyConfig}
-
-              resolver ${dnsResolver};
-              set $backend http://ruler.loki.svc.cluster.local:1080/prometheus/api/v1/rules$is_args$query_string;
+              # Do not change the request URL
+              set $backend http://ruler.loki.svc.cluster.local:1080;
               proxy_pass $backend;
             }
 
             location /prometheus/api/v1/alerts {
               ${generalProxyConfig}
-
               resolver ${dnsResolver};
-              set $backend http://ruler.loki.svc.cluster.local:1080/prometheus/api/v1/alerts$is_args$query_string;
+              # Do not change the request URL
+              set $backend http://ruler.loki.svc.cluster.local:1080;
               proxy_pass $backend;
             }
 
-
-            # Loki specific ruler routes
-
             location /loki/api/v1/rules {
               ${generalProxyConfig}
-
               resolver ${dnsResolver};
               # Do not change the request URL, this works for the following endpoints:
               # /loki/api/v1/rules
@@ -300,7 +291,6 @@ http {
             # Loki legacy route for saving rules (legacy routes which the Grafana 8 alerting feature uses)
             location ~ ^/api/prom/rules(?<urlsuffix>.*)$
               ${generalProxyConfig}
-
               resolver ${dnsResolver};
               set $backend http://ruler.loki.svc.cluster.local:1080/loki/api/v1/rules$urlsuffix;
               proxy_pass $backend;
--- a/packages/controller/src/resources/monitoring/tenants/grafanaDatasources.ts
+++ b/packages/controller/src/resources/monitoring/tenants/grafanaDatasources.ts
@@ -317,61 +317,30 @@ http {
             `
           location / {
             ${generalProxyConfig}
-
             resolver ${dnsResolver};
             set $backend "http://query-frontend.cortex.svc.cluster.local:80";
             proxy_pass $backend;
           }
 
-          location /api/v1/rules {
+          location ~ ^/api/v1/rules(?<urlsuffix>.*)$
             ${generalProxyConfig}
-
             resolver ${dnsResolver};
-            set $backend "http://ruler.cortex.svc.cluster.local/prometheus/api/v1/rules$is_args$query_string";
-            proxy_pass $backend;
-
-          }
-
-          location /api/v1/rules/ {
-            ${generalProxyConfig}
-
-
-            resolver ${dnsResolver};
-            set $backend http://ruler.cortex.svc.cluster.local/prometheus/api/v1/rules$is_args$query_string;
-
-          }
-
-          location /api/v1/alerts {
-            ${generalProxyConfig}
-
-            resolver ${dnsResolver};
-            set $backend "http://ruler.cortex.svc.cluster.local/prometheus/api/v1/alerts$is_args$query_string";
+            set $backend "http://ruler.cortex.svc.cluster.local/prometheus/api/v1/rules$urlsuffix";
             proxy_pass $backend;
           }
 
-          location /api/v1/alerts/ {
+          location ~ ^/api/v1/alerts(?<urlsuffix>.*)$
             ${generalProxyConfig}
-
             resolver ${dnsResolver};
-            set $backend http://ruler.cortex.svc.cluster.local/prometheus/api/v1/alerts$is_args$query_string;
+            set $backend "http://ruler.cortex.svc.cluster.local/prometheus/api/v1/alerts$urlsuffix";
             proxy_pass $backend;
           }
 
           # Cortex specific ruler routes (legacy routes which the Grafana 8 alerting feature uses)
-
-          location /rules {
+          location ~ ^/rules(?<urlsuffix>.*)$
             ${generalProxyConfig}
-
             resolver ${dnsResolver};
-            set $backend http://ruler.cortex.svc.cluster.local/api/v1/rules$is_args$query_string;
-            proxy_pass $backend;
-          }
-
-          location /rules/ {
-            ${generalProxyConfig}
-
-            resolver ${dnsResolver};
-            set $backend http://ruler.cortex.svc.cluster.local/api/v1/rules$is_args$query_string;
+            set $backend http://ruler.cortex.svc.cluster.local/api/v1/rules$urlsuffix;
             proxy_pass $backend;
           }
           `
--- a/packages/controller/src/resources/monitoring/tenants/grafanaDatasources.ts
+++ b/packages/controller/src/resources/monitoring/tenants/grafanaDatasources.ts
@@ -255,43 +255,39 @@ http {
               ${generalProxyConfig}
 
               resolver ${dnsResolver};
-              set $backend http://querier.loki.svc.cluster.local:1080/$request_uri;
+              set $backend http://querier.loki.svc.cluster.local:1080;
               proxy_pass $backend;
             }
 
             location /prometheus/api/v1/rules {
               ${generalProxyConfig}
 
-              # Check https://github.com/opstrace/opstrace/issues/896 for more details.
               resolver ${dnsResolver};
-              set $backend http://ruler.loki.svc.cluster.local:1080/prometheus/api/v1/rules$request_uri;
+              set $backend http://ruler.loki.svc.cluster.local:1080/prometheus/api/v1/rules$is_args$query_string;
               proxy_pass $backend;
             }
 
             location /prometheus/api/v1/rules/ {
               ${generalProxyConfig}
 
-              # Check https://github.com/opstrace/opstrace/issues/896 for more details.
               resolver ${dnsResolver};
-              set $backend http://ruler.loki.svc.cluster.local:1080/prometheus/api/v1/rules/$request_uri;
+              set $backend http://ruler.loki.svc.cluster.local:1080/prometheus/api/v1/rules$is_args$query_string;
               proxy_pass $backend;
             }
 
             location /prometheus/api/v1/alerts {
               ${generalProxyConfig}
 
-              # Check https://github.com/opstrace/opstrace/issues/896 for more details.
               resolver ${dnsResolver};
-              set $backend http://ruler.loki.svc.cluster.local:1080/prometheus/api/v1/alerts$request_uri;
+              set $backend http://ruler.loki.svc.cluster.local:1080/prometheus/api/v1/alerts$is_args$query_string;
               proxy_pass $backend;
             }
 
             location /prometheus/api/v1/alerts/ {
               ${generalProxyConfig}
 
-              # Check https://github.com/opstrace/opstrace/issues/896 for more details.
               resolver ${dnsResolver};
-              set $backend http://ruler.loki.svc.cluster.local:1080/prometheus/api/v1/alerts/$request_uri;
+              set $backend http://ruler.loki.svc.cluster.local:1080/prometheus/api/v1/alerts$is_args$query_string;
               proxy_pass $backend;
             }
 
@@ -300,18 +296,16 @@ http {
             location /loki/api/v1/rules {
               ${generalProxyConfig}
 
-              # Check https://github.com/opstrace/opstrace/issues/896 for more details.
               resolver ${dnsResolver};
-              set $backend http://ruler.loki.svc.cluster.local:1080/loki/api/v1/rules$request_uri;
+              set $backend http://ruler.loki.svc.cluster.local:1080/loki/api/v1/rules$is_args$query_string;
               proxy_pass $backend;
             }
 
             location /loki/api/v1/rules/ {
               ${generalProxyConfig}
 
-              # Check https://github.com/opstrace/opstrace/issues/896 for more details.
               resolver ${dnsResolver};
-              set $backend http://ruler.loki.svc.cluster.local:1080/loki/api/v1/rules/$request_uri;
+              set $backend http://ruler.loki.svc.cluster.local:1080/loki/api/v1/rules$is_args$query_string;
               proxy_pass $backend;
             }
 
@@ -319,9 +313,8 @@ http {
             location /api/prom/rules/ {
               ${generalProxyConfig}
 
-              # Check https://github.com/opstrace/opstrace/issues/896 for more details.
               resolver ${dnsResolver};
-              set $backend http://ruler.loki.svc.cluster.local:1080/loki/api/v1/rules/$request_uri;
+              set $backend http://ruler.loki.svc.cluster.local:1080/loki/api/v1/rules$is_args$query_string;
               proxy_pass $backend;
             }
             `
--- a/packages/controller/src/resources/monitoring/tenants/grafanaDatasources.ts
+++ b/packages/controller/src/resources/monitoring/tenants/grafanaDatasources.ts
@@ -283,13 +283,6 @@ http {
               proxy_pass $backend;
             }
 
-            location /prometheus/api/v1/alerts/ {
-              ${generalProxyConfig}
-
-              resolver ${dnsResolver};
-              set $backend http://ruler.loki.svc.cluster.local:1080/prometheus/api/v1/alerts$is_args$query_string;
-              proxy_pass $backend;
-            }
 
             # Loki specific ruler routes
 
@@ -297,15 +290,10 @@ http {
               ${generalProxyConfig}
 
               resolver ${dnsResolver};
-              set $backend http://ruler.loki.svc.cluster.local:1080/loki/api/v1/rules$is_args$query_string;
-              proxy_pass $backend;
-            }
-
-            location /loki/api/v1/rules/ {
-              ${generalProxyConfig}
-
-              resolver ${dnsResolver};
-              set $backend http://ruler.loki.svc.cluster.local:1080/loki/api/v1/rules$is_args$query_string;
+              # Do not change the request URL, this works for the following endpoints:
+              # /loki/api/v1/rules
+              # /loki/api/v1/rules/{namespace}
+              set $backend http://ruler.loki.svc.cluster.local:1080;
               proxy_pass $backend;
             }
 
--- a/packages/controller/src/resources/monitoring/tenants/grafanaDatasources.ts
+++ b/packages/controller/src/resources/monitoring/tenants/grafanaDatasources.ts
@@ -340,18 +340,16 @@ http {
           location / {
             ${generalProxyConfig}
 
-            # Check https://github.com/opstrace/opstrace/issues/896 for more details.
             resolver ${dnsResolver};
-            set $backend "http://query-frontend.cortex.svc.cluster.local:80$request_uri";
+            set $backend "http://query-frontend.cortex.svc.cluster.local:80";
             proxy_pass $backend;
           }
 
           location /api/v1/rules {
             ${generalProxyConfig}
 
-            # Check https://github.com/opstrace/opstrace/issues/896 for more details.
             resolver ${dnsResolver};
-            set $backend "http://ruler.cortex.svc.cluster.local/prometheus/api/v1/rules$request_uri";
+            set $backend "http://ruler.cortex.svc.cluster.local/prometheus/api/v1/rules$is_args$query_string";
             proxy_pass $backend;
 
           }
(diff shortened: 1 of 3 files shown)


test/test-remote/test_ui.ts
```diff
@@ -275,6 +278,73 @@ suite("test_ui_with_headless_browser", function () {
...
   });
 
+  test("test_grafana_datasource_proxy_loki_get_labels", async function () {
+    assert(COOKIES_AFTER_LOGIN);
+
+    const cookie_header_value = COOKIES_AFTER_LOGIN.map(
+      c => `${c.name}=${c.value}`
+    ).join("; ");
+
+    const url = `https://system.${CLUSTER_NAME}/grafana/api/datasources/proxy/2/loki/api/v1/label`;
```
jgehrcke, line 288: needs a code comment: this uses the fixed order of data sources in the Grafana data sources config:

- 0 is cortex
- 1 is loki
- 2 is cortex alertmanager

packages/controller/src/resources/monitoring/tenants/grafanaDatasources.ts
```diff
@@ -255,43 +255,39 @@ http {
...
               proxy_pass $backend;
             }
 
             location /prometheus/api/v1/rules {
               ${generalProxyConfig}
 
-              # Check https://github.com/opstrace/opstrace/issues/896 for more details.
               resolver ${dnsResolver};
-              set $backend http://ruler.loki.svc.cluster.local:1080/prometheus/api/v1/rules$request_uri;
+              set $backend http://ruler.loki.svc.cluster.local:1080/prometheus/api/v1/rules$is_args$query_string;
```
jgehrcke, line 266: again, the insight here is that what's written here _fully defines_ the URL that the nginx upstream is going to see. We want to drop the `/prometheus/api/v1/rules` part of the 'original' request URL and therefore should not use `$request_uri`.

jgehrcke:
It's freaking amazing how often something automatically imports something and I don't realize, only when CI throws an error:
```
[2021-06-18T13:46:35Z] test_ui.ts:61:1 - error TS6133: 'query' is declared but its value is never read.
[2021-06-18T13:46:35Z]
[2021-06-18T13:46:35Z] 61 import { query } from "express";
[2021-06-18T13:46:35Z]    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
[2021-06-18T13:46:35Z]
[2021-06-18T13:46:35Z] test_ui.ts:62:1 - error TS6133: 'http' is declared but its value is never read.
[2021-06-18T13:46:35Z]
[2021-06-18T13:46:35Z] 62 import { http } from "winston";
[2021-06-18T13:46:35Z]    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

jgehrcke:
in the last CI run I see that some deployments don't get ready so I probably broke the nginx config in one of the last few commits -- trying to find that.

jgehrcke:
last build was affected by https://github.com/opstrace/opstrace/issues/873#issuecomment-864084192 -- rebuild triggered

jgehrcke:
last build was affected by instability https://github.com/opstrace/opstrace/issues/709#issuecomment-864098630

packages/controller/src/resources/monitoring/tenants/grafanaDatasources.ts
```diff
@@ -346,64 +317,30 @@ http {
...
             proxy_pass $backend;
           }
 
-          location /api/v1/alerts/ {
+          location ~ ^/api/v1/alerts(?<urlsuffix>.*)$ {
             ${generalProxyConfig}
-
             resolver ${dnsResolver};
-            set $backend http://ruler.cortex.svc.cluster.local/prometheus/api/v1/alerts/$request_uri;
+            set $backend "http://ruler.cortex.svc.cluster.local/prometheus/api/v1/alerts$urlsuffix";
```
jgehrcke, line 335: note: review quotation marks here, should not be necessary; didn't notice before.

test/test-remote/test_ui.ts
```diff
@@ -131,7 +134,54 @@ async function performLoginFlow(br: Browser) {
...
+    },
+    https: { rejectUnauthorized: false }
+  } as GotOptions;
+}
+
+async function waitFor200Resp(
+  url: string,
+  httpopts: any
+): Promise<GotResponse<string>> {
+  const maxWaitSeconds = 2100;
```
jgehrcke, line 155: note: reduce this to maybe ~5 mins

test/test-remote/test_ui.ts
```diff
@@ -131,7 +134,54 @@ async function performLoginFlow(br: Browser) {
...
+    try {
+      resp = await httpcl(url, httpopts);
+    } catch (err) {
+      log.info(`request failed: ${err}`);
+    }
+
+    if (resp !== undefined) {
+      logHTTPResponse(resp);
+      if (resp.statusCode === 200) {
+        log.info("goot 200 response");
```
jgehrcke, line 174: goot goot goot

jgehrcke:
```
[2021-06-18T16:16:45Z] 2021-06-18T16:16:45.058Z info: GET HTTP request to https://system.prs-bk-4733-8bf-g.opstrace.io/grafana/api/datasources/proxy/2/loki/api/v1/label?start=1623672664679000000&end=1624036264679000000 failed. Attempt: 9. Error: ETIMEDOUT, Timeout awaiting 'request' for 30000ms. Retry in 15.0 s.
[2021-06-18T16:17:30Z] 2021-06-18T16:17:30.062Z info: GET HTTP request to https://system.prs-bk-4733-8bf-g.opstrace.io/grafana/api/datasources/proxy/2/loki/api/v1/label?start=1623672664679000000&end=1624036264679000000 failed. Attempt: 10. Error: ETIMEDOUT, Timeout awaiting 'request' for 30000ms. Retry in 15.0 s.
[2021-06-18T16:18:15Z] 2021-06-18T16:18:15.068Z info: request failed: TimeoutError: Timeout awaiting 'request' for 30000ms
```
No response returns within 30 seconds.

Interestingly, the data source splitter nginx is also waiting for that response from upstream. When the client goes away, it logs an access log line with the nginx-specific status code 499:

Screenshot from 2021-06-18 19-15-56

Note that, of course, from the data source splitter nginx point of view the HTTP client is Grafana.

The access log line
```
2021-06-18 19:11:10	
10.120.1.9 - - [18/Jun/2021:17:11:10 +0000] "GET /loki/api/v1/rules HTTP/1.1" 499 0 "-" "Grafana/8.0.3"
```
was me accessing the cluster via curl:
```
$ curl -v -L 'https://system.prs-bk-4733-8bf-g.opstrace.io/grafana/api/datasources/proxy/2/loki/api/v1/rules'   -H 'authority: system.jp-1623662956.opstrace.io'   -H 'pragma: no-cache'   -H 'cache-control: no-cache'   -H 'sec-ch-ua: " Not;A Brand";v="99", "Google Chrome";v="91", "Chromium";v="91"'   -H 'accept: application/json, text/plain, */*'   -H 'dnt: 1'   -H 'x-grafana-org-id: 1'   -H 'sec-ch-ua-mobile: ?0'   -H 'user-agent: Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/91.0.4472.77 Safari/537.36'   -H 'sec-fetch-site: same-origin'   -H 'sec-fetch-mode: cors'   -H 'sec-fetch-dest: empty'   -H 'referer: https://system.jp-1623662956.opstrace.io/grafana/explore?orgId=1&left=%5B%22now-1h%22,%22now%22,%22tenant-goodteam-metrics%22,%7B%7D%5D'   -H 'accept-language: en-US,en;q=0.9,de;q=0.8'   -H 'cookie: opstrace.sid=s%3AEH89Ok5iglYyecTz8WU0TkITB_4oar_p.yZoKCcjhHAtkNimqb1ETk%2Bb63MTVB6XO%2FRGGae%2B5wk0'   --compressed   --insecure
*   Trying 34.102.47.110:443...
* Connected to system.prs-bk-4733-8bf-g.opstrace.io (34.102.47.110) port 443 (#0)
* ALPN, offering h2
* ALPN, offering http/1.1
* successfully set certificate verify locations:
*   CAfile: /etc/pki/tls/certs/ca-bundle.crt
  CApath: none
* TLSv1.3 (OUT), TLS handshake, Client hello (1):
* TLSv1.3 (IN), TLS handshake, Server hello (2):
* TLSv1.3 (IN), TLS handshake, Encrypted Extensions (8):
* TLSv1.3 (IN), TLS handshake, Certificate (11):
* TLSv1.3 (IN), TLS handshake, CERT verify (15):
* TLSv1.3 (IN), TLS handshake, Finished (20):
* TLSv1.3 (OUT), TLS change cipher, Change cipher spec (1):
* TLSv1.3 (OUT), TLS handshake, Finished (20):
* SSL connection using TLSv1.3 / TLS_AES_256_GCM_SHA384
* ALPN, server accepted to use h2
* Server certificate:
*  subject: CN=prs-bk-4733-8bf-g.opstrace.io
*  start date: Jun 18 14:47:58 2021 GMT
*  expire date: Sep 16 14:47:57 2021 GMT
*  issuer: C=US; O=(STAGING) Let's Encrypt; CN=(STAGING) Artificial Apricot R3
*  SSL certificate verify result: unable to get local issuer certificate (20), continuing anyway.
* Using HTTP2, server supports multi-use
* Connection state changed (HTTP/2 confirmed)
* Copying HTTP/2 data in stream buffer to connection buffer after upgrade: len=0
* Using Stream ID: 1 (easy handle 0x559b5b2e6d90)
> GET /grafana/api/datasources/proxy/2/loki/api/v1/rules HTTP/2
> Host: system.prs-bk-4733-8bf-g.opstrace.io
> accept-encoding: deflate, gzip, br
> authority: system.jp-1623662956.opstrace.io
> pragma: no-cache
> cache-control: no-cache
> sec-ch-ua: " Not;A Brand";v="99", "Google Chrome";v="91", "Chromium";v="91"
> accept: application/json, text/plain, */*
> dnt: 1
> x-grafana-org-id: 1
> sec-ch-ua-mobile: ?0
> user-agent: Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/91.0.4472.77 Safari/537.36
> sec-fetch-site: same-origin
> sec-fetch-mode: cors
> sec-fetch-dest: empty
> referer: https://system.jp-1623662956.opstrace.io/grafana/explore?orgId=1&left=%5B%22now-1h%22,%22now%22,%22tenant-goodteam-metrics%22,%7B%7D%5D
> accept-language: en-US,en;q=0.9,de;q=0.8
> cookie: opstrace.sid=s%3AEH89Ok5iglYyecTz8WU0TkITB_4oar_p.yZoKCcjhHAtkNimqb1ETk%2Bb63MTVB6XO%2FRGGae%2B5wk0
> 
* TLSv1.3 (IN), TLS handshake, Newsession Ticket (4):
* TLSv1.3 (IN), TLS handshake, Newsession Ticket (4):
* old SSL session ID is stale, removing
* Connection state changed (MAX_CONCURRENT_STREAMS == 128)!
* ...
```

woof, debugging needs to continue from here.

test/test-remote/test_ui.ts
MatApple: 😂

jgehrcke:
Huh. On AWS; this test succeeded:

```
[2021-06-18T16:30:23Z] 2021-06-18T16:30:23.127Z [32minfo[39m: Waiting for a 200 OK response to be returned by https://system.prs-bk-4733-8bf-a.opstrace.io/grafana/api/datasources/proxy/2/loki/api/v1/label
[2021-06-18T16:30:23Z] 2021-06-18T16:30:23.610Z [32minfo[39m: HTTP resp to GET(https://system.prs-bk-4733-8bf-a.opstrace.io/grafana/api/datasources/proxy/2/loki/api/v1/label?start=1623673823125000000&end=1624037423125000000):
[2021-06-18T16:30:23Z]   status: 200
[2021-06-18T16:30:23Z]   body[:500]: {"status":"success","data":["__name__","k8s_container_name","k8s_host","k8s_namespace_name","k8s_pod_name"]}
[2021-06-18T16:30:23Z] 
[2021-06-18T16:30:23Z]   headers: {"date":"Fri, 18 Jun 2021 16:30:23 GMT","content-type":"application/json; charset=UTF-8","content-length":"109","connection":"close","x-content-type-options":"nosniff","x-xss-protection":"1; mode=block","strict-transport-security":"max-age=15724800; includeSubDomains","access-control-allow-origin":"https://prs-bk-4733-8bf-a.opstrace.io","access-control-allow-credentials":"true"}
[2021-06-18T16:30:23Z]   totalTime: 0.404 s
[2021-06-18T16:30:23Z]   dnsDone->TCPconnectDone: 0.002 s
[2021-06-18T16:30:23Z]   connectDone->reqSent 0 s
[2021-06-18T16:30:23Z]   reqSent->firstResponseByte: 0.294 s
[2021-06-18T16:30:23Z]   
[2021-06-18T16:30:23Z] 2021-06-18T16:30:23.610Z [32minfo[39m: goot 200 response
[2021-06-18T16:30:23Z] 2021-06-18T16:30:23.613Z [32minfo[39m: r.data: [
[2021-06-18T16:30:23Z]   '__name__',
[2021-06-18T16:30:23Z]   'k8s_container_name',
[2021-06-18T16:30:23Z]   'k8s_host',
[2021-06-18T16:30:23Z]   'k8s_namespace_name',
[2021-06-18T16:30:23Z]   'k8s_pod_name'
[2021-06-18T16:30:23Z] ]
[2021-06-18T16:30:23Z] 2021-06-18T16:30:23.613Z [32minfo[39m: found `k8s_container_name` label, success
[2021-06-18T16:30:23Z]   [32m  ✓[0m[90m test_grafana_datasource_proxy_loki_get_labels[0m[31m (490ms)[0m
```

This also succeeded:
```
[2021-06-18T16:30:23Z]   [32m  ✓[0m[90m test_grafana_datasource_proxy_cortex_get_rules_legacy[0m[31m (292ms)[0m
```

This test needs adjustment, no 200 OK expected:
```
a.opstrace.io/grafana/api/datasources/proxy/2/loki/api/v1/rules):
[2021-06-18T16:30:23Z]   status: 404
[2021-06-18T16:30:23Z]   body[:500]: no rule groups found
```
(but also works, yeah!)

This passed:
```
[2021-06-18T17:05:29Z]   [32m  ✓[0m[90m test_grafana_datasource_proxy_loki_get_alerts_legacy[0m[33m (40ms)[0m
```

So, on AWS the only failing test really was
```
[2021-06-18T17:16:00Z] [0m  1) test_ui_api
[2021-06-18T17:16:00Z]        test_grafana_datasource_proxy_loki_get_rules:
[2021-06-18T17:16:00Z] [0m[31m     Error: Expectation not fulfilled within 2100 s[0m[90m
```

What the heck happened in the GCP run...?

jgehrcke:
> What the heck happened in the GCP run...?

maybe https://github.com/grafana/loki/issues/3085 or something similar? (What the heck happened in the GCP run...? is s3-specific, but why should this problem not also exist for gcs?)

MatApple:
It's possible that the ID for the datasource is incorrect (it might not be consistent across runs) https://grafana.com/docs/grafana/latest/http_api/data_source/

MatApple:
Adding the test as a browser test would solve for that, because the Grafana UI just uses the name of the datasource instead of the ID.

jgehrcke:
> It's possible that the ID for the datasource is incorrect (it might not be consistent across installs) https://grafana.com/docs/grafana/latest/http_api/data_source/

The order is static; defined by the JSON array defining the data sources. The container logs (screenshot above) also show that the requests that after all timed out correctly passed through the `grafana-datasource-splitter-loki` nginx.

MatApple:
> The order is static; defined by the JSON array defining the data sources. 

As long as we're super careful not to change the order of the datasources array. Doing so has a nice side-effect in this case! Perhaps some comments in code will suffice

jgehrcke:
> Perhaps some comments in code will suffice

Oh yeah right, in terms of code comments the current patch is super weak. 

Good code comments in both places (`grafanaDatasources.ts` as well as in the testing code) are indeed absolutely necessary to save our future selves from pain :) Specifically:

- a warning in `grafanaDatasources.ts`, in the lines of: "changing the order might break tests that rely on the order"
- an explanation in the testing code, showing the translation (1: cortex, 2: loki, 3: ...; as defined by .... in `grafanaDatasources.ts`).

If this convention/reliance becomes a problem in the future I suppose there are good ways to iterate on the situation; I am sure there also is a dynamic lookup method to our avail. To identify the desired ID based on a name -- the UI after all also has to do this.

MatApple:
> naive question: can't we just replace data sources during the upgrade w/o changing their annotation/version?

We can, we just have to move to using the Grafana API instead of this provisioning method!

MatApple:
It's worth noting that the configuration used for nginx in this PR worked for all endpoints (I thoroughly tested that). https://github.com/opstrace/opstrace/pull/870

What was broken in the PR though, was the lack of nginx dynamic backend resolution, so when a backend failed (e.g. pod went away due to a restart), nginx didn't update the list of working backends.

@jgehrcke as long as this new nginx proxy configs result in the same proxy_pass locations as used in this linked PR, then the endpoints should work.

MatApple:
The previously working `proxy_pass` config can be found here https://github.com/opstrace/opstrace/blob/8f90daa42bc6343db8e521f212217f66fbd72a20/packages/controller/src/resources/monitoring/tenants/grafanaDatasources.ts

MatApple:
A quick check and I believe the updated routes look correct @jgehrcke, based on what the working `proxy_pass` routes were previously

MatApple:
After an 1h 40m, AWS has passed and GCP is still stuck on:
```
[2021-06-18T20:35:51Z] 2021-06-18T20:35:51.052Z info: GET HTTP request to https://system.prs-bk-4742-a58-g.opstrace.io/grafana/api/datasources/proxy/1/rules failed. Attempt: 6. Error: ETIMEDOUT, Timeout awaiting 'request' for 30000ms. Retry in 15.0 s.
```

jgehrcke:
> [...] the lack of nginx dynamic backend resolution, [...]  @sreis added a fix for this, but in the process we ended up proxying some endpoints to different places).

The fix you refer to here can be called "DNS caching workaround", and this is what I was referring to:

> If the goal is to rewrite parts of the URL and to use the DNS caching workaround one cannot rely on the proxy_pass URL rewriting behavior documented at http://nginx.org/en/docs/http/ngx_http_proxy_module.html#proxy_pass

To make it really clear: we want both in the reverse proxy in question:
- rewrite parts of the URL **and**
- the DNS caching workaround

As described, wanting _both_ at the same time requires special measures -- hope it's now clearer why the changes here are needed.

The GCP problem currently reproduces. Fun.

I did a bit more log digging in the affected cluster. This is how this looks from the Grafana proxy point of view:

> t=2021-06-18T20:17:29+0000 lvl=eror msg="Data proxy error" logger=data-proxy-log userId=2 orgId=1 uname=ci-test@opstrace.com path=/api/datasources/proxy/2/loki/api/v1/label remote_addr=192.168.0.3 referer="https://system.prs-bk-4742-a58-g.opstrace.io/grafana/explore?orgId=1&left=%5B%22now-1h%22,%22now%22,%22logs%22,%7B%7D%5D" error="http: proxy error: net/http: timeout awaiting response headers"

Reminder:

`Browser -> Grafana web app -> Grafana proxy -> data source splitter proxy -> loki`

Grafana proxy implements its own timeout criterion here which is when it logs `error="http: proxy error: net/http: timeout awaiting response headers"` and emits a 502 response, rightfully.

ingress nginx then logs the corresponding access log line:

```
192.168.0.3 - - [18/Jun/2021:20:17:29 +0000] "GET /grafana/api/datasources/proxy/2/loki/api/v1/label?from=1624025819030000000&to=1624047419030000000 HTTP/2.0" 502 0 "https://system.prs-bk-4742-a58-g.opstrace.io/grafana/explore?orgId=1&left=%5B%22now-1h%22,%22now%22,%22logs%22,%7B%7D%5D" "Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/91.0.4472.77 Safari/537.36" 89 30.027 [system-tenant-grafana-3000] [] 10.12.1.11:3000 0 30.020 502 82b7a2db143f7daca4c646888d7fd170
```

and, now the actual error, in the error log emitted by the `grafana-datasource-splitter-loki` nginx:

```
2021-06-18 22:17:29	
2021/06/18 20:17:29 [error] 21#21: *58 querier.loki.svc.cluster.local could not be resolved (110: Operation timed out), client: 10.12.1.11, server: _, request: "GET /loki/api/v1/label?from=1624025819030000000&to=1624047419030000000 HTTP/1.1", host: "grafana-datasource-splitter-loki.system-tenant.svc.cluster.local"
```

a DNS name resolution _timeout_. :fire: 

What is the relevant difference between AWS / EKS and GCP/GKE here?

MatApple:
Perhaps the value of dnsResolver in `resolver ${dnsResolver};` for GCP isn't correct.

jgehrcke:
> The kubelet agent running on each Node configures the Pod's etc/resolv.conf to use the kube-dns service's ClusterIP. An example of this configuration is shown below, in this example the IP address of the kube-dns service is 10.0.0.10 **(this IP address will be different in other clusters):**

(emphasis mine)

from https://cloud.google.com/kubernetes-engine/docs/concepts/service-discovery

jgehrcke:
@MatApple ok I am working on a quick patch, the last one for today certainly haha.

MatApple:
@jgehrcke I've seen this for the resolver too:
`resolver kube-dns.kube-system.svc.cluster.local valid=5s;`

From not the most relevant docs but does have some relevancy: https://www.nginx.com/blog/announcing-nginx-ingress-controller-for-kubernetes-release-1-4-0/

jgehrcke:
`resolver kube-dns.kube-system.svc.cluster.local valid=5s;` -- hmm. Seems legit enough to try this. SOmewhat nicer than reading and parsing the file.

jgehrcke:
@MatApple please take it from here! I also added the code comments we were musing about.

MatApple:
Tests passed on GCP 🎉 the new resolver seems to have done the trick! Thanks for sorting this out @jgehrcke!

MatApple:
I should also note that I logged into the CI cluster during the run and can confirm all the endpoints are doing what we expect.

MatApple:
All tests passed! Merging while waiting for BK to report the successful run to land this.

MatApple:
@jgehrcke it's probably worth attempting to set the new resolver for AWS too, next week!

MatApple:
Triggering a main build

jgehrcke:
> I should also note that I logged into the CI cluster during the run and can confirm all the endpoints are doing what we expect.

Thanks for this manual test!

PR #921, opened 2021-06-21
chore: don't use headless services in grafana datasource splitter proxy

sreis:
Instead of setting a DNS resolver in the grafana datasource proxy use services of type clusterIP. This way we avoid nginx caching issues when services restart.

On upgrade, the controller will fail to update the spec.clusterIP field because it's immutable. The proposed solution is to delete the old services before installing the new controller. To avoid deleting the services on every upgrade the CLI checks if the new `query-frontend-discovery` service exists or not. It it doesn't exist it means the Opstrace instance is running with the deprecated services.

### DIFF
--- a/packages/controller/src/resources/monitoring/tenants/grafanaDatasources.ts
+++ b/packages/controller/src/resources/monitoring/tenants/grafanaDatasources.ts
@@ -23,7 +23,7 @@ import {
 } from "@opstrace/kubernetes";
 import { State } from "../../../reducer";
 import { Tenant } from "@opstrace/tenants";
-import { getTenantNamespace, getControllerConfig } from "../../../helpers";
+import { getTenantNamespace } from "../../../helpers";
 import { KubeConfig } from "@kubernetes/client-node";
 import { DockerImages, getImagePullSecrets } from "@opstrace/controller-config";
 
@@ -58,33 +58,6 @@ export function GrafanaDatasourceResources(
     }
     return datasourcesToDelete;
   };
-
-  // Check https://github.com/opstrace/opstrace/issues/896 for more details.
-  // Using a variable in proxy_pass forces re-resolution of the DNS names
-  // because NGINX treats variables differently to static configuration. From
-  // the NGINX proxy_pass documentation
-  // http://nginx.org/en/docs/http/ngx_http_proxy_module.html#proxy_pass. For
-  // this to work we need to also set a resolver but EKS and GKE use different
-  // ClusterIPs for the pod DNS upstream resolver.
-  let dnsResolver = "";
-  const { target } = getControllerConfig(state);
-
-  if (target === "aws") {
-    dnsResolver = "10.100.0.10";
-  }
-
-  if (target === "gcp") {
-    // This IP address depends on the GKE cluster, see
-    // https://cloud.google.com/kubernetes-engine/docs/concepts/service-discovery
-    // We could parse /etc/resolv.conf but it seems like
-    // a _dynamic_ resolver should also work:
-    // kube-dns.kube-system.svc.cluster.local valid=5s;
-    // Refs:
-    //   https://www.nginx.com/blog/announcing-nginx-ingress-controller-for-kubernetes-release-1-4-0/
-    //   https://kubernetes.io/docs/tasks/administer-cluster/dns-debugging-resolution/
-    // Does/should this work for AWS, too?
-    dnsResolver = "kube-dns.kube-system.svc.cluster.local valid=5s";
-  }
 
   // Warning: the order in the `datasources` array defines the IDs assigned to
   // individual data sources -- these IDs are being used in certain tests,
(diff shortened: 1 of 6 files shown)


jgehrcke:
ignorant question: why are no changes needed for the loki information flow?

sreis:
> ignorant question: why are no changes needed for the loki information flow?

@jgehrcke  Because the Loki ruler and querier services are already of type clusterIP. (BTW, services of type ClusterIP but with spec.clusterIP=None are called [headless services](https://kubernetes.io/docs/concepts/services-networking/service/#headless-services)).

```
> kubectl -n loki get services
NAME                      TYPE        CLUSTER-IP      EXTERNAL-IP   PORT(S)              AGE
compactor                 ClusterIP   10.100.250.22   <none>        1080/TCP,9095/TCP    4h31m
distributor               ClusterIP   10.100.36.70    <none>        1080/TCP,9095/TCP    4h31m
ingester                  ClusterIP   10.100.182.68   <none>        1080/TCP,9095/TCP    4h31m
loki-gossip-ring          ClusterIP   None            <none>        7946/TCP             4h31m
memcached                 ClusterIP   None            <none>        11211/TCP,9150/TCP   4h31m
memcached-index-queries   ClusterIP   None            <none>        11211/TCP,9150/TCP   4h31m
querier                   ClusterIP   10.100.34.248   <none>        1080/TCP,9095/TCP    4h31m
ruler                     ClusterIP   10.100.15.130   <none>        1080/TCP,9095/TCP    4h19m
```

packages/controller/src/resources/monitoring/tenants/grafanaDatasources.ts
```diff
@@ -269,44 +242,52 @@ http {
...
+            }
+
+            location /loki/api/v1/rules/ {
+              ${generalProxyConfig}
+
+              proxy_pass http://ruler.loki.svc.cluster.local:1080/loki/api/v1/rules/;
             }
 
             # Loki legacy route for saving rules (legacy routes which the Grafana 8 alerting feature uses)
-            location ~ ^/api/prom/rules(?<urlsuffix>.*)$ {
```
sreis, line 305: Using a variable in the proxy_pass requires us to set the DNS resolver otherwise nginx returns a 502 and logs the error no resolver defined to resolve. This is why I reverted back to setting the location explicitly here and in the other instances.

jgehrcke:
Huh, okay, we backed out of the 'nginx DNS re-resolution workaround'. Sounds great! Overall that was a really complicated way involving a pretty big number of people to end up with a solution that we maybe should have picked from the start :D

I trust that you made the right choices here. But I would like to understand a little better. Going through this step by step.

Now, we use nginx's default `resolver` behavior again  (with `resolver` being _not_ set). I was wondering what that default behavior really is -- it's not well specified in http://nginx.org/en/docs/http/ngx_http_core_module.html#resolver. After reading some resources on the web, it's really the following: by default, nginx resolves DNS names in the config file only _once_ during startup, and then only ever again during config reload if that's ever triggered. I want to quote

> If your config file has static DNS names (not generated), and you do not care about track IP changes without nginx reload, you don't need nginx's resolver. In this case all DNS names will be resolved on startup.

from https://stackoverflow.com/a/40331256/145400.

So. Now that we use services of type cluster IP I think the key question is: can these IP addresses ever change? Because in that case, only an nginx restart would pick up the change.

https://kubernetes.io/docs/concepts/services-networking/service/ says about ClusterIP:

> Exposes the Service on a cluster-internal IP. Choosing this value makes the Service only reachable from within the cluster. This is the default ServiceType. 

Says nothing about the guarantees implied by that IP.

https://stackoverflow.com/a/39656020/145400 says

> The short answer is "Yes, the service IP can change" [...]
> it can get re-resolved to a new IP (which will probably never happen unless you explicitly delete and recreate service) 

Are we happy with that?

sreis:
> So. Now that we use services of type `ClusterIP` I think the key question is: can these IP addresses ever change? Because in that case, only an nginx restart would pick up the change.
> 
> https://kubernetes.io/docs/concepts/services-networking/service/ says about ClusterIP:
> 
> > Exposes the Service on a cluster-internal IP. Choosing this value makes the Service only reachable from within the cluster. This is the default ServiceType.
> 
> Says nothing about the guarantees implied by that IP.
> 
> https://stackoverflow.com/a/39656020/145400 says
> 
> > The short answer is "Yes, the service IP can change" [...]
> > it can get re-resolved to a new IP (which will probably never happen unless you explicitly delete and recreate service)
> 
> Are we happy with that?

@jgehrcke As you point out, the service IP changes if you delete and create the Kubernetes service. Let me ask you this, what is more likely to occur, a pod restart or recreating the Kubernetes service?

sreis:
As discussed with @jgehrcke in slack, for the lifetime of the Opstrace instance, Kubernetes services are not expected to change, at least not via normal operations such as during an upgrade.

sreis:
Triggering the upgrade test manually for this PR to check it doesn't introduce any new bugs.

sreis:
Upgrade test failed. I'll set this PR to draft and come back to it when upgrades are green.

MatApple:
> Huh, okay, we backed out of the 'nginx DNS re-resolution workaround'. Sounds great! Overall that was a really complicated way involving a pretty big number of people to end up with a solution that we maybe should have picked from the start :D That is how it is!

@jgehrcke just want to point out that this is now exactly what we had at the start (my original PR), but with a bug fix (making sure all relevant services aren't headless). Nonetheless, finding problems can be a journey and can/should put the original solution into question.

MatApple:
@sreis is this still something we want to land?

sreis:
Things have stabilized since I worked on this. We can revisit again if necessary.

Issue #905, opened 2021-06-17
check cortex distributor logs: Error: Expectation not fulfilled within 30 s


sreis:
**Describe the bug**

This test is failing consistently after upgrading to latest main: 

```
[2021-06-17T05:36:37Z]   1) system logs test suite
[2021-06-17T05:36:37Z]        check cortex distributor logs:
[2021-06-17T05:36:37Z]      Error: Expectation not fulfilled within 30 s
[2021-06-17T05:36:37Z]       at Object.waitForLokiQueryResult (testutils/logs.ts:275:9)
[2021-06-17T05:36:37Z]       at query (test_systemlogs.ts:52:20)
[2021-06-17T05:36:37Z]       at Context.<anonymous> (test_systemlogs.ts:72:5)
```

sreis:
We bumped [Loki](https://github.com/opstrace/opstrace/commit/a8ae291b471b37c2a1e733c0cb9e7ea5786958ff) and that means the queriers restart after the upgrade which [triggers this closed but not resolved issue](https://github.com/grafana/loki/issues/3085).

sreis:
This is still bitting the upgrade CI pipeline.

sreis:
Ran another test locally this time after the upgrade the only pod logs in the cortex namespace were for the alertmanager and store-gateway. (Thought I had a screenshot of the log ingestion rate in this period but I don't). This is the log browser output:

![Screenshot from 2021-06-22 12-03-56](https://user-images.githubusercontent.com/495294/122922119-d177cb00-d352-11eb-9113-22aed349c0ae.png)

Used kubectl to check fluentd pod logs and didn't find any errors. Nevertheless, I did a rollout restart of the fluentd daemonset and it's now ingesting logs.

![Screenshot from 2021-06-22 12-12-16](https://user-images.githubusercontent.com/495294/122922465-2ddaea80-d353-11eb-90de-cb50c8449b19.png)

Log browser output.

![Screenshot from 2021-06-22 12-14-00](https://user-images.githubusercontent.com/495294/122922631-5a8f0200-d353-11eb-8388-e3ccd25cf647.png)

sreis:
Was able to repro by restarting the Loki ingesters via `kubectl rollout restart`. This is how it looks right now:

![Screenshot from 2021-06-22 13-16-34](https://user-images.githubusercontent.com/495294/122931154-25d37880-d35c-11eb-851b-d7752eded659.png)

[...]

PR #925, opened 2021-06-22
controller: revert loki bump

sreis:
Until we root cause https://github.com/opstrace/opstrace/issues/905 and fix the failing upgrades test let's revert the Loki bump.

I'll run the upgrades test pipeline for this PR when the cli build artifacts are published.


MatApple:
Despite the reporting here, all CI tests did actually pass in the latest run. One of the tests reported here is cached from the previous run that had failed due to a flake.

Screen Shot 2021-06-22 at 1 17 55 PM

PR #933, opened 2021-06-24
ui: preserve route when switching tenants

### LINKED ISSUE #924: feat: maintain the same page when switching tenants in the UI

Currently, when you select a new tenant from the UI you are automatically redirected to `/tenant/<name>/getting-started`, taking you away from whatever other page you were on.  In the case of cycling through many tenants to see various types of information (tenant overview metrics, integrations, etc.) this default behavior becomes tedious.

Instead, remain on whichever page you are viewing after switching tenants.

MoSattler:
closes #924

superset of #948

# Have you...

### DIFF
--- a/packages/app/src/client/views/tenants/TenantPicker.tsx
+++ b/packages/app/src/client/views/tenants/TenantPicker.tsx
@@ -26,6 +26,7 @@ import { Tenant } from "state/tenant/types";
 import { selectedTenantChanged } from "state/global/actions";
 
 import { PickerOption, usePickerService } from "client/services/Picker";
+import { useLastSelectedTenant } from "state/tenant/hooks/useTenant";
 
 type TenantPickerOption = Pick<PickerOption, "id" | "text"> & { data: Tenant };
 
@@ -43,6 +44,7 @@ const TenantPicker = () => {
   const dispatch = useDispatch();
   const history = useHistory();
   const tenants = useTenantList();
+  const lastSelectedTenant = useLastSelectedTenant();
 
   const { activatePickerWithText } = usePickerService(
     {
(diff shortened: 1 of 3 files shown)


packages/app/src/client/views/tenants/TenantPicker.tsx
```diff
@@ -43,17 +44,27 @@ const TenantPicker = () => {
...
         dispatch(selectedTenantChanged({ tenant: option.data }));
-        history.push(`/tenant/${option.id}`); // todo: would be nice to keep the current path and simply change the tenant rather than send user back to the home for the tenant
+        const { pathname } = history.location;
+        const newRoute = currentTenant
+          // if we're on a page with a tenant selected, stay on page and just replace the tenant
+          ? pathname.replace(
+              `/tenant/${currentTenant.name}`,
+              `/tenant/${option.id}`
+            )
+          // if no tenant is currently selected, reroute to tenant root route
```
MoSattler, line 62: this is currently the case in routes for cluster admin

MatApple: perhaps we can memoize the last value of `currentTenant.name` where the routes are declared and then we update that "root route" as referenced in the comment, to route to `/tenant/<memoized_tenant>` instead of `/tenant/system`?

MatApple: and only fallback to "system" tenant if the value of the memoized tenant is nullish

MoSattler: not sure if memoization is the right tool for this job. React even goes as far as to discourage using `useMemo` for these cases.

> You may rely on useMemo as a performance optimization, not as a semantic guarantee. In the future, React may choose to “forget” some previously memoized values and recalculate them on next render, e.g. to free memory for offscreen components. Write your code so that it still works without useMemo — and then add it to optimize performance. - [React Docs](https://reactjs.org/docs/hooks-reference.html#usememo)

For now I just added a new hook for this use case (that includes tests) [here](https://github.com/opstrace/opstrace/pull/933/commits/d3ef3d2231df2e36bc563aec6da9409c0f4316ba).

MatApple (approved):
LGTM!

PR #949, opened 2021-06-28
ui: enabled playwrite vids/pics on failure and get buildkit to save them

terrcin:


terrcin:
Wahoo! 
image

terrcin:
Playwright failures are now being stored in a mounted volume and copied to the buildkite artifacts folder.

I noted that the `test-remote` tests were have a dedicated `/test-remote-artifacts` folder setup but it was not being used so I updated the make and teardown scripts to use it. Expected artifacts have come though.

Interestingly, this branch is a branch of this https://github.com/opstrace/opstrace/pull/912 which failed again. Didn't see that failure on this PRs latest build - only the expected deliberate failure. If this next build pases merging this will include that other PR.

PR #952, opened 2021-06-28
looker: README, build changes, make more independent

jgehrcke:


jgehrcke:
got a little deep here with the changes; this really needs to go green and land.

Issue #920, opened 2021-06-21
tests/authentication: after auth0 authentication user should see homepage failed


sreis:
**Describe the bug**

```
[2021-06-21T13:22:42Z]   1) tests/authentication.spec.ts:25:3 › [WebKit] after auth0 authentication user should see homepage 
[2021-06-21T13:22:42Z]
[2021-06-21T13:22:42Z]     Timeout of 30000ms exceeded.
```

https://buildkite.com/opstrace/scheduled-main-builds/builds/2525#93045184-4e3b-4ae2-894f-9e7dcbc50e1d/9297-9321

sreis:
https://buildkite.com/opstrace/scheduled-main-builds/builds/2526#22de88c8-9a2e-4a8b-9c9e-23c7ba96f776/9244-9268

```
[2021-06-21T16:59:39Z]   1) tests/authentication.spec.ts:25:3 › [Chromium] after auth0 authentication user should see homepage 
[2021-06-21T16:59:39Z]
[2021-06-21T16:59:39Z]     Timeout of 30000ms exceeded.
```

jgehrcke:
```
[2021-06-25T01:03:30Z]   1 failed
[2021-06-25T01:03:30Z]     tests/authentication.spec.ts:25:3 › [Chromium] after auth0 authentication user should see homepage 
[2021-06-25T01:03:30Z]   3 passed (34s)
```
https://buildkite.com/opstrace/scheduled-main-builds/builds/2546#592f5c8f-b319-42dd-b832-422767ac78e8/9106-9160

jgehrcke:
https://buildkite.com/opstrace/prs/builds/4816#e8d0d0f1-13c3-4a60-8c79-b2a43610a913/3354-3385

jgehrcke:
```
[2021-06-28T11:11:50Z]   1) tests/authentication.spec.ts:25:3 › [WebKit] after auth0 authentication user should see homepage 
[2021-06-28T11:11:50Z]
[2021-06-28T11:11:50Z]     Timeout of 30000ms exceeded.
```
https://buildkite.com/opstrace/prs/builds/4833#67ddf005-e806-470b-8372-0e37044d8bd5/9183-9207

sreis:
```
[2021-06-29T01:42:49Z]   1) tests/authentication.spec.ts:25:3 › [Chromium] after auth0 authentication user should see homepage 
[2021-06-29T01:42:49Z]
[...]

PR #962, opened 2021-06-30
controller: switch pull policy from Always to IfNotPresent to reduce requests

triclambert:
As far as I can tell, there is no real reason to use an image pull policy of `Always` for some of our pods; this just adds extra requests to the registry which can trigger rate limits.  Instead, this PR switches the policy to `IfNotPresent`, similar to most of the other pods.

### DIFF
--- a/packages/controller-config/src/resources/controller.ts
+++ b/packages/controller-config/src/resources/controller.ts
@@ -106,7 +106,7 @@ export function ControllerResources({
                 {
                   name: "opstrace-controller",
                   image: `${controllerImage}`,
-                  imagePullPolicy: "Always",
+                  imagePullPolicy: "IfNotPresent",
                   command: ["node", "--prof", "./cmd.js"],
                   args: controllerCmdlineArgs,
                   resources: {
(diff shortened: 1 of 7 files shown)


jgehrcke:
I wanted to really understand the impact of this and make up my mind for what we want.

Went into reference docs:

> by **default** the pull policy of all containers in that pod will be set to IfNotPresent if it is not explicitly specified. This policy causes the kubelet to skip pulling an image if it already exists.

(https://kubernetes.io/docs/concepts/containers/images/)

That is, `IfNotPresent` is the default.

From a behavioral point of view, `IfNotPresent` might have undesired side effects only when the image reference does not refer to a specific version (but to a moving target instead), and when we explicitly desire the moving target to be followed across container restarts.

I believe in our case we actually want to use image references that do _not_ point to a moving target. Also, we even might consider it a bug when the actual image being varies across container restarts (I think this really applies to all components comprising an Opstrace instance).

I found it interesting to understand that with `Always`, the bulk of the image is _not_ newly fetched when the correct image (identified by digest/hash) is present locally:

> imagePullPolicy: Always: every time the kubelet launches a container, the kubelet queries the container image registry to resolve the name to an image digest. If the kubelet has a container image with that exact digest cached locally, the kubelet uses its cached image; otherwise, the kubelet downloads (pulls) the image with the resolved digest, and uses that image to launch the container.

(https://kubernetes.io/docs/concepts/configuration/overview/#container-images)

But yes, indeed the pretty lightweight request to translate the container image name to a specific version contributes to rate limiting considerations.

Only HEAD requests to Docker Hub do not contribute to rate limit considerations:

https://www.docker.com/pricing/resource-consumption-updates
https://docs.docker.com/docker-hub/download-rate-limit/

But a HEAD request cannot be used to translate an image tag into a specific digest/hash.

jgehrcke (approved):
:rocket: 

And let's also apply this for all other images in the system. As argued in my larger comment in the main thread above.

jgehrcke:
`main-test-gcp` failed with a known instability: https://github.com/opstrace/opstrace/issues/920#issuecomment-871228945

merging

triclambert:
> And let's also apply this for all other images in the system. As argued in my larger comment in the main thread above.

I don't think there are any others using Always, but maybe I'm missing some?

jgehrcke:
> I don't think there are any others using Always, but maybe I'm missing some?

If that is you saying that you had a look with e.g. `grep` and didn't find any more: :+1: -- all we need for now.

triclambert:
Yup, I used VScode followed by a sanity check:

```
> grep -Ri '"Always"' packages/**/*.ts | wc -l
       0
```

PR #963, opened 2021-06-30
ui: Create Tenant test

terrcin:


packages/app/src/client/services/Picker/PickerService.tsx
```diff
@@ -154,6 +155,10 @@ function PickerService({ children }: { children: React.ReactNode }) {
       </pickerContext.Provider>
       <Dialog
         data-testid="dialog"
+        data-test={
+          "pickerService/dialog" +
+          (activePicker?.dataTest ? "/" + activePicker.dataTest : "")
+        }
```
MoSattler, line 161: In my experience relying too much on data attributes in E2E tests can cause problems later on. Have you considered to rather rely on semantics to target the DOM? E.g. role/labels/text?
 
Not sure if playwright makes that easy, but it would be the equivalent of testing-library's `getByRole` et al.

MoSattler: Just found something about this in the [playwright's best practices](https://playwright.dev/docs/selectors/#prioritize-user-facing-attributes).

test/browser/tests/tenant.spec.ts
```diff
@@ -0,0 +1,56 @@
...
+  test.beforeEach(logUserIn);
+
+  test.beforeEach(async ({ page }) => {
+    await page.hover("[data-test='sidebar/clusterAdmin/Tenants']");
+    await page.click("[data-test='sidebar/clusterAdmin/Tenants']");
+    expect(await page.isVisible("[data-test='tenant/list']")).toBeTruthy();
+  });
+
+  test("user can create a new Tenant", async ({ page, cluster }) => {
+    const newTenant = `treetops-${Math.floor(Math.random() * 10000)}`;
```
jgehrcke, line 32: That is the name of the tenant, right? In this case I always like to encode that idea in the variable name, e.g. `tenantName`.

Note that you choose a `dash` here in the name of the tenant. 

We still need to add tenant name validation to the API, see https://github.com/opstrace/opstrace/issues/710#issuecomment-869479025 -- would still greatly appreciate help.

When we do this, we will find that (as of today) a dash should not be allowed: https://github.com/opstrace/opstrace/issues/957

Looking forward to seeing these tests here then also _cover_ the API-based tenant name validation.

jgehrcke: (just ideas that ran through my head while skimming over this -- feel free to ignore for now, whatever makes sense to you)

packages/app/src/client/services/Picker/PickerService.tsx
terrcin: As a general rule of thumb, if a button moves slightly, or content/text changes these browser tests shouldn't fail or need to be changed. Testing against labels/text etc... can be a huge maintenance nightmare during refactors. I've found putting explicit data attributes in for the testing to use to be the way to go - as described in your link as "Define explicit contract".

test/browser/tests/tenant.spec.ts
terrcin: that's a shame about dashing not being allowed, I'll just change to an underscore. Regarding the var name we use `tenant` to mean the name sometimes and sometimes not. Given the ID and the Name are the same thing it's all a bit confusing. :shrug:

terrcin:
I hit rebuild on CI as there were timeout on both GCP and AWS for test-browser and test-remote-ui.

PR #970, opened 2021-06-30
ui(ring-health): improve error handling

MoSattler:
Improve error handling of ring-health errors, as outlined [here](https://github.com/opstrace/opstrace/pull/900#issuecomment-868528836).

# Have you...

### DIFF
--- a/packages/app/src/server/routes/api/cortex/index.ts
+++ b/packages/app/src/server/routes/api/cortex/index.ts
@@ -73,18 +73,20 @@ cortexProxy.on("proxyRes", (proxyRes, req, res) => {
         /* setting the status code as 412, as usually HTML error responses
          * of cortex are due to misconfigurations
          */
-        res.writeHead(412, {
-          "content-type": "text/plain"
+        res.writeHead(500, {
+          "content-type": "application/json"
         });
-        res.end(errorMessage);
+        res.end(JSON.stringify({
+          errorMessage,
+        }));
       } catch (e) {
         res.writeHead(500, {
-          "content-type": "text/plain"
+          "content-type": "application/json"
         });
-        res.end(
-          "There has been an error parsing a cortex HTML response. Response:\n\n" +
-            body
-        );
+        res.end(JSON.stringify({
+          errorMessage: "There has been an error parsing a cortex HTML response.",
+          data: body.slice(0, 500)
+        }));
       }
     } else {
       /* if response is not HTML, the chunks have already been written
@@ -107,8 +109,8 @@ const proxyTo = (target: string) => (
       target,
       headers: req.params.tenant
         ? {
-            "X-Scope-OrgID": req.params.tenant
-          }
+          "X-Scope-OrgID": req.params.tenant
+        }
         : {}
     },
     (err: Error) => {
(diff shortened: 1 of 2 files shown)


packages/app/src/server/routes/api/cortex/index.ts
```diff
@@ -62,16 +62,30 @@ cortexProxy.on("proxyRes", (proxyRes, req, res) => {
...
+        const errorMessage = parsedHTML.querySelector("html body p")?.text;
+
+        if (!errorMessage) {
+          throw new Error("Could not extract error message");
+        }
+
+        /* setting the status code as 412, as usually HTML error responses
+         * of cortex are due to misconfigurations
+         */
+        res.writeHead(412, {
```
jgehrcke, line 76: btw, I know this was already in your previous patch but I didn't really notice and would like to comment on that now: all 4xx HTTP responses are supposed to mean that something is wrong with the _request_. The 412 response has a very specific meaning with respect to the request headers. So, semantically, it's a bit of an odd choice to communicate this problem in this kind of response.

Things like that are fine to be wrapped in a 500 response. Then of course one should find a way to allow the client to distinguish a generic "unexpected error" in the backend, from a specific (and potentially expected) error in the backend. In APIs this is often done by putting error detail into a JSON document into the response body. One can then do the following on the client side:
- get a 500 response. inspect the body. is this JSON?
    - yes: 
        - is this JSON of expected format containing err information?
           - yes: the backend communicated an expected, specific error to us -- we know exactly what that is and can handle it in a specific way

If any of the answers is "no": something unexpected happened in the backend and we handle that differently

jgehrcke: (just food for thought, this is your playground here :))

jgehrcke: here is an example in our own code base where we use this kind of appraoch to distinguish 404 responses from 404 responses via said method:
https://github.com/opstrace/opstrace/blob/9d683c95f934788cd38860d4df148310f0489cbf/lib/dns/src/opstrace.ts#L345

(in this case not via a specific JSON structure, but via presence of a specific error code which serves as a contract/convention)

packages/app/src/server/routes/api/cortex/index.ts
```diff
@@ -62,16 +62,30 @@ cortexProxy.on("proxyRes", (proxyRes, req, res) => {
...
+          "content-type": "text/plain"
+        });
+        res.end(errorMessage);
+      } catch (e) {
+        res.writeHead(500, {
+          "content-type": "text/plain"
+        });
+        res.end(
+          "There has been an error parsing a cortex HTML response. Response:\n\n" +
+            body
```
jgehrcke, line 86: Also food for thought for you to learn from the mess we've experienced in the past: so -- this here is _specifically_ for handling an _unexpected_ HTTP response. In which ways can an HTTP response be unexpected? In many ways. In all of them. And this piece of code _can_ handle all of them, it's actually pretty easy. There's just one thing to keep in mind: sometimes, an unexpected HTTP response may contain a _wild_ response body full of gibberish. This can be 1 byte or a gigabyte. In those cases it makes sense to emit only first N bytes.

MoSattler: makes sense - though I shortened it to 500 characters instead of 500 bytes.

packages/app/src/server/routes/api/cortex/index.ts
MoSattler: > In APIs this is often done by putting error detail into a JSON document into the response body.

Do we already have a standardised JSON format for our "internal" error transportation via HTTP? If we go this route, it might make sense to keep them consistent across the app.

MoSattler: I just added one for now - if that works great, if not, I am happy to change it.

jgehrcke (approved):
:rocket:

PR #971, opened 2021-06-30
ci: fix target rename

sreis:
Target was renamed in #952 but calls weren't updated.


ci/cd/run.sh
```diff
@@ -6,7 +6,7 @@ make rebuild-ci-container-image
 # testrunner run tsc which requires buildinfo package to be set
 make set-build-info-constants
 make rebuild-testrunner-container-images
-make rebuild-looker-container-images
+make rebuild-looker-container-image
```
jgehrcke, line 9: Wow, I keep bugging you changes in the upgrade test pipeline. Sorry, I should do more grepping @sreis

PR #975, opened 2021-07-01
go: log as debug message

sreis:
Default log level is info and in busy instances, handling hundreds of requests per second, this message spams the logs.

Found this tracking down why systemlog fluentd CPU usage was higher than expected;

screenshot

cortex-api pod logs:

Screenshot from 2021-07-01 14-36-05


jgehrcke:
Yes, thanks @sreis :)

sreis:
> Is this the collection cluster?

No, it's a customer cluster.

Issue #966, opened 2021-06-30
cortex: Assign default limits for inflight push requests


nickbp:
The cortex ingester and distributor pods each have message queues. By default these queues have no limit to their size, so if the ingesters get backed up then they (or the distributors) can end up OOMing, making the cortex cluster unhealthy. This state can potentially last for several minutes as the ingesters need to process/recover their WAL before they will accept traffic again.

However, to avoid this problem, we assign limits to the size of these queues, where the ingesters/distributors will reject messages if they're too backed up, reducing the likelihood of an OOM.

I think these values are reasonable for now:
- **ingester**: `1000`. This may be scaled proportional to the amount of memory on the host machine. However, I've seen in practice that when the ingester starts to develop a backlog because it's falling behind on writes under load, the backlog can go from zero to tens of thousands very quickly - and then drop away just as quickly once the hiccup has passed. As a result, if we set this to a higher value, e.g. `10000`, it would only give us maybe a few more seconds before that limit would be hit as well. So there isn't much harm in just staying at `1000`.
- **distributor**: `50`. This may be scaled proportional to the amount of memory on the host machine. In practice the distributor should only develop a backlog if the ingesters are already very sad. I've found in practice that the distributor backlog normally stays in the single digits. Additionally the distributor queue contains full POST requests from clients, so each request could be in the tens of megabytes territory, therefore this setting should stay pretty low to avoid excessive memory use from queued requests.

In terms of implementation, the ingester and distributor limits must be assigned in different places:

For the **ingester**, the limit could either be set in the runtime config or in the standard cortex config. However, if the runtime config facility is enabled (as it is in Opstrace), the ingester will ignore the standard cortex config for this value and only look at the runtime config value. **A catch here is that we'd want to avoid conflicts with any cortex runtime configuration provided by the user via the UI - I'm not sure what the best solution is here.** See also [test-infra/kube-overrides.yaml](https://github.com/opstrace/test-infra/blob/main/opstrace-kube-overrides.yaml#L66):
```
# in runtime config (`cortex/cortex-runtime-config` ConfigMap)
ingester_limits:
  max_inflight_push_requests: 1000
```

For the **distributor**, the limit is not assignable in the runtime config and therefore we can just set it in the [`distributor` section of the regular cortex config](https://github.com/opstrace/opstrace/blob/20694ef/packages/controller/src/resources/cortex/index.ts#L235):
```
# in regular cortex config managed by controller
distributor:
  instance_limits:
[...]

PR #986, opened 2021-07-05
controller: Configure inflight push request limits

nickbp:
By default, the ingester and distributor have an unlimited queue for pending push requests. This can lead to OOMing pods under heavy load (ingesters: lots of new series, distributors: large payloads with lots of labels).

This is a particular issue with ingesters, which lose in-memory state and take a while to recover, during which time cortex as a whole can become degraded.

These limits should reduce the likelihood of OOMing regardless of load. The comments give some justification around the chosen values, for more background see #966.

# Have you...


nickbp:
(rebase to pick up recent CI fixes)

PR #994, opened 2021-07-06
ui: browser tests - tenant name validation

terrcin:
testing to make sure user can't create tenant that have invalid chars in their names: `-`, `_`, `:`, ` `, `.`


terrcin:
thanks @jgehrcke. have updated the regex and tests.

PR #995, opened 2021-07-06
looker: Shift label randomness into invocation_id, allow configuration 

nickbp:
Randomness is provided via `invocation_id`, and then uniqueness across cycles and concurrent_streams is added on top via the indexes. End result:
- Metrics:
    - metricName: `<invocation_id>-<cycle#>`
    - uniqueName: `<invocation_id>-<cycle#>-<stream#>`
- Logs:
    - uniqueName: `<invocation_id>-<cycle#>-<stream#>`
    
With all the randomness moved into `invocation_id` (and with it exposed in the args), it can be overridden manually, allowing e.g. series counts to remain stable across looker runs - or the caller can guarantee unique values by using a custom randomness.

---

ALSO:

Add args for configuring range of metric series start times: Allow using ranges other than default 20-30 minutes.

### DIFF
--- a/test/test-remote/looker/src/args.ts
+++ b/test/test-remote/looker/src/args.ts
@@ -400,7 +400,7 @@ export function parseCmdlineArgs(): void {
     }
   }
 
-  if (CFG.max_concurrent_writes === 0) { // TODO this is always zero - remove from CFG?
+  if (CFG.max_concurrent_writes === 0) {
     CFG.max_concurrent_writes = CFG.n_concurrent_streams;
   }
 
--- a/test/test-remote/looker/src/args.ts
+++ b/test/test-remote/looker/src/args.ts
@@ -215,8 +215,9 @@ export function parseCmdlineArgs(): void {
   parser.add_argument("--change-streams-every-n-cycles", {
     help:
       "Use the same log/metric stream for N cycles, then create a new set of " +
-      "streams (unique label sets). Default: new streams are created with every " +
-      "write/read cycle. For log streams, when a new stream is initialized it " +
+      "streams (unique label sets), or <=0 to reuse streams for the process lifetime. " +
+      "Default: new streams are created with every write/read cycle (1). " +
+      "For log streams, when a new stream is initialized it " +
       "re-uses the same synthetic start time as set before (program invocation time " +
       "or log_start_time). Metric streams are always guided by wall time.",
     type: "int",
@@ -304,6 +305,15 @@ export function parseCmdlineArgs(): void {
     default: ""
   });
 
+  parser.add_argument("--invocation-id", {
+    help:
+      "Name of this looker instance to be included in labels. " +
+      "By default a random looker-<X> string is used. " +
+      "If read validation is enabled then care should be taken to ensure uniqueness at the server.",
+    type: "str",
+    default: ""
+  });
+
   CFG = parser.parse_args();
 
   setLogger(
@@ -312,9 +322,13 @@ export function parseCmdlineArgs(): void {
     })
   );
 
-  // const uniqueInvocationId = `looker-${START_TIME_EPOCH}-${rndstring(6)}`;
-  const uniqueInvocationId = `looker-${rndstring(6)}`;
-  CFG.invocation_id = uniqueInvocationId;
+  if (CFG.invocation_id === "") {
+    // common case: the id was not specified manually, so generate one
+    // ensure it has a good amount of randomness to ensure reads are unique (if enabled)
+    // const uniqueInvocationId = `looker-${START_TIME_EPOCH}-${rndstring(10)}`;
+    const uniqueInvocationId = `looker-${rndstring(10)}`;
+    CFG.invocation_id = uniqueInvocationId;
+  }
 
   if (CFG.log_start_time) {
     // validate input, let this blow up for now if input is invalid
--- a/test/test-remote/looker/src/args.ts
+++ b/test/test-remote/looker/src/args.ts
@@ -48,6 +48,8 @@ interface CfgInterface {
   stream_write_n_seconds_jitter: number;
   fetch_n_entries_per_query: number;
   metrics_mode: boolean;
+  metrics_past_start_range_min_seconds: number;
+  metrics_past_start_range_max_seconds: number;
   metrics_time_increment_ms: number;
   bearer_token_file: string;
   retry_post_deadline_seconds: number;
@@ -139,6 +141,22 @@ export function parseCmdlineArgs(): void {
       "log stream (between log entry timestamps) (ignored in metrics mode)",
     type: "int",
     default: 1
+  });
+
+  parser.add_argument("--metrics-past-start-range-max-seconds", {
+    help:
+      "Starting amount in seconds to subtract from current wall time for metric series starts. " +
+      "Must be greater or equal to --metrics-past-start-range-min-seconds",
+    type: "int",
+    default: 30 * 60
+  });
+
+  parser.add_argument("--metrics-past-start-range-min-seconds", {
+    help:
+      "Ending amount in seconds to subtract from current wall time for metric series starts. " +
+      "Must be less or equal to --metrics-past-start-range-max-seconds",
+    type: "int",
+    default: 20 * 60
   });
 
   parser.add_argument("--metrics-time-increment-ms", {
@@ -353,6 +371,13 @@ export function parseCmdlineArgs(): void {
     process.exit(1);
   }
 
+  if (CFG.metrics_past_start_range_min_seconds > CFG.metrics_past_start_range_max_seconds) {
+    log.error(
+      "metrics_past_start_range_min_seconds must not be larger than metrics_past_start_range_max_seconds"
+    );
+    process.exit(1);
+  }
+
   if (CFG.max_concurrent_writes > CFG.n_concurrent_streams) {
     log.error(
       "max_concurrent_writes must not be larger than n_concurrent_streams"
@@ -375,7 +400,7 @@ export function parseCmdlineArgs(): void {
     }
   }
 
-  if (CFG.max_concurrent_writes === 0) {
+  if (CFG.max_concurrent_writes === 0) { // TODO this is always zero - remove from CFG?
     CFG.max_concurrent_writes = CFG.n_concurrent_streams;
   }
 
--- a/test/test-remote/looker/src/index.ts
+++ b/test/test-remote/looker/src/index.ts
@@ -51,8 +51,6 @@ import {
 
 import { log } from "./log";
 
-// TODO consolidate these two imports
-import { rndstring, timestampToRFC3339Nano, httpTimeoutSettings } from "./util";
 import * as util from "./util";
 
 export const START_TIME_JODA = ZonedDateTime.now(ZoneOffset.UTC);
@@ -105,22 +103,26 @@ async function main() {
   parseCmdlineArgs();
   const httpServerTerminator = pm.setupPromExporter();
 
-  //let dummystreams: Array<LogSeries>;
   let dummystreams: Array<LogSeries | MetricSeries>;
 
   for (let cyclenum = 1; cyclenum < CFG.n_cycles + 1; cyclenum++) {
     setUptimeGauge();
 
     const numString = `${cyclenum.toString().padStart(3, "0")}`;
-    const guniqueCycleId = `${CFG.invocation_id}-${numString}-${rndstring(4)}`;
+    // invocation_id contains 10 characters of random data by default
+    const guniqueCycleId = `${CFG.invocation_id}-${numString}`;
 
     log.info("enter write/read cycle  %s", cyclenum);
 
-    if ((cyclenum - 1) % CFG.change_streams_every_n_cycles === 0) {
-      // rely on cyclenum to start at 1, 0 % N is 0 so that in the first
-      // run some dummystreams are created!
+    if (cyclenum === 1) {
+      // always create initial streams regardless of update configuration
       log.info("cycle %s: create new dummystreams", cyclenum);
-      dummystreams = await createNewLogSeriess(guniqueCycleId);
+      dummystreams = await createNewSeries(guniqueCycleId);
+    } else if (CFG.change_streams_every_n_cycles > 0 &&
+               (cyclenum - 1) % CFG.change_streams_every_n_cycles === 0) {
+      // update streams at the configured number of cycles
+      log.info("cycle %s: reset dummystreams", cyclenum);
+      dummystreams = await createNewSeries(guniqueCycleId);
     } else {
       log.info(
         "cycle %s: continue to use dummystreams of previous cycle",
@@ -149,7 +151,7 @@ async function main() {
   httpServerTerminator.terminate();
 }
 
-async function createNewLogSeriess(
+async function createNewSeries(
   guniqueCycleId: string
 ): Promise<Array<LogSeries | MetricSeries>> {
   const streams = [];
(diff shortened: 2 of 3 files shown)


test/test-remote/looker/src/args.ts
```diff
@@ -361,7 +400,7 @@ export function parseCmdlineArgs(): void {
     }
   }
 
-  if (CFG.max_concurrent_writes === 0) {
+  if (CFG.max_concurrent_writes === 0) { // TODO this is always zero - remove from CFG?
```
jgehrcke, line 403: > this is always zero - remove from CFG?

I don't understand that part :-)

This is what was set via `--max-concurrent-writes <N>` and I've found this to be one of the most important parameters, I pretty much always set it.

jgehrcke:
> allowing e.g. series counts to remain stable across looker runs 

can you clarify? :) My brain might be slow today; looks like this is an important point and I am not sure I get it! :)

test/test-remote/looker/src/index.ts
```diff
@@ -105,22 +103,28 @@ async function main() {
...
-  //let dummystreams: Array<LogSeries>;
   let dummystreams: Array<LogSeries | MetricSeries>;
 
   for (let cyclenum = 1; cyclenum < CFG.n_cycles + 1; cyclenum++) {
     setUptimeGauge();
 
     const numString = `${cyclenum.toString().padStart(3, "0")}`;
-    const guniqueCycleId = `${CFG.invocation_id}-${numString}-${rndstring(4)}`;
+    // invocation_id contains 10 characters of random data by default
+    const guniqueCycleId = `${CFG.invocation_id}-${numString}`;
```
jgehrcke, line 113: we need to make sure there is no (realistic) upper bound for the number of cycles we can run -- I think that is why I had `${rndstring(4)}`. Maybe change `padStart(3, "0")}` to `padStart(X, "0")}` with X being larger than 3 :).

Also.. what do you think: I somehow like fixed width IDs but maybe here we don't want one after all. Just use `cyclenum`, w/o zero-padding?

jgehrcke:
one of the looker runs failed with
```
2021-07-06T08:07:35.216Z [32minfo[39m: stream initialization finished
2021-07-06T08:07:35.217Z [32minfo[39m: cycle 1: entering write phase
2021-07-06T08:07:35.217Z [32minfo[39m: mark all streams to not keep track of validation info: --skip-read
2021-07-06T08:07:35.223Z [32minfo[39m: re-labeling series for validation info collection took 0.01 s
2021-07-06T08:07:35.223Z [32minfo[39m: create 100000 pushrequestProducerAndPOSTer()
...
2021-07-06T08:07:38.517Z [32minfo[39m: HTTP resp to POST(https://cortex.default.prs-bk-4953-887-g.opstrace.io/api/v1/push (PushMessage(md5=d641ab2a4d4cbf524830d69b27c494fb))):
    status: 400
    body[:500]: user=default: per-metric series limit of 50000 exceeded, please contact administrator to raise it (local limit: 50000 global limit: 0 actual local limit: 50000) for series {__name__="looker_IbQdoEIPZ__001", uniquename="looker-IbQdoEIPZ_-001-25984"}

    headers: {"date":"Tue, 06 Jul 2021 08:07:38 GMT","content-type":"text/plain; charset=utf-8","content-length":"249","connection":"close","vary":"Accept-Encoding","x-content-type-options":"nosniff","strict-transport-security":"max-age=15724800; includeSubDomains"}
    totalTime: 1.277 s
    dnsDone->TCPconnectDone: 0.034 s
    connectDone->reqSent 0.123 s
    reqSent->firstResponseByte: 0.962 s
    
2021-07-06T08:07:38.522Z [31mcrit[39m: consider critical: Error: POST PushMessage(md5=d641ab2a4d4cbf524830d69b27c494fb): Bad HTTP request (see log above)
    at customPostWithRetryOrError (/build/_tscbuild/index.js:752:19)
...
```
(this is in one of the looker stdout/err files in the BK artifacts)

test/test-remote/looker/src/index.ts
```diff
@@ -177,10 +177,25 @@ async function createNewLogSeriess(
...
+            CFG.metrics_past_start_range_max_seconds
+        );
       stream = new MetricSeries(
         {
-          metricName: `looker_${rndstring(4)}`, // might collide among streams, which is OK as long as the label set adds uniqueness
-          uniqueName: streamname, // must not collide among streams
+          // the same across concurrent streams, contains invocation_id
+          // ensure any dashes in metric name are switched to underscores: required by prometheus
+          // see also: https://prometheus.io/docs/concepts/data_model/#metric-names-and-labels
+          metricName: guniqueCycleId.replace(/-/g, "_"),
```
jgehrcke, line 196: The idea is that `--n-concurrent-streams` can be O(10^6). That is, we cannot use the same metric name for all time series written in a cycle.

I think for simplicity we should have each metric name be unique.

If we want to be a little more sophisticated we can use the same metric name for M time series, with M guaranteed to be below an upper bound. (but before building _that_, I thought, it makes sense to simply have unique metric names).

Which requirements do we have for the metric name, maybe in terms of predictability and discoverability? I did use the looker log for finding metric names, and used those to go into Grafana Explore for querying.

test/test-remote/looker/src/metrics/dummyseries.ts
```diff
@@ -205,7 +205,7 @@ export class MetricSeries extends TimeseriesBase {
 
   protected buildLabelSetFromOpts(opts: MetricSeriesOpts): LabelSet {
     // Merge the metric name into it using the well-known special prom label
-    // __name__. Always set `uniquename` and `__name__`. If `opts.labelset` is
```
jgehrcke, line 208: why did we remove "`uniquename` and`" here? I think the code indeed always sets `uniquename=<uniquename>` :)

test/test-remote/looker/src/args.ts
nickbp: Oh this is my fault, I had written this comment early on and then forgot about it. I had done a text search for `max_concurrent_writes` and didn't see any setters (hadn't realized at the time that I should've also checked `max-concurrent-writes`)

test/test-remote/looker/src/index.ts
nickbp: Oh good point. I'll switch this back to being unique - as of this commit it's sharing the same metric name across the concurrent streams, with the difference being in the uniquename label. This also would explain the CI failure with hitting the 50k series cap.

At some point it'd be useful to be able to configure a number of dynamic labels to be included in the metrics to simulate a high-cardinality metric. So for example a test scenario could be to have the metrics each with 10 dynamic labels, with 1000 distinct label permutations against the same metric name. having `metricName=guniqueCycleId` sort of did this - where the number of distinct labels for a common metric name were based on the concurrency value. But this is _really_ unintuitive, and it'd be better to just have an explicit option for label cardinality.

For the names themselves I've personally just been testing things write-only so far and hadn't been thinking too much about the metric names themselves outside of configuring the cardinality.

test/test-remote/looker/src/metrics/dummyseries.ts
nickbp: I think this had been a change that I had started and then backed out... I'll revert the comment change

test/test-remote/looker/src/index.ts
jgehrcke: Btw, let's think of `--n-concurrent-streams` more like `--n-series-per-cycle` -- the "concurrent" part became more or less meaningless / misleading when I introduced the `--max-concurrent-[writes|reads]`

jgehrcke: >  But this is really unintuitive, and it'd be better to just have an explicit option for label cardinality.

We could think of something like `--reuse-metric-name-across-n-series N`. Then require `--n-concurrent-streams N` to be larger than that. Then automatically add further labels to achieve this.

With "cardinatlity" AFAIK being defined as the number of unique combinations we could instead call this thing `--metrics-cardinality C`. We might want to default that to 1 or to 10.

> At some point it'd be useful to be able to configure a number of dynamic labels to be included in the metrics to simulate a high-cardinality metric

With the above's approach, labels would be added automatically. However, that of course may require guidance. I would say, the parameter that might be most useful here is the maximum number of values for a specific label, e.g. `--max-values-per-label M`. We might default that to 10. 

After all, this does not require setting the "number of dynamic labels"; that would be one degree of freedom too much.

test/test-remote/looker/src/args.ts
```diff
@@ -141,6 +143,22 @@ export function parseCmdlineArgs(): void {
     default: 1
   });
 
+  parser.add_argument("--metrics-past-start-range-max-seconds", {
+    help:
+      "Starting amount in seconds to subtract from current wall time for metric series starts. " +
```
jgehrcke, line 148: We might want to try to make this easier to understand in a subsequent iteration.

A suggestion:

> Metrics time series start in the past compared to the initialization wall time. The exact time is picked randomly from an interval. This defines the lower bound of the interval.

Not perfect, but maybe a tiny little clearer?

test/test-remote/looker/src/args.ts
```diff
@@ -215,8 +233,9 @@ export function parseCmdlineArgs(): void {
   parser.add_argument("--change-streams-every-n-cycles", {
     help:
       "Use the same log/metric stream for N cycles, then create a new set of " +
-      "streams (unique label sets). Default: new streams are created with every " +
-      "write/read cycle. For log streams, when a new stream is initialized it " +
+      "streams (unique label sets), or <=0 to reuse streams for the process lifetime. " +
+      "Default: new streams are created with every write/read cycle (1). " +
+      "For log streams, when a new stream is initialized it " +
```
jgehrcke, line 238: Thanks!

jgehrcke: future: we should not allow negative values and document what `0` means : )

test/test-remote/looker/src/args.ts
```diff
@@ -339,6 +371,13 @@ export function parseCmdlineArgs(): void {
     process.exit(1);
   }
 
+  if (CFG.metrics_past_start_range_min_seconds > CFG.metrics_past_start_range_max_seconds) {
+    log.error(
+      "metrics_past_start_range_min_seconds must not be larger than metrics_past_start_range_max_seconds"
+    );
+    process.exit(1);
+  }
```
jgehrcke, line 379: future: we should also warn (maybe not err) when this is too far in the past or in the future (and therefore known to not work with Cortex TSDB)

test/test-remote/looker/src/index.ts
jgehrcke: dropping padding is fine

test/test-remote/looker/src/index.ts
```diff
@@ -105,22 +103,28 @@ async function main() {
...
 
-    if ((cyclenum - 1) % CFG.change_streams_every_n_cycles === 0) {
-      // rely on cyclenum to start at 1, 0 % N is 0 so that in the first
-      // run some dummystreams are created!
+    if (cyclenum === 1) {
+      // we know better than:
+      //  error TS2454: Variable 'dummystreams' is used before being assigned.
+      // always create initial streams regardless of update configuration
       log.info("cycle %s: create new dummystreams", cyclenum);
-      dummystreams = await createNewLogSeriess(guniqueCycleId);
```
jgehrcke, line 123: funny search/replace artifact this was; leading to working code :shrug:

test/test-remote/looker/src/index.ts
```diff
@@ -134,9 +138,8 @@ async function main() {
     // to more reliably crash looker, see CH1288. I'd love for all unexpected
     // errors to crash the runtime instead.
     try {
-      // we know better than:
-      //  error TS2454: Variable 'dummystreams' is used before being assigned.
-      // the first iteration sets dummystreams.
+      // we pass the updated cycle id here even if dummystreams doesnt use it
+      // this ensures the cycle report json has a distinct filename
       //@ts-expect-error: see comment above
       await performWriteReadCycle(cyclenum, dummystreams, guniqueCycleId);
```
jgehrcke, line 144: just noticing while I read this: the `gunique..` way of naming things was past JP expressing that this should be _globally_ unique, i.e. always have enough randomness in it to never collide with anything. The obvious question: is this still true?

test/test-remote/looker/src/index.ts
```diff
@@ -177,10 +177,27 @@ async function createNewLogSeriess(
 
     let stream: MetricSeries | LogSeries;
     if (CFG.metrics_mode) {
+      // Calculate amount to subtract from start time, or avoid random() if subtraction is disabled
+      const subtractSecs = (CFG.metrics_past_start_range_max_seconds === 0)
```
jgehrcke, line 181: Thinking this through, I don't think we should allow `metrics_past_start_range_max_seconds` to be 0. If code does not crash, then this should result in _immediate_ throttle. Which is a little weird. TODO: think this through.

jgehrcke: I think we may also want to rename this params after all:

`--metrics-start-time-delay-lower-bound` 

`--metrics-start-time-delay-upper-bound` 

A difficulty with `max` and `min` in this case are the various corners one has to think around. "max delay" is of course furthest in the past, i.e. the left/lower bound of the time interval. Whereas the way you've used "max" here defines th _minimal_ shift compared to 'now' (the right/upper bound of the time interval).

jgehrcke:
Thank you @nickbp so cool to see more eyes on this code base; thanks for these ideas and for bearing with the state of affairs.

I am merging this now to not make things complicated; but I hope it's totally OK if we change things quickly based on the idea exchange here : )

test/test-remote/looker/src/metrics/dummyseries.ts
```diff
@@ -418,11 +418,11 @@ export class MetricSeries extends TimeseriesBase {
...
-    const genduration = mtimeDiffSeconds(t0);
-    log.debug(
-      "MetricSeriesFragment addSample loop took: %s s",
-      genduration.toFixed(3)
-    );
+    //const genduration = mtimeDiffSeconds(t0);
+    //log.debug(
+    //  "MetricSeriesFragment addSample loop took: %s s",
+    //  genduration.toFixed(3)
+    //);
```
jgehrcke, line 425: I might reactivate this : ) it's a simple benchmark for some core pieces -- of course this yields larger numbers only for extreme cases, which I had done for testing locally.

test/test-remote/looker/src/index.ts
nickbp: I remember trying this with zero, just to see what would happen, and it indeed gave immediate throttling due to the 10 minute check. At the time I was thinking of making the 10 minute value configurable but didn't get around to it. Maybe it could be somehow autoscaled based on the `maxTimeLeapComparedToPreviousFragmentSeconds` math?

test/test-remote/looker/src/index.ts
nickbp: trying out `invocationCycleId` in followup PR

test/test-remote/looker/src/args.ts
nickbp: added a warning in follow-up PR (was already in there and not difficult to add)

test/test-remote/looker/src/args.ts
nickbp: switched to require >= 0

test/test-remote/looker/src/args.ts
nickbp: yeah that's a lot better. updated docs for both the min/max args

test/test-remote/looker/src/index.ts
nickbp: dropping padding in followup PR

test/test-remote/looker/src/index.ts
jgehrcke: > Maybe it could be somehow autoscaled based on ...

Yes. In my current work the interval discussed in this thread here is chosen automatically, based on the wall time coupling parameters (min lag, max lag).

PR #997, opened 2021-07-06
config: hasura action for validating tenant names

MoSattler:
Adds an endpoint for hasura actions to validate tenant names. See #710 

Blocked by [this](https://github.com/opstrace/opstrace/issues/710#issuecomment-875531025).

# Have you...

### DIFF
--- a/go/cmd/config/hasura_handler.go
+++ b/go/cmd/config/hasura_handler.go
@@ -20,6 +20,7 @@ import (
 	"io/ioutil"
 	"net/http"
 	"net/url"
+	"regexp"
 	"strings"
 	"time"
 
@@ -162,6 +163,29 @@ func (h *HasuraHandler) handler(w http.ResponseWriter, r *http.Request) {
 		path := fmt.Sprintf("/api/v1/rules/%s/%s", request.Input.Namespace, request.Input.RuleGroupName)
 		httpresp, err := h.cortexQuery(request.Input.TenantID, "DELETE", path, "")
 		response = actions.ToDeleteResponse("Rule group", httpresp, err)
+
+	case "validateTenantName":
+		var request actions.ValidateTenantNamePayload
+		err = json.Unmarshal(reqbody, &request)
+		if err != nil {
+			writeGraphQLError(w, "invalid request payload")
+			return
+		}
+		regex := "^[a-z0-9]([-a-z0-9]*[a-z0-9])?$"
+		reg, _ := regexp.Compile(regex)
+		invalidNames := request.Input.TenantNames[:0]
+		for _, tenantName := range request.Input.TenantNames {
+			if !reg.MatchString(tenantName) {
+				// Append desired values to slice
+				invalidNames = append(invalidNames, tenantName)
+			}
+		}
+
+		if len(invalidNames) > 0 {
+			errorMessage := fmt.Sprintf("Invalid tenant names: %v.\nThey must match %v", invalidNames, regex)
+			writeGraphQLError(w, errorMessage)
+			return
+		}
 	}
 
 	data, err := json.Marshal(response)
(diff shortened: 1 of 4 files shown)


go/cmd/config/hasura_handler.go
```diff
@@ -162,6 +163,29 @@ func (h *HasuraHandler) handler(w http.ResponseWriter, r *http.Request) {
...
+				// Append desired values to slice
+				invalid_names = append(invalid_names, tenant_name)
+			}
+		}
+
+		if len(invalid_names) > 0 {
+			error_message := fmt.Sprintf("Invalid tenant names: %v.\nThey must match %v", invalid_names, regex)
+			writeGraphQLError(w, error_message)
+			return
+		}
```
MoSattler, line 188: How to I return the equivalent of a `200 ok` response here?

go/cmd/config/hasura_handler.go
```diff
@@ -162,6 +163,29 @@ func (h *HasuraHandler) handler(w http.ResponseWriter, r *http.Request) {
...
 		response = actions.ToDeleteResponse("Rule group", httpresp, err)
+
+	case "validateTenantName":
+		var request actions.ValidateTenantNamePayload
+		err = json.Unmarshal(reqbody, &request)
+		if err != nil {
+			writeGraphQLError(w, "invalid request payload")
+			return
+		}
+		regex := "^[a-z0-9]([-a-z0-9]*[a-z0-9])?$"
```
MoSattler, line 174: Compare with https://github.com/opstrace/opstrace/blob/main/packages/config/src/clusterconfig.ts#L33

terrcin: With the recent compability issues @nickbp encountered we're moved to only allowing `alphanumeric` characters in Tenant names, the regex you've highlighted allows `-` in them which needs to be removed. For reference I've updated and then more recently [simplified the UI validation regex](https://github.com/opstrace/opstrace/pull/999/files#diff-7325966d61eaee7624a540e5774552cdf48b41e5bbd72d76a9db24a29ea48062R23) to being:

```
/^[a-z0-9]{1,63}$/
```

Also there is a character limit of 63 as K8s and/or DNS doesn't like anything longer than that.

MoSattler: In my understanding @jgehrcke introduced the RegEx above, because it's the very same K8S is using.
https://github.com/opstrace/opstrace/blob/d8194a3e9b9dedc5faaa7bce9d5b8fd45a96ba32/packages/config/src/clusterconfig.ts#L31

jgehrcke:
Thank you for proposing something specific @MoSattler!

Architecture discussion :-) Is not supposed to slow this down; but there are important questions.

I am a tiny little bit shocked ... there really is no simpler solution than having to reach out to _another HTTP server_ for doing the validation? :) Well. I trust the answer is 'no'. Downside of this is not _just_ complexity, but also the involved additional error paths. May this HTTP call cross machine boundaries?

Why did we pick the `config API` to implement that HTTP endpoint? I have not really found that bit in https://github.com/opstrace/opstrace/issues/710 :).

Update: I have added my current perspective to https://github.com/opstrace/opstrace/issues/710.

go/cmd/config/hasura_handler.go
jgehrcke: https://hasura.io/docs/latest/graphql/core/actions/action-handlers.html#returning-a-success-response says

> To return a success response, you must send back a response payload of action’s response type. 

I don't understand this sentence.

It also says

> The HTTP status code must be 2xx for a successful response.

Maybe it's enough to send a 200 response with an empty body?

Probably not.

Further down in these docs they say

> To return a success response, you must send the response of the action’s output type (in this case, UserInfo) with a status code 2xx

that example there might be enlightening.

MoSattler:
I'll leave this PR be, until we resolved [this](https://github.com/opstrace/opstrace/issues/710#issuecomment-875531025) discussion in #710

jgehrcke:
Can be closed after https://github.com/opstrace/opstrace/pull/1037.

PR #998, opened 2021-07-06
cli: cluster config: tenant name may not contain hyphen

MoSattler:
Addresses [this](https://github.com/opstrace/opstrace/issues/710#issuecomment-871300624).

# Have you...

### DIFF
--- a/packages/cli/src/schemasv2.test.ts
+++ b/packages/cli/src/schemasv2.test.ts
@@ -1,0 +1,81 @@
+/**
+ * Copyright 2021 Opstrace, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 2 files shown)


packages/cli/src/schemasv2.test.ts
```diff
@@ -0,0 +1,81 @@
...
+
+        it("name may not contain special characters", async () => {
+            const configFile = generateValidClusterConfigFile({ tenants: ["abc^def"] })
+            expect(await ClusterConfigFileSchemaV2.isValid(configFile)).toBe(false)
+        })
+
+        it("name may not contain hyphen", async () => {
+            const configFile = generateValidClusterConfigFile({ tenants: ["my-tenant"] })
+            expect(await ClusterConfigFileSchemaV2.isValid(configFile)).toBe(false)
+        })
```
MoSattler, line 79: Not all of them reflex the *desired* behaviour - but it does reflex the current behaviour. 

Meaning some of those case might be in fact bugs.

packages/config/src/clusterconfig.ts
```diff
@@ -30,7 +30,7 @@ export const CLUSTER_NAME_REGEX = /^[a-z0-9-_]+$/;
 
 // This is the same regex the k8s ecosystem uses for dns-related label values
 // see https://github.com/opstrace/opstrace/issues/710
```
MoSattler, line 32: I guess this comment is outdated now - should I edit/remove it?

jgehrcke: maybe edit it! leave the reference to issue 710 there and add another reference explaining why we also have to remove the dash for now.

jgehrcke:
:rocket:

Issue #944, opened 2021-06-28
ci instability: End-to-end alert tests: 502 upon POST or DELETE to https://<instane-dns-name>/api/v1/rules/testremote


jgehrcke:
https://buildkite.com/opstrace/scheduled-main-builds/builds/2563
AWS
```
[2021-06-27T21:08:46Z] [92m [0m[32m 21 passing[0m[90m (16m)[0m
[2021-06-27T21:08:46Z] [36m [0m[36m 4 pending[0m
[2021-06-27T21:08:46Z] [31m  1 failing[0m
[2021-06-27T21:08:46Z] 
[2021-06-27T21:08:46Z] [0m  1) End-to-end alert tests
[2021-06-27T21:08:46Z]        End-to-end alerts for default tenant:
[2021-06-27T21:08:46Z] 
[2021-06-27T21:08:46Z]       [31mAssertionError [ERR_ASSERTION]: false == true[0m
[2021-06-27T21:08:46Z]       [32m+ expected[0m [31m- actual[0m
[2021-06-27T21:08:46Z] 
[2021-06-27T21:08:46Z]       [31m-false[0m
[2021-06-27T21:08:46Z]       [32m+true[0m
[2021-06-27T21:08:46Z]       [0m[90m
[2021-06-27T21:08:46Z]       at deleteE2EAlertsConfig (test_alerts.ts:292:3)
[2021-06-27T21:08:46Z]       at processTicksAndRejections (internal/process/task_queues.js:95:5)
[2021-06-27T21:08:46Z]       at testE2EAlertsForTenant (test_alerts.ts:327:3)
[2021-06-27T21:08:46Z]       at Context.<anonymous> (test_alerts.ts:422:5)
[2021-06-27T21:08:46Z] [0m
```

and more detail:

```
[2021-06-27T20:52:50Z] 2021-06-27T20:52:50.179Z [32minfo[39m: read token from /tmp/opstrace-bk-2563-562d4de-1-aws/tenant-api-token-default
[2021-06-27T20:52:50Z] 2021-06-27T20:52:50.240Z [32minfo[39m: HTTP resp to DELETE(https://schedul-bk-2563-562-a.opstrace.io/api/v1/alerts):
[2021-06-27T20:52:50Z]   status: 200
[2021-06-27T20:52:50Z]   body[:500]: 
[...]

PR #999, opened 2021-07-06
ui: simplier tenant name regex

terrcin:
Now that tenant names can't have a `-` in them the ui validation regex can be waaay simplier.


terrcin:
Getting this happening: https://github.com/opstrace/opstrace/issues/944

Gonna wait a bit and then try again. This is unrelated to the work in this PR so I might just merge it despite the failure.

MoSattler:
@terrcin may you also update the other tenant name regex for consistency?
https://github.com/opstrace/opstrace/blob/5825d3d97a8c36f78f93ba24fcf4dcd2ecbccf34/packages/config/src/clusterconfig.ts#L35

Maybe we're even able to have a SSOT for this, at least in the TS realm?

jgehrcke:
> Maybe we're even able to have a SSOT for this

We absolutely should have one after all. For simple situations like this I want to at least have said that the source of truth does not need to exist in code, it may exist in docs/wiki etc.

I think the actual challenge here is that we're still in the process of making up our mind. And that we need to disambiguate the 'sane end goal' vs. the 'current state of affairs as of current limitations'.

Considerations:

* The fact that the hyphen is not supported today should be looked at as a temporary problem; let's add back support soon -- that is why I liked how @MoSattler simply removed it non-invasively via https://github.com/opstrace/opstrace/pull/998.
* After all there will need to be some length constraint on a tenant name, but I am not sure if it's 61. Did we test this? Maybe we should limit that to something lower, like 20.

terrcin:
PR now nolonger needed as test failing has been identified and fixed in another PR which contained the work from this PR.

https://github.com/opstrace/opstrace/pull/1019

PR #1001, opened 2021-07-07
docs: update content that is now out of date

triclambert:


triclambert:
Merging this now; will re-introduce the alerts content in a separate PR.  This is so we can preview that content with the already-added images.

PR #1008, opened 2021-07-07
ui: set playwright debugging env var

terrcin:
Getting lots of timeout intermittent failures for browser tests. Enabling debugging to get more information.


Makefile
```diff
@@ -684,6 +684,7 @@ test-browser:
 		-v ${OPSTRACE_BUILD_DIR}/browser-test-results:/build/test/browser/test-results \
 		-e OPSTRACE_CLUSTER_NAME \
 		-e OPSTRACE_CLOUD_PROVIDER \
+		-e DEBUG=pw:api \
```
jgehrcke, line 687: Absolutely the right choice, even w/o too many instabilities. When something happens we want to have useful debug info; especially in those cases when the problem is difficult to reproduce.

PR #1010, opened 2021-07-08
[wip] ui: do auth as part of each browser test rather than per worker

terrcin:
We're having some frustrating intermittent timeout issues on CI with the browser tests at the moment. Sometimes it's looking to be occuring during the auth setup as part of the worker. When this happens we get no debug artefacts from playwright.

This change allows toggling the auth to occur either once per worker as it current occurs, or to happen before each test - which the CI tests are now set to do. Hopefully this will help debug the problems we're having.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -687,6 +687,7 @@ test-browser:
 		-e OPSTRACE_CLUSTER_NAME \
 		-e OPSTRACE_CLOUD_PROVIDER \
 		-e OPSTRACE_INSTANCE_DNS_NAME \
+		-e OPSTRACE_AUTH_METHOD=test \
 		-e DEBUG=pw:api \
 		opstrace/test-browser:$(CHECKOUT_VERSION_STRING) \
  			yarn playwright test --workers 1 --forbid-only --retries 1
--- a/Makefile
+++ b/Makefile
@@ -688,7 +688,6 @@ test-browser:
 		-e OPSTRACE_CLOUD_PROVIDER \
 		-e OPSTRACE_INSTANCE_DNS_NAME \
 		-e OPSTRACE_AUTH_METHOD=test \
-		-e DEBUG=pw:api \
 		opstrace/test-browser:$(CHECKOUT_VERSION_STRING) \
  			yarn playwright test --workers 1 --forbid-only --retries 1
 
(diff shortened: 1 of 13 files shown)


jgehrcke:
Now I created merge conflicts for you; hope you can bear with me here and integrate my hopefully good ideas with yours :).

Makefile
```diff
@@ -684,6 +684,7 @@ test-browser:
 		-v ${OPSTRACE_BUILD_DIR}/browser-test-results:/build/test/browser/test-results \
 		-e OPSTRACE_CLUSTER_NAME \
 		-e OPSTRACE_CLOUD_PROVIDER \
+		-e OPSTRACE_AUTH_METHOD=test \
```
jgehrcke, line 687: Let's be deliberate about differentiating between "login" and "authentication".

* "login" is exchanging primary credentials into secondary credentials (a single sign-on, for example, or a username/password login in the classical sense). This is a _rare_ event compared to "authentication".
* "authentication" usually refers to decorating an HTTP request with authentication proof where the secondary credentials (like an authentication token) is _added_ to the request. This is a _frequent_ event compared to "login".

As of having dealt with this in the past years it really really really makes sense to get aligned on this terminology :-), especially in code and then also issues/bug reports.

jgehrcke: (if this is actually LOGIN_METHOD... let's rename :))

test/browser/utils/authentication.ts
```diff
@@ -14,8 +14,51 @@
...
+  system,
+  cluster,
+  user
+}) => {
+  if (system.workerAuth) {
+    context.addCookies(authCookies);
+    await page.goto(cluster.baseUrl);
+    await page.waitForSelector("[data-test=getting-started]");
+  } else {
+    await performLogin({ page, cluster, user });
```
jgehrcke, line 56: right, this `performLogin()` seems to be good use of terminology. The hole module maybe should be called `login.ts` maybe? Up to you, given the distinction between "login" and "authentication" I tried to provide above.

terrcin:
Now that I've got tests finally passing, just looping back and looking at the above comments. No action needed as current usage/naming fits the suggestions made.

Issue #1004, opened 2021-07-07
ci-instability: test_ui_api create_tenant_and_use_custom_authn_token failed: Expectation not fulfilled within 2100 s


sreis:
**Describe the bug**

```
[2021-07-07T05:46:46Z]   1) test_ui_api
[2021-07-07T05:46:46Z]        create_tenant_and_use_custom_authn_token:
[2021-07-07T05:46:46Z]      Error: Expectation not fulfilled within 2100 s
[2021-07-07T05:46:46Z]       at Context.<anonymous> (test_ui.ts:426:15)
```

For now, it seems to be failing consistently on AWS:

https://buildkite.com/opstrace/scheduled-main-builds/builds/2620#72c302c8-1970-4fea-874b-b0e5de3b47b4/6-4040
https://buildkite.com/opstrace/scheduled-main-builds/builds/2619#9ada46f3-49fc-4dcf-bc7e-8d8914334938/6-3133

jgehrcke:
Transition from ENOUTFOUND to 503 response:
```
[2021-07-07T01:20:31Z] 2021-07-07T01:20:31.746Z info: GET HTTP request to https://cortex.testtenant-ccefvm.schedul-bk-2619-2d9-a.opstrace.io/api/v1/labels failed. Attempt: 6. Error: ENOTFOUND, getaddrinfo ENOTFOUND cortex.testtenant-ccefvm.schedul-bk-2619-2d9-a.opstrace.io. Retry in 15.0 s.
[2021-07-07T01:20:46Z] 2021-07-07T01:20:46.944Z info: GET HTTP request to https://cortex.testtenant-ccefvm.schedul-bk-2619-2d9-a.opstrace.io/api/v1/labels failed. Attempt: 7. Error: undefined, Response code 503 (Service Temporarily Unavailable). Retry in 15.0 s. Response code: 503. Response body[:500]: <html>
[2021-07-07T01:20:46Z] <head><title>503 Service Temporarily Unavailable</title></head>
[2021-07-07T01:20:46Z] <body>
[2021-07-07T01:20:46Z] <center><h1>503 Service Temporarily Unavailable</h1></center>
[2021-07-07T01:20:46Z] <hr><center>nginx</center>
[2021-07-07T01:20:46Z] </body>
[2021-07-07T01:20:46Z] </html>
```

jgehrcke:
was trying to follow the 503 through logs but didn't get anywhere in reasonable time, aborting

[...]

PR #1025, opened 2021-07-09
cli(status): bump check cluster status wait timeout

sreis:
Tried bumping the [request timeout](https://github.com/opstrace/opstrace/blob/29bb931b2ff765b445ce0632d1d76f0971e6879e/packages/installer/src/index.ts#L479) in the hopes it would help fix the issue but it didn't seem to make a difference.

Might as well bump the timeout as [suggested here](https://github.com/opstrace/opstrace/issues/981#issuecomment-877114437).


sreis:
Hit a know flake #1004 .

PR #861, opened 2021-06-09
[WIP] feature work: custom_dns_name / custom_auth0_client_id

jgehrcke:
Update July 1st: I'll now start adding notes for our future selves.

* Idea is to cover these two features in one of both CI runs that we do anyway: GCP/AWS -- I've more or less arbitrarily decided for GCP because this is a little quicker to develop against -- and I've bought into that by building the relevant `gcloud` CLI-based DNS zone management commands for executing the 'manual' DNS setup/teardown work required for the `custom_dns_name` feature.
* CI now has to use the concept of `OPSTRACE_INSTANCE_DNS_NAME` for the source of truth: this is the DNS name pointing to the Opstrace instance; it's not always `<instance_name>.opstrace.io` anymore. Of course this affects test runner logic. I've just made basic changes to the test runners; can and probably should see consolidation/cleanup in the future.

Issue #990, opened 2021-07-05
ci instability: timeout during `waiting for selector "[data-test=getting-started]" to be visible`


jgehrcke:
```
[2021-07-05T17:18:44Z] Running 20 tests using 1 worker
[2021-07-05T17:21:09Z] °F·····°···°°······°
[2021-07-05T17:21:09Z]
[2021-07-05T17:21:09Z]   1) tests/authentication.spec.ts:25:3 › [Chromium] after auth0 authentication user should see homepage 
[2021-07-05T17:21:09Z]
[2021-07-05T17:21:09Z]     page.waitForSelector: Timeout 30000ms exceeded.
[2021-07-05T17:21:09Z]     =========================== logs ===========================
[2021-07-05T17:21:09Z]     waiting for selector "[data-test=getting-started]" to be visible
[2021-07-05T17:21:09Z]     ============================================================
[2021-07-05T17:21:09Z]     Note: use DEBUG=pw:api environment variable to capture Playwright logs.
```
https://buildkite.com/opstrace/prs/builds/4943#fe335df1-340f-4278-bd33-6019e7292dc0/1661-1706

jgehrcke:
https://buildkite.com/opstrace/scheduled-main-builds/builds/2613#3b418eca-bf23-4850-9404-b14faf91848d/1435-1456

jgehrcke:
https://buildkite.com/opstrace/scheduled-main-builds/builds/2625#f501a6a2-a1a4-45d2-9f9a-ca60e8088d94/1886-1910

sreis:
https://buildkite.com/opstrace/scheduled-main-builds/builds/2636#22967163-3d4d-41ed-bff6-42c7050065bb/7206-7382

```
[2021-07-09T20:57:38Z]   1) test_ui_api
[2021-07-09T20:57:38Z]        "before all" hook for "test_grafana_datasource_proxy_loki_get_labels":
[2021-07-09T20:57:38Z]      page.waitForSelector: Timeout 30000ms exceeded.
[2021-07-09T20:57:38Z] =========================== logs ===========================
[2021-07-09T20:57:38Z] waiting for selector "text=Getting Started" to be visible
```

sreis:
https://buildkite.com/opstrace/upgrade-tests/builds/978#2572a3c0-4f0d-4b99-8202-a27a0773b198/7550-7724

jgehrcke:
```
[...]

PR #1036, opened 2021-07-12
fix: set custom_auth0_client_id in controller config

sreis:
#861 Introduced a new field in the user instance config file and CI requires it to be set to run the login flow for the UI tests. 

The code path that handles the controller configuration upgrade does not take into account updates to the user cluster config (ex. add new tenants via the config file). For now, just pass the custom_auth0_client_id. Other fields (ex. number of nodes in the instance) need to be handled on a case by case basis to ensure safe(ish) upgrades.

### DIFF
--- a/packages/upgrader/src/upgrade.ts
+++ b/packages/upgrader/src/upgrade.ts
@@ -17,6 +17,7 @@
 import { select, call, Effect } from "redux-saga/effects";
 import { KubeConfig } from "@kubernetes/client-node";
 
+import { getClusterConfig } from "@opstrace/config";
 import { log, die } from "@opstrace/utils";
 import {
   ControllerResourcesDeploymentStrategy,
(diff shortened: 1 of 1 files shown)


sreis:
Triggering upgrades tests.

sreis:
GCP failed with #990 but at least AWS [was green](https://buildkite.com/opstrace/upgrade-tests/builds/978#_).

jgehrcke:
> Other fields (ex. number of nodes in the instance) need to be handled on a case by case basis to ensure safe(ish) upgrades.

node count is an interesting example -- could it be that as part of the infrastructure upgrade the (potentially new, updated) node count in the user-given config is already taken into account? looks like it, based on code.

here we're talking about the part of the user-given opstrace instance config that is not used during infra creation/upgrade, but solely during controller config construction. interesting distinction :)

PR #1042, opened 2021-07-13
test-remote: rename test_ui to test_ui_api, always run add_tenant test

jgehrcke:


jgehrcke:
@sreis this will require setting environment variables for making upgrade tests. Instead of duplicating code and running behind changes --- should we make an effort to use the _same_ test invocation code in the upgrade tests, i.e. all the following?

https://github.com/opstrace/opstrace/blob/fc6622bde59d781d39fb3cbfee5cbd0a6b1b95d7/ci/deploy-testremote-teardown.sh#L336-L402

sreis:
> should we make an effort to use the same test invocation code in the upgrade tests, i.e. all the following?

I'm never against improving the state of things. Please go ahead and make whatever improvements you see are required.

jgehrcke:
The tenant creation test was executed, looks good.

PR #1046, opened 2021-07-13
cortex: fix ingester replicas cliff at 21 nodes

### LINKED ISSUE #1045: ingester replicas drops from <=20 to >20 nodes
- At nodes=20, we run replicas=Nnodes
- At nodes=21, we run replicas=Nnodes/2

Should update the math so that the number of ingesters doesn't drop as the number of nodes increases, e.g. could keep replicas=20 for nodes=20-40 and then increase at Nnodes/2 for nodes>40

nickbp:
Also remove unnecessary cases that already align with what `roundDownToOdd(Nnodes)` calculates.

(Fixes #1045)

### DIFF
--- a/packages/controller/src/resources/cortex/index.ts
+++ b/packages/controller/src/resources/cortex/index.ts
@@ -173,29 +173,23 @@ export function CortexResources(
       resources: {},
       replicas: select(getNodeCount(state), [
         {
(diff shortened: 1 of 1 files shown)


triclambert (approved):
Lgtm.  

This should not change any behavior.

Issue #1023, opened 2021-07-09
Add structured version info to all build artifacts


jgehrcke:
Make each build artifact (usually: container image) aware of its own 'version string'. Examples:

- controller
- UI app image

Then, make it so that there is an easy way to retrieve this version information.

UI: show the UI app image version info

CLI: show everything, also see https://github.com/opstrace/opstrace/issues/454

In a fresh setup, all versions will be equal.

During an upgrade, versions may differ.

Maybe _after_ an upgrade, versions may differ.

jgehrcke:
This is the foundation for https://github.com/opstrace/opstrace/issues/1024.

MoSattler:
@jgehrcke where is the relevant version currently defined in the build steps. Is it this?
https://github.com/opstrace/opstrace/blob/03f91eb8a57ef9b10f152d87a070df97a64d2b30/.buildkite/pipeline.yml#L3

Thinking about app specifically, I was wondering if it would be enough to pass this as a build var to the `app` docker container, to have it persisted during build time. What do you think?

jgehrcke:
> I was wondering if it would be enough to pass CHECKOUT_VERSION_STRING as a build arg to the docker images, to apply the value during image build time.

Injecting CHECKOUT_VERSION_STRING via env would already be a great start, but I think maybe it's better to have a _static_ and _structured_ thing baked into artifacts, something like a `buildinfo.json`, containing more pieces of information than just CHECKOUT_VERSION_STRING.

The complete build info that we encode as of today in the CLI build artifact (which is not a container image) can be seen here:
https://github.com/opstrace/opstrace/blob/717d5b940dcc95740d51f215aaa90dd6c8eac030/Makefile#L337

So, there's the branch name, the time, ...

I am thinking that we should improve upon that approach.

Thoughts:
* Generate `buildinfo.json` once, and then use it for building artifacts
* For container images, `COPY` it into the image as pretty much the last step (to make most of Docker image layer cache)
* For the pkg-based CLI build, find a way to inject this as a static file

If you'd like to, I can work on that, so that after all there is a `buildinfo.json` in the UI/app container image placed at a well-known path -- you could take it from there, towards https://github.com/opstrace/opstrace/issues/1024.

MoSattler:
@jgehrcke sounds good!

MoSattler:
@jgehrcke do you know in how far #1047 addresses part of this?

jgehrcke:
[...]

PR #869, opened 2021-06-11
feat: integrate cortex-operator

sreis:
Close #845

Issue #20, opened 2020-11-30
create operation: detect "continuation", otherwise fallout such as key material split brain (idempotency violation)


jgehrcke:
With each `create` operation, the CLI generates fresh key material early in the process: an RSA key pair, and derived authentication tokens for the data API. It's doing that before doing any remote state inspection.

When the Opstrace cluster already existed before invoking the current `create` operation, then the `create` operation will push the new public key into the existing cluster. However, the deployments will not pick it up and subsequently, the CLI will use the new (bad) authentication tokens for probing cluster readiness in the last phase of said `create` operation.

That will fail with something like
```
[2020-11-24T20:57:27Z] 2020-11-24T20:57:27.056Z info: https://loki-external.default.bk-2962-d71-a.opstrace.io:8443/loki/api/v1/labels: still waiting, unexpected HTTP response
[2020-11-24T20:57:27Z] 2020-11-24T20:57:27.543Z debug: HTTP response details:
[2020-11-24T20:57:27Z]   status: 401
[2020-11-24T20:57:27Z]   body[:500]: bad authentication token
```

This is a known limitation not yet documented; and a good reason to create this ticket.

This behavior also certainly a violation of an idempotency constraint that we talk about every now and then (where we think and want the `create` operation to be idempotent).

For example, from the current quickstart documentation

> So you know: The CLI is re-entrant, so if you kill it or it dies, it will pick up right where it left off.

This topic raises so many interesting points!

**First-level thinking:** we could detect when the k8s cluster & cluster-internal config (controller config) exists; and then do not regenerate key material and data api auth tokens; try to read existing authentication tokens (and fail when they can't be discovered).

**Second-level thinking**: the previous thought reveals a bigger picture insight: when the cluster already exists then we don't want to overwrite _any_ part of the cluster-internal config state -- that is not well-specified yet.

**Third-level thinking**: ok, we can explicitly _ignore_ the user-given config, emit a clear warning message, and move on with the Nth `create` operation on the same cluster; trying to inform the admin via log messages that we just ignored the config they provided.

**Fourth-level thinking**: but this kind of fallback would be too magic -- we'd be _ignoring_ the user-given config file (or parts of it!) w/o providing a clear signal. No, this must lead to non-zero exit of the current `create` operation.

[...]

PR #1047, opened 2021-07-14
feat(cli,config): info command prints stored/read cluster metadata

### LINKED ISSUE #454: CLI: add cluster-info command

The CLI version flag only prints the CLI installer version. 

```bash
$ opstrace --version 
   Version vX.Y.Z (CLI installer version)
```

When upgrading the Opstracer controller with the current work-in-progress `upgrade` command, the installer and controller versions will diverge. We need a way to show this information to the user.

```bash
$ opstrace cluster-info CLOUD_PROVIDER CLUSTER_NAME
   cluster/infra: version A.B.C
   controller: version X.Y.Z
   installer: version A.B.C
```

## Context

Cluster or infra version refers to all the components not created by the opstrace-controller, for example, the cloud instances, the VPC, and the Kubernetes cluster.

The installer version refers to the Opstrace CLI the user downloads and executes to run commands.

Controller version refers to the Docker image tag used in the Opstrace controller deployment installed in the Kubernetes cluster.

## Possible Technical Solution

Add cluster/infra (CLI version that created the cluster infrastructure) and controller version in the controller config map. Add version subcommand to print cluster/infra version, controller version, and installer CLI version.

estroz:
`opstrace info <aws|gcp> <cluster name>` will print Opstrace and Kubernetes cluster info in the following format:

```
cluster info:
  cluster_version: <kubernetes version>
  controller_version: <image tag>
  installer_version: <installer commit string>
```

Installer version is stored in the controller's configmap and updated on each upgrade event, while other version information is read from k8s cluster configuration at runtime.

Closes #454

Might also address part of #1023

/cc @sreis 

# Have you...

### DIFF
--- a/packages/cli/src/index.ts
+++ b/packages/cli/src/index.ts
@@ -45,6 +45,7 @@ import * as create from "./create";
 import * as destroy from "./destroy";
 import * as list from "./list";
 import * as status from "./status";
+import * as info from "./info";
 import * as upgrade from "./upgrade";
 import * as util from "./util";
 import * as ctoken from "./createTenantAuthToken";
@@ -84,58 +85,45 @@ async function main() {
   if (CLIARGS.instanceName !== undefined) {
     util.validateClusterNameOrDie(CLIARGS.instanceName);
   }
-
-  if (CLIARGS.command == "destroy") {
-    await destroy.destroy();
-    throw new ExitSuccess();
+  switch (CLIARGS.command) {
+    case "destroy":
+      await destroy.destroy();
+      break;
+    case "create":
+      await create.create();
+      break;
+    case "list":
+      await list.list();
+      break;
+    case "status":
+      await status.status();
+      break;
+    case "info":
+      await info.info();
+      break;
+    case "upgrade":
+      await upgrade.upgrade();
+      break;
+    case "ta-create-token":
+      await ctoken.create();
+      break;
+    case "ta-create-keypair":
+      await aks.createKeypair();
+      break;
+    case "ta-pubkeys-add":
+      await aks.addKeyToAuthenticatorConfig();
+      break;
+    case "ta-pubkeys-list":
+      await aks.listKeys();
+      break;
+    case "ta-pubkeys-remove":
+      await aks.removeKey();
+      break;
+    default:
+      throw Error("should never be here");
   }
 
-  if (CLIARGS.command == "create") {
-    await create.create();
-    throw new ExitSuccess();
-  }
-
-  if (CLIARGS.command == "list") {
-    await list.list();
-    throw new ExitSuccess();
-  }
-
-  if (CLIARGS.command == "status") {
-    await status.status();
-    throw new ExitSuccess();
-  }
-
-  if (CLIARGS.command == "upgrade") {
-    await upgrade.upgrade();
-    throw new ExitSuccess();
-  }
-
-  if (CLIARGS.command == "ta-create-token") {
-    await ctoken.create();
-    throw new ExitSuccess();
-  }
-
-  if (CLIARGS.command == "ta-create-keypair") {
-    await aks.createKeypair();
-    throw new ExitSuccess();
-  }
-
-  if (CLIARGS.command == "ta-pubkeys-add") {
-    await aks.addKeyToAuthenticatorConfig();
-    throw new ExitSuccess();
-  }
-
-  if (CLIARGS.command == "ta-pubkeys-list") {
-    await aks.listKeys();
-    throw new ExitSuccess();
-  }
-
-  if (CLIARGS.command == "ta-pubkeys-remove") {
-    await aks.removeKey();
-    throw new ExitSuccess();
-  }
-
-  throw Error("should never be here");
+  throw new ExitSuccess();
 }
 
 /**
@@ -190,6 +178,10 @@ function parseCmdlineArgs() {
   const parserStatus = subparsers.add_parser("status", {
     help:
       "Check the status of an Opstrace instance (experimental, no promises)."
+  });
+  const parserInfo = subparsers.add_parser("info", {
+    help:
+      "Get infrastructure and Opstrace instance version information (experimental, no promises)."
   });
   const parserUpgrade = subparsers.add_parser("upgrade", {
     help: "Upgrade an existing Opstrace instance."
@@ -248,6 +240,7 @@ function parseCmdlineArgs() {
     parserDestroy,
     parserList,
     parserStatus,
+    parserInfo,
     parserUpgrade,
     mainParser
   ]) {
@@ -266,6 +259,7 @@ function parseCmdlineArgs() {
     parserDestroy,
     parserList,
     parserStatus,
+    parserInfo,
     parserUpgrade
   ]) {
     p.add_argument("cloudProvider", {
@@ -276,7 +270,13 @@ function parseCmdlineArgs() {
     });
   }
 
-  for (const p of [parserCreate, parserDestroy, parserStatus, parserUpgrade]) {
+  for (const p of [
+    parserCreate,
+    parserDestroy,
+    parserStatus,
+    parserInfo,
+    parserUpgrade
+  ]) {
     p.add_argument("instanceName", {
       help:
         "The Opstrace instance name ([a-z0-9-_], no more than 23 characters).",
--- a/packages/cli/src/info.ts
+++ b/packages/cli/src/info.ts
@@ -1,0 +1,151 @@
+/**
+ * Copyright 2021 Opstrace, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { call, race, delay } from "redux-saga/effects";
+import createSagaMiddleware from "redux-saga";
+import { createStore, applyMiddleware } from "redux";
+
+import {
+  smErrorLastResort,
+  getKubeConfigForOpstraceClusterOrDie
+} from "./util";
+import * as cli from "./index";
+import { log, die, SECOND, parseVersionFromImageTag } from "@opstrace/utils";
+import {
+  STORAGE_KEY,
+  CONFIGMAP_NAME,
+  CONTROLLER_NAME,
+  CONTROLLER_NAMESPACE,
+  LatestControllerConfigSchema
+} from "@opstrace/controller-config";
+import {
+  AppsV1Api,
+  CoreV1Api,
+  VersionApi,
+  VersionInfo,
+  V1ConfigMap,
+  V1Deployment
+} from "@kubernetes/client-node";
+
+/**
+ * getClusterInfo reads Kubernetes and Opstrace cluster state
+ * from the apiserver then prints this information in human-readable format.
+ */
+async function getClusterInfo() {
+  // All information printed by this command hinges on k8s cluster availability.
+  // getKubeConfigForOpstraceClusterOrDie will log an indicative message and
+  // throw an exception if the kubeconfig it generates cannot contact a cluster
+  // at the expected host, which is desirable here.
+  const kcfg = await getKubeConfigForOpstraceClusterOrDie(
+    cli.CLIARGS.cloudProvider,
+    cli.CLIARGS.instanceName
+  );
+
+  // Read controller ConfigMap and die if not found.
+  let cm: V1ConfigMap;
+  try {
+    const coreClient = kcfg.makeApiClient(CoreV1Api);
+    cm = (await coreClient.readNamespacedConfigMap(CONFIGMAP_NAME, "default"))
+      .body;
+  } catch (err) {
+    die(
+      `failed to read Opstrace controller config map: ${
+        err.response ? err.response.body.message : err
+      }`
+    );
+  }
+  const cfgJSON = JSON.parse(cm.data?.[STORAGE_KEY] ?? "");
+  if (cfgJSON === "") {
+    die(`invalid Opstrace controller config map`);
+  }
+
+  log.debug(`controller config: ${JSON.stringify(cfgJSON, null, 2)}`);
+
+  // Set lastCLIVersion to the config's latest value if available. CLI metadata
+  // will be present in the latest config version(s) only, but its non-existence should
+  // not result in a command error.
+  let lastCLIVersion = "undefined";
+  if (LatestControllerConfigSchema.isValidSync(cfgJSON)) {
+    const cfg = LatestControllerConfigSchema.validateSync(cfgJSON);
+    lastCLIVersion =
+      cfg.cliMetadata.allCLIVersions[cfg.cliMetadata.allCLIVersions.length - 1]
+        ?.version;
+  }
+
+  // Determine the current controller version.
+  let cd: V1Deployment;
+  try {
+    const appsClient = kcfg.makeApiClient(AppsV1Api);
+    cd = (
+      await appsClient.readNamespacedDeployment(
+        CONTROLLER_NAME,
+        CONTROLLER_NAMESPACE
+      )
+    ).body;
+  } catch (err) {
+    die(
+      `failed to read Opstrace controller deployment: ${
+        err.response ? err.response.body.message : err
+      }`
+    );
+  }
+  // Image will always be defined since it is retrieved from the controller deployment's podspec.
+  const controllerVersion = parseVersionFromImageTag(
+    cd.spec!.template.spec!.containers[0].image!
+  );
+
+  // Read Kubernetes cluster info for the full server version.
+  let k8sSrvInfo: VersionInfo;
+  try {
+    k8sSrvInfo = (await kcfg.makeApiClient(VersionApi).getCode()).body;
+  } catch (err) {
+    die(`failed to read Kubernetes server info: ${err}`);
+  }
+
+  // Finally, write cluster info.
+  console.log(`
+cluster info:
+  kubernetes_version: ${k8sSrvInfo.gitVersion}
+  controller_version: ${controllerVersion}
+  installer_version: ${lastCLIVersion}
+  `);
+}
+
+// Race to get cluster info and fail if it takes longer than 60 seconds.
+function* rootTaskInfoCore() {
+  const { timeout } = yield race({
+    info: call(getClusterInfo),
+    timeout: delay(60 * SECOND)
+  });
+
+  if (timeout) {
+    throw new Error("timeout getting Opstrace cluster info");
+  }
+}
+
+// Wrapper to make redux-saga happy.
+function* rootTaskInfo() {
+  yield call(rootTaskInfoCore);
+}
+
+// Empty reducer to make redux-saga happy.
+const reducer = (state = 0) => state;
+
+export async function info(): Promise<void> {
+  const sm = createSagaMiddleware({ onError: smErrorLastResort });
+  createStore(reducer, applyMiddleware(sm));
+  await sm.run(rootTaskInfo).toPromise();
+}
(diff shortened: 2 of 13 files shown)


estroz:
will update CLI docs once the command name/output is agreed on.

packages/cli/src/info.ts
```diff
@@ -0,0 +1,133 @@
...
+async function getClusterInfo() {
+  const kcfg = await getKubeConfigForOpstraceClusterOrDie(
+    cli.CLIARGS.cloudProvider,
+    cli.CLIARGS.instanceName,
+  );
+
+  // Read controller ConfigMap and die if not found.
+  let cm: V1ConfigMap;
+  try {
+    const coreClient = kcfg.makeApiClient(CoreV1Api);
```
estroz, line 53: These kubeconfig API calls work, but is there a better library to use, like update reducers?

packages/cli/src/index.ts
```diff
@@ -84,58 +85,45 @@ async function main() {
...
+      await aks.addKeyToAuthenticatorConfig();
+      break;
+    case "ta-pubkeys-list":
+      await aks.listKeys();
+      break;
+    case "ta-pubkeys-remove":
+      await aks.removeKey();
+      break;
+    default:
+      throw Error("should never be here");
```
jgehrcke, line 123: :+1: thanks

packages/cli/src/index.ts
```diff
@@ -191,6 +179,10 @@ function parseCmdlineArgs() {
     help:
       "Check the status of an Opstrace instance (experimental, no promises)."
   });
+  const parserInfo = subparsers.add_parser("info", {
+    help:
+      "Get infrastructure and Opstrace cluster information (experimental, no promises)."
```
jgehrcke, line 184: I like that. 

Please replace "cluster" with "instance", for terminology consistency (we made that decision a while ago :shrug:).

Let's maybe say "version information" because that's all there is for now.

jgehrcke:
@estroz I am genuinely impressed with how nicely you found your way through our chaotic code base :).

Thanks for this porposal.

Sometimes it makes a little more sense to first define 'desired / correct behavior' in a discussion before writing code, and this here might be just one of these cases; because I think we misled you a bit here and I am a little sorry for that.

The fact that we _update_ the controller config as part of `create` seems like this is well thought through and by design; but no. We should not do this; it is a source of problems. Running `create` twice is not generally supported as of today and known to lead to a number of problems, as of a rather huge design gap. Also see https://github.com/opstrace/opstrace/issues/20, a tiny quote:

> this means the same command run twice (unaltered) can't just magically do the right thing in an idempotent fashion. There's a logical conflict here. Even with proper config upgrade mechanism we should not magically switch between 'initial create' and 'config upgrade'.

(note to self: it's really time to make changes to the installer code base to make that clear and to error out)

That is, the idea of setting `lastInstallerVersion` in the controller config as part of `create` is flawed.

But I think we can do a more or less tiny tweak to this PR, and persist the installer version of the _initial_ install:

During initial `create`, i.e. in `installer/index.ts`, instead of `lastInstallerVersion` let's persist the concept of the _initial_ installer version (maybe call this `initialInstallerVersion`). That's useful information.

Then, subsequently, we can add to this structure in the _upgrader_. Not sure if you've seen the upgrader, and how it mutates the controller config:
https://github.com/opstrace/opstrace/blob/2344fa516f6b96b3d7fde54d45b6ccc35e4bba4f/packages/upgrader/src/index.ts#L194

The upgrader itself as of today doesn't really respect much of the input it gets -- most of the user-given instance config -- but at least _by design_ it is supposed to mutate the controller config. It can therefore make sense to introduce a `lastUpgraderVersion`, as part of that mutation.

And honestly, while we are here, instead of doing just `lastUpgraderVersion` we could make each upgrade add to a JSON array (i.e. an ordered structure) and add timestamps; and with that record a history of initial install and all upgrades in the controller version (I can see that this may seem dirty in many ways, but honestly I think this is pragmatic and there will be little practical fallout and it might come in handy every now and then).

jgehrcke (changes requested):
Forgot to leave the previous comment as a review, doing this now :).

jgehrcke:
you motivated me to do this tiny part: https://github.com/opstrace/opstrace/pull/1050

packages/controller-config/src/schemav2.ts
```diff
@@ -66,11 +66,18 @@ export const ControllerConfigSchemaV2 = yup
     infrastructureName: yup.string().required(),
     aws: yup.mixed<AWSConfig | undefined>(),
     gcp: yup.mixed<GCPConfig | undefined>(),
-    controllerTerminated: yup.bool().default(false)
+    controllerTerminated: yup.bool().default(false),
+    installerMetadata: yup.object({
+      lastInstallerVersion: yup.string(),
+    }),
```
sreis, line 72: Because this adds a new field to the v2 schema it'll fail during controller configuration upgrades. The reason being that [upgradeControllerConfigMapToLatest](https://github.com/opstrace/opstrace/blob/e4c1c289ecfb917aca62143d6c942bd7ceffe265/packages/controller-config/src/schema.ts#L79-L107) will not be able to validate the (now old) v2 schema configuration on a running instance because the `installerMetadata` is now a required field.

One option is to make `installerMetadata` not required the other would be to introduce a new schema version (v3) and a `V2toV3` upgrade function that would set the new field.

Makes sense?

sreis (commented):
This is a great contribution and overall I like the direction. 

Unfortunately, we don't have CI configured to run jobs when the PR is from external contributors so I will volunteer to clean up any potential failures in the main branch when this is merged.

packages/cli/src/info.ts
estroz: The kubeconfig returned from [`getKubeConfigForOpstraceClusterOrDie`](https://github.com/opstrace/opstrace/blob/696f18892cf3ec621129c9054489105265233797/packages/cli/src/util.ts#L303) appears to just generate a kubeconfig dynamically using the same [underlying function](https://github.com/opstrace/opstrace/blob/696f18892cf3ec621129c9054489105265233797/packages/cli/src/util.ts#L348) as [`getKubecfgIfk8sClusterExists`](https://github.com/opstrace/opstrace/blob/main/packages/uninstaller/src/index.ts#L91-L105); will they not yield the same result? Perhaps exiting is not the right behavior though, and returning "undefined" for cluster version for an unavailable cluster using the function you suggested is desirable instead. Thoughts?

packages/controller-config/src/schemav2.ts
estroz: I was concerned that adding a new field would break some contract. Typically when I think of changes that require a major version upgrade, I think of breaking changes to an existing contract; in this case, `installerMetadata` is a net-new field that isn't _required_ for any CLI functionality to work; however I could see some project component eventually relying on it's presence for all v2 configs when that isn't guaranteed if it remains optional.

How about schema v3alpha1, in case `installerMetadata`'s schema is changed soon based on new ideas? Is there a stabilization timeline you typically follow?

estroz:
@jgehrcke thanks for the thorough response. Your clarification on `create`'s intended behavior makes sense. A few questions:

>During initial create, i.e. in installer/index.ts, instead of lastInstallerVersion let's persist the concept of the initial installer version (maybe call this initialInstallerVersion). That's useful information.

Is treating the initial and subsequent installer versions different useful? Assuming all upgrader versions are stored in a time-ordered list, could we not treat the first item in that list the initial installer version and do away with a separate field?

>And honestly, while we are here, instead of doing just lastUpgraderVersion we could make each upgrade add to a JSON array (i.e. an ordered structure) and add timestamps; and with that record a history of initial install and all upgrades in the controller config (I can see that this may seem dirty in many ways, but honestly I think this is pragmatic and there will be little practical fallout and it might come in handy every now and then).

I'm fine with adding each upgrader version to an array, as it may help with debugging. I will note that the max ConfigMap size is [1MiB](https://kubernetes.io/docs/concepts/configuration/configmap/#motivation), which is more than reasonable for a typical config file; however, now we're discussing adding a field whose size scales with number of upgrades, which could cause issues as more config fields are added and the array is not garbage collected. Thoughts on this?

>Cosmetically, about cluster_version: if this is supposed to show the underlying k8s version, then let's maybe call this k8s_version :).

:+1: 

>Unfortunately, we don't have CI configured to run jobs when the PR is from external contributors so I will volunteer to clean up any potential failures in the main branch when this is merged.

@sreis I really appreciate it! I'll attempt to make this as failure/bug-free as possible prior to merge.

jgehrcke:
> Assuming all upgrader versions are stored in a time-ordered list, could we not treat the first item in that list the initial installer version and do away with a separate field?

Yes. How would we call the single list?

> Is treating the initial and subsequent installer versions different useful?

Useful maybe yes. But not essential; from an information / storage point of view, I agree to that.

It may be useful for human interpretation I think: I think I like the idea of using expressive variable names here where it's immediately obvious that there is the "initial installer version" (or something equally expressive) keeping record of the version used for `opstrace create`  and then the "upgrader versions" (or something equally expressive), keeping record of the version(s) used for `opstrace upgrade`.

That is, I think this is 'only' about naming and semantics and expressiveness.



Feel free to make a decision here!

> I will note that the max ConfigMap size is 1MiB, which is more than reasonable for a typical config file; however, now we're discussing adding a field whose size scales with number of upgrades, which could cause issues as more config fields are added and the array is not garbage collected. Thoughts on this?

Let's add garbage collection when it's needed. It's likely never needed. We have a lot of time / headroom to figure that out. Back of the envelope calc: when each `upgrade` entry adds 50 bytes then we have space for about 20000 of those which means more than 350 years of upgrades when one does an upgrade every 7 days.

packages/controller-config/src/schemav2.ts
sreis: > How about schema v3alpha1, in case installerMetadata's schema is changed soon based on new ideas? Is there a stabilization timeline you typically follow?

At this point, we don't have a schedule to handle these schema upgrades. AFAIK, there are no plans to touch the controller config map schema in the short term. You can go ahead and introduce `V3` then.

packages/controller-config/src/schemav3.ts
```diff
@@ -0,0 +1,84 @@
...
+    tlsCertificateIssuer: yup
+      .mixed<"letsencrypt-staging" | "letsencrypt-prod">()
+      .oneOf(["letsencrypt-staging", "letsencrypt-prod"])
+      .required(),
+    infrastructureName: yup.string().required(),
+    aws: yup.mixed<AWSConfig | undefined>(),
+    gcp: yup.mixed<GCPConfig | undefined>(),
+    controllerTerminated: yup.bool().default(false),
+    cliMetadata: yup.object({
+      allCLIVersions: yup
```
estroz, line 69: @jgehrcke I've made this into an array, and changed the name to `allCLIVersions` since the `opstrace` CLI will always be the agent creating or updating the config. If you have a better suggestion I'll use it, since I do not have a strong opinion here.

packages/controller-config/src/schemav2.ts
estroz: Done.

packages/cli/src/index.ts
estroz: Done.

packages/controller-config/src/schemav3.ts
```diff
@@ -0,0 +1,84 @@
...
+      .string()
+      .typeError()
+      .strict(true),
+    disable_data_api_authentication: yup.bool().required(),
+    uiSourceIpFirewallRules: yup.array(yup.string()).ensure(),
+    apiSourceIpFirewallRules: yup.array(yup.string()).ensure(),
+    postgreSQLEndpoint: yup.string().notRequired(),
+    opstraceDBName: yup.string().notRequired(),
+    envLabel: yup.string(),
+    // Note: remove one of cert_issuer and `tlsCertificateIssuer`.
```
estroz, line 55: Since I'm creating a new config schema, should I handle this now?

estroz:
I was not able to successfully run `create` locally (informer creation kept 404'ing) so new changes have not been tested yet. Will try again with a clean environment soon.

jgehrcke (approved):
:rocket: I am happy with that current state! @sreis will you take another look w.r.t. the schema versions?

sreis:
> We should also trigger a CI run. :)

Yes, but we disabled running builds from forks.

> I was not able to successfully run create locally (informer creation kept 404'ing) so new changes have not been tested yet. Will try again with a clean environment soon.

Let's wait for further feedback from @estroz on his local testing. Depending on that, we can try to enable builds from PRs and kick off a CI build. When that is finished, we'll need to disable builds from PRs again.

packages/upgrader/src/index.ts
```diff
@@ -198,8 +198,6 @@ function* triggerControllerDeploymentUpgrade() {
     // handle upgrades from clusters that are not running the cortex-operator
     yield call(cortexOperatorPreamble, kubeConfig);
 
-    yield call(upgradeControllerConfigMap, kubeConfig);
```
estroz, line 201: @sreis was this call added for a reason in https://github.com/opstrace/opstrace/pull/869/files#diff-d85ea48a17112e49ec6714d9c57517cc0192ca603adad37fbb65bc893c52db0bR201? It looks extraneous to me but I can't be sure.

packages/controller-config/src/schemav3.ts
```diff
@@ -0,0 +1,106 @@
...
+ * The yup date() schema uses a transformer to parse date strings,
+ * like those retrieved from the controller-config's raw value
+ * stored in a ConfigMap. This works when not running in strict mode,
+ * which turns off transformers. When upgrading the config however,
+ * strict validation must be turned on to make sure an old schema that
+ * is a strict subset (same but fewer values) of of a new one doesn't
+ * coerce to the new schema. Then validation fails. This custom schema
+ * permits a string timestamp with ISO format enforced.
+ */
+const timestampSchema = yup
```
estroz, line 32: Had to add this Date parsing workaround.

estroz:
Tested this an seems to work, using explicit image tags so the upgrade checks pass:
```console
$ opstrace create aws test-cluster -c config.yaml
$ kubectl get configmap opstrace-controller-config -o json | jq -r '.data."config.json" | fromjson | .cliMetadata'
{
  "allCLIVersions": [
    {
      "version": "3683e58ec-dev",
      "timestamp": "2021-07-24T21:20:00.332Z"
    }
  ]
}
$ opstrace info aws test-cluster
...
cluster info:
  kubernetes_version: v1.18.16-eks-7737de
  controller_version: 2e4ef97a-ci
  installer_version: 3683e58ec-dev
$ opstrace upgrade aws test-cluster -c config.yaml
$ kubectl get configmap opstrace-controller-config -o json | jq -r '.data."config.json" | fromjson | .cliMetadata'
{
  "allCLIVersions": [
    {
      "version": "3683e58ec-dev",
      "timestamp": "2021-07-24T21:20:00.332Z"
    },
    {
      "version": "3683e58ec-dev",
      "timestamp": "2021-07-24T21:20:05.172Z"
    }
  ]
}
$ opstrace info aws test-cluster
...
cluster info:
  kubernetes_version: v1.18.16-eks-7737de
  controller_version: a0fb5b72-ci
  installer_version: 3683e58ec-dev
```

packages/upgrader/src/index.ts
sreis: No, it's a good catch from you to delete one of the calls.

packages/controller-config/src/schemav3.ts
sreis: I'd say no otherwise this will turn out to be an even bigger PR.

packages/cli/src/info.ts
sreis: If we can't get the kubeconfig to access the cluster, then we might as well fail early. wdyt?

sreis:
Going to try to get Buidkite to run a CI build for this PR now.

sreis:
Got it running but it failed in the build step:

```
[2021-07-26T18:27:24Z] + make build-and-push-controller-image
...
[2021-07-26T18:31:12Z] src/schemav3.ts(18,26): error TS2307: Cannot find module 'date-fns' or its corresponding type declarations.
```

The changes to the schema affect the controller which you haven't caught since you haven't built the docker image. Try running `make tsc` locally and let me know if it catches the error.

estroz:
@sreis surprisingly `make tsc` works fine in my env, and this wasn't an issue when running `make cli` either. I'm guessing `date-fns` is installed globally somewhere in my env. Will add this dependency and push.

packages/cli/src/info.ts
estroz: Yep agreed! That function is being used above to instantiate `kcfg`.

jgehrcke:
@triclambert @PatrickHeneise why would commits like this end up on this branch?

https://github.com/opstrace/opstrace/pull/1047/commits/c6dbf871f429e5edf650f75cfa8cc3725179394a

PatrickHeneise:
> @triclambert @PatrickHeneise why would commits like this end up on this branch? Do we need to tame that robot?
> 
> [c6dbf87](https://github.com/opstrace/opstrace/commit/c6dbf871f429e5edf650f75cfa8cc3725179394a)

Strange. The Action should run on branches that are not `main` and only if `/docs` are modified, which isn't the case here. I'll review the trigger conditions, thanks for raising the issue.

jgehrcke (changes requested):
TLDR: for encoding timestamps as text, I'd love to fully specify the time string format and to settle on using `js-joda` for the parsing.

---

Instead of `"must be an ISO date string"` let's maybe lock down a specific format :) More specifically, let's do RFC 3339 
- without the fractional second part
- with the `Z` specifier 

Example: `1990-12-31T23:59:59Z` 

`yyyy-MM-ddTHH:mm:ssZ`


One of the lessons I've learned in the past years is that "an ISO date string" can have a gazillion different shapes, assuming we talk ISO 8601. The separator may be a T or a white space, the fraction of a second maybe present or not; it may have this or that precision -- and the time zone may be encoded in many different way, for example via `Y` or `+00:00`.

For datetime operations including serialization and deserialization we've already bought into `js-joda`, a beautiful JavaScript port of a famous lib from the Java ecosystem. Example:

https://github.com/opstrace/opstrace/blob/5d8a612c14fe042ce219a16e70704b511293b7b4/test/test-remote/testutils/index.ts#L31

I think this would do the validation job:

```javascript
> LocalDateTime.parse("1990-12-31T23:59:59Z", DateTimeFormatter.ofPattern("yyyy-MM-dd'T'HH:mm:ss'Z'"))
LocalDateTime {
  _date: LocalDate { _year: 1990, _month: 12, _day: 31 },
  _time: LocalTime { _hour: 23, _minute: 59, _second: 59, _nano: 0 }
}
```

When passing bad values, the lib throws nice exceptions, such as
```
DateTimeParseException: Text '1990-12-31T23:59:60Z' could not be parsed: Invalid int value for SecondOfMinute: 60: 1990-12-31T23:59:60Z, at index: 0
-------
Caused by: DateTimeException: Invalid int value for SecondOfMinute: 60
```

One could also parse into a `ZonedDateTime` object via `DateTimeFormatter.ofPattern("yyyy-MM-dd'T'HH:mm:ssX")` -- the `X` format specifier however allows for many options, see https://js-joda.github.io/js-joda/manual/formatting.html#format-patterns -- `Z; -08; -0830; -08:30; -083015; -08:30:15;`. That's why I would prefer to _require_ Z, and hence the `LocalDateTime.parse()` proposal above.

I want to add that I thought about this before and then thought 'ah, I keep this thinking for myself'. Now that we talk about this new datetime library `date-fns` and how it was missing in CI I changed my mind :).

jgehrcke:
@estroz do you have a checkbox on the right-hand side here "allow edits from maintainers"? If you check this, I am happy to do the js-joda thing, and I think with that we also get another CI run for free :)

https://docs.github.com/assets/images/help/pull_requests/allow-maintainers-to-make-edits-sidebar-checkbox.png

https://docs.github.com/en/github/collaborating-with-pull-requests/working-with-forks/allowing-changes-to-a-pull-request-branch-created-from-a-fork

> Additionally, you can modify an existing pull request to let repository maintainers make commits to your branch.

estroz:
@jgehrcke sounds great, thanks! Yes you should be able to push to this branch.

jgehrcke:
OK I rebased on latest `main` (in upstream i.e. this repo) and removed the 'reading time' commits while I was there. Added my commit for said timestamp string validation (and creation).

I also initiated a CI build through our buildkite pipeline by setting `branch` to `refs/pull/1047/head` -- setting `branch` to `estroz:feat/cli-info-cmd` did not work CC @sreis (not sure which method you used before).

jgehrcke:
> I'll review the trigger conditions, thanks for raising the issue

Thank you @PatrickHeneise and also thanks for the quick response!!

jgehrcke (approved):
:rocket: 

CI went green. Let's land this, and change all the things we'd still like to improve after that.

Thank you so much @estroz for bearing with us here!

estroz:
@jgehrcke @sreis happy to contribute, thanks for assisting!

PR #1048, opened 2021-07-14
[wip] ui: update constraint violating tenant names before constraint is added

terrcin:
If a cluster has tenant names that contain chars that violate the newly added constraint on the tenant.name field upgrading the cluster will fail (well, atleast the hasura migrations will) with the following error:

```
FATA[0000] apply failed: [postgres-error] query execution failed ($.args[0].args)
File: '1626132643036_alter_table_public_tenant_add_check_constraint_tenant_name_validation/up.sql'
{
    "check_metadata_consistency": false,
    "sql": "alter table \"public\".\"tenant\" add constraint \"tenant_name_validation\" check (name ~ '^[a-z0-9]{2,63}$'::text);\n"
}
[23514] FatalError: check constraint "tenant_name_validation" is violated by some row 
```

This PR inserts a migration that runs immediately before the constraint is added. It ensures all tenant names will pass the new constraint by removing unsupport charts and appending a random number if the new name is not unique.

### DIFF
--- a/packages/app/migrations/1626017036575_simplify-existing-tenant-names/down.sql
+++ b/packages/app/migrations/1626017036575_simplify-existing-tenant-names/down.sql
@@ -1,0 +1,4 @@
+do $$
+begin
+  --nothing to do
+end $$
--- a/packages/app/migrations/1626017036575_simplify-existing-tenant-names/up.sql
+++ b/packages/app/migrations/1626017036575_simplify-existing-tenant-names/up.sql
@@ -13,22 +13,22 @@ begin
     if not cleaned_name = tenant_row.name then
       raise notice 'updating tenant name: "%"', tenant_row.name;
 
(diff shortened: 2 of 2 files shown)


terrcin:
For testing this I have been manually running the hasura migration against my clusters. This is an example of what I've been testing:

Screen Shot 2021-07-14 at 4 28 02 PM

The `key` column is what the `name` column was before the migration modified it.

jgehrcke:
> If a cluster has tenant names that contain chars that violate the newly added constraint on the tenant.name field upgrading the cluster will fail (well, atleast the hasura migrations will) with the following error:

It is important to say that it isn't just the upgrade that 'breaks', but the state before the upgrade was already 'broken' -- said tenants never worked (in the sense that they exposed a functional API). So, it would also be fine to delete them IMO. :) But a human operator can still do this manually of course.

Issue #237, opened 2021-01-05
cli: store cluster state in directory (or single-file container)


jgehrcke:
- log files
- opstrace DNS service credentials
- what else?

Use a convention to translate the cluster name an provider to the path (of directory or single-file container).

That would make it easier to switch between clusters.

There may be downsides, too.

Notes:
- thought often times: cluster name should be in the log file names
- credential files need to have more expressive names, be hidden, and/or otherwise consolidated (only one of both should be needed)

This is brainstorm level.

jgehrcke:
affects tenant API tokens, of course

sreis:
A couple of questions to understand what you're trying to achieve.

> store cluster state in directory

In a way, we are already doing this. But you are proposing the target directory wouldn't be the current working dir?

> Use a convention to translate the cluster name an provider to the path (of directory or single-file container).

How would this work in practice, for example, for a user that wants cluster A in GCP and cluster B in AWS? What happens if the user runs the CLI first with `/home/user/Downloads` and then `/home/user/foo` as current working directories?

triclambert:
Some additional thoughts...

* Configuration file
* kubeconfig.conf
* Downloaded YAML from one of our integrations
* Ancillary workload or configuration files

> How would this work in practice...

Download the CLI twice, once in each directory.  Inelegant, but practical.  (CLI management/versioning is probably something we should talk about in the future.)

jgehrcke:
> But you are proposing the target directory wouldn't be the current working dir?

Yes, right. Could encode the instance name in the directory, and then put all files into that directory (tenant API authentication tokens, and also DNS service authentication state).

> How would this work in practice, for example, for a user that wants cluster A in GCP and cluster B in AWS? What happens if the user runs the CLI first with /home/user/Downloads and then /home/user/foo as current working directories

Oh, to keep things simple that state directory would still be stored _relative to CWD_. That is, when I do

```
cd /home/jp
./opstrace create aws powerinstance ..
```
I will get a directory  `/home/jp/opstrace-powerinstance` and when I do 

```
cd /home/user/Downloads
./opstrace create gcp powerinstance ..
```
I will get a directory  `/home/user/Downloads/opstrace-powerinstance`

Yasumoto:
Some decent prior art:

[...]

PR #1050, opened 2021-07-14
installer: do not update controller config if it exists

jgehrcke:
See #20 and #237, also commit messages and code comments.

### DIFF
--- a/lib/kubernetes/src/index.ts
+++ b/lib/kubernetes/src/index.ts
@@ -2,7 +2,16 @@
  * Copyright 2020 Opstrace, Inc.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
- * you may not use this file except in compliance with the License.
+
+      // Handle transient/retryable errors here, mainly TCP connect() timeout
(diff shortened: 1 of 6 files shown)


lib/kubernetes/src/index.ts
```diff
@@ -2,7 +2,16 @@
...
+
+      // Handle transient/retryable errors here, mainly TCP connect() timeout
+      // For context, see opstrace-prelaunch/issues/1039 and
+      // https://github.com/kubernetes-client/javascript/issues/544
+      // TODO: handle other transient ones, ENOUTFOUND and 5xx HTTP responses.
+      if (e.code && e.code === "ETIMEDOUT") {
+        log.info("retry config map creation/update soon (%s)", e.message);
+        await sleep(3);
+        continue;
+      }* you may not use this file except in compliance with the License.
```
jgehrcke, line 14: oh, bad patch :D

jgehrcke: updated.

PR #1053, opened 2021-07-14
loki: move memberlist bind port to 7947

sreis:
Move to another port to avoid any possibility of joining pods belonging to cortex memberlist ring.

Copying the comment from https://github.com/opstrace/private/issues/169#issuecomment-879736441

> Just conjecture at this point but the constant crash looping of the pods could lead to the kubelet recycling IP addresses. Meaning, at some point, a pod belonging to one memberlist ring (cortex) could get an IP that was recently assigned to a pod in the other memberlist ring (loki). This would mean the new pod would connect to cortex memberlist ring but get connections from members of the loki ring. This would create the "bridge" between the two rings. I thought about this because left_ingesters_timeout is set to 5m0s which seems like enough time for an ingester to be up and advertise itself on the memberlist ring and accept connections from the other ring before crashing again.

In #869 I'm already taking care of renaming the gossip-ring service in cortex to avoid any confusion.

### DIFF
--- a/packages/controller/src/resources/loki/index.ts
+++ b/packages/controller/src/resources/loki/index.ts
@@ -170,6 +170,8 @@ export function LokiResources(
     env: []
   };
 
+  const memberlist_bind_port = 7947;
+
   const lokiDefaultConfig = {
     server: {
       http_listen_port: 1080,
@@ -266,11 +268,11 @@ export function LokiResources(
       // don't allow split-brain / individual components that think they are
       // not part of a cluster.
       abort_if_cluster_join_fails: true,
-      bind_port: 7946,
+      bind_port: memberlist_bind_port,
       join_members: [
         // use a kubernetes headless service for all distributor, ingester and
         // querier components
-        `loki-gossip-ring.${namespace}.svc.cluster.local:7946`
+        `loki-gossip-ring.${namespace}.svc.cluster.local:${memberlist_bind_port}`
       ],
       // these settings are taken from examples in  official documentation
       // https://github.com/grafana/loki/blob/master/docs/sources/configuration/examples.md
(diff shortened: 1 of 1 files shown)


nickbp (approved):

nickbp:
I'll try manually changing these settings in the test cluster where the problem was encountered, and see if the cross-pollination goes away

nickbp:
Steps were:
- Manually edited `loki` configmap to have xxx7 port rather than xxx6 and removing `opstrace: owned` annotation` so it doesnt get reverted automatically
- Manually edited `loki-gossip-ring` service to have xxx7 port rather than xxx6 and removing `opstrace: owned` annotation` so it doesnt get reverted automatically
- Reenabling controller to deploy the above

Waiting for the deploy to roll out, will check in on it later

nickbp:
It looks like switching the loki port fixed cortex without losing anything. Pushes went from HTTP 500s to 200s once the cortex ingesters were no longer trying to talk to loki ingesters:
metrics-fixed

On loki's side of things, I was still seeing errors after the fix. I think this is because the cortex ingesters never "moved away" from their current endpoints so loki could still see them. For example I tried removing the cortex ingester entries from a loki distributor's `/ring` endpoint (over `port-forward 1080`) but they kept coming back. Next I tried just deleting and recreating the loki distributor Deployment and that did the trick, the loki distributors now only saw the loki ingesters:
logs-fixed

Summary: The fix was successfully applied to the cluster where the 'cross-pollination' problem was occurring. No data loss was observed either in the process of the change rolling out. Going to just merge now.

sreis:
Thanks @nickbp.

triclambert:
Nice teamwork, gents.  👏

spahl:
The things one discovers... Great find!

PR #1065, opened 2021-07-16
ci(upgrades): dissociate test-core

sreis:
~In upgrades pipeline, run test-core.sh in a container using CI docker image. This is required to fix the missing gcloud and aws binaries to refresh credentials.~

~Did some changes to make docker create the test artifact folders but need to check it still runs the PR pipeline.~

See https://github.com/opstrace/opstrace/pull/1065#issuecomment-881414748 bello

### DIFF
--- a/.buildkite/test-upgrade-pipeline.yml
+++ b/.buildkite/test-upgrade-pipeline.yml
@@ -1,3 +1,5 @@
+env:
+  CHECKOUT_VERSION_STRING: "${BUILDKITE_COMMIT:0:8}-ci"
 steps:
   - label: "🔨 preamble"
     key: "preamble"
--- a/ci/test-upgrade/create-cluster-artifacts.sh
+++ b/ci/test-upgrade/create-cluster-artifacts.sh
@@ -63,5 +63,10 @@ do
 done
 
 # make sure that looker-related artifacts are collected
-cp /build/looker*log /build/bk-artifacts
-cp /build/looker*report.json /build/bk-artifacts
+cp /build/looker*log /build/bk-artifacts || true
+cp /build/looker*report.json /build/bk-artifacts || true
+
+cp test-remote-artifacts/uishot-*.png /build/bk-artifacts || true
+
+mkdir -p /build/bk-artifacts/browser-test-result || true
+mv browser-test-results /build/bk-artifacts/ || true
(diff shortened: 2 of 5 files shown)


sreis:
Well, at least test-remote passed on both builds.

test-remote-ui failure:
```
Error: page.screenshot: EACCES: permission denied, open '/test-remote-artifacts/uishot-rootpage.png'
``` 

test-browser:

```
[2021-07-16T09:58:39Z] Running 90 tests using 1 worker
[2021-07-16T10:47:32Z] °····°············°××TT××TT××TT××TT××TT°×·×2021-07-16T10:47:32.735Z error: env variable OPSTRACE_CLOUD_PROVIDER must be set to `aws`, `gcp` or `dev`
[2021-07-16T11:11:12Z] T±··········×··T°×··T°×°T··×··×2021-07-16T11:11:12.463Z error: env variable OPSTRACE_CLOUD_PROVIDER must be set to `aws`, `gcp` or `dev`
[2021-07-16T11:14:46Z] T±····
```

The missing env vars error is weird.

Most of the errors seem to match up with the failures in the [upgrades pipeline for this branch](https://buildkite.com/opstrace/upgrade-tests/builds/1004). I call this progress.

sreis:
The deploy-testremote-teardown script (part of the PR pipeline) relies on docker-in-docker magic. For example, amongst other things, the PR pipeline moves the code checkout over to `/tmp` to facilitate passing docker volumes around when that script (running in a docker container) runs the test targets also in docker containers. 

None of the docker-in-docker stuff was required for the upgrade tests until we tried to consolidate how we run them in both pipelines.

For now, this commit reverts the code sharing between the CI pipelines and adds a script to set up the requirements for one of the tests.

I'll tackle the code-sharing between CI pipelines after we unblock the pending work and get a release out.

sreis:
Follow-up work is required to handle the loki empty memberlist ring after an upgrade but at least the pipeline is running the tests now.

jgehrcke:
Yeah right the CI container is mainly used for creating the _build environment_, where the outcome are the individual build artifacts (container images and CLI binary build). 

However, the `make ci -...` approach also brings the OPSTRACE_BUILD_DIR concept. Which is valuable so we have a systematic way for individual build steps in the same pipeline to _not_ operate in the same working directory.

sreis:
> However, the make ci -... approach also brings the OPSTRACE_BUILD_DIR concept. Which is valuable so we have a systematic way for individual build steps in the same pipeline to not operate in the same working directory.

Buildkite [checkouts in a new directory for every step](https://buildkite.com/docs/pipelines/defining-steps ).   

> When a step is run by an agent, it will be run with a clean checkout of the pipeline's repository. If your commands or scripts rely on the output from previous steps, you will need to either combine them into a single script or use artifacts to pass data between steps.

jgehrcke:
Thanks for quoting and linking the docs @sreis!

PR #1067, opened 2021-07-18
utils: when running local dev env use a default BUILD_INFO if there isn't a real one to use

terrcin:
Local dev for the `packages/app` stopped working with [this commit](https://github.com/opstrace/opstrace/commit/b13281f18da1). So as to not add further complexity to running the app locally I've updated `readBuildInfo` to use a default set of values instead of erroring when `process.env.NODE_ENV === "development"`.

### DIFF
--- a/lib/utils/src/buildinfo.ts
+++ b/lib/utils/src/buildinfo.ts
@@ -28,6 +28,7 @@
 // import chain)
 
 import fs from "fs";
+import { log } from "./log";
 
 export interface BUILD_INFO_TYPE {
   BRANCH_NAME: string;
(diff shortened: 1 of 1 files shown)


lib/utils/src/buildinfo.ts
```diff
@@ -101,10 +102,24 @@ function readBuildInfo() {
 
   // None of the candidate file paths led to success. Fall back to env.
   if (!process.env.OPSTRACE_BUILDINFO_PATH) {
-    throw Error(
-      `environment variable OPSTRACE_BUILDINFO_PATH not set ` +
-        `and none of [${JSON.stringify(CANDIDATE_PATHS)}] found/readable`
-    );
+    if (process.env.NODE_ENV === "development") {
+      log.info(
+        "using hardcoded 'local-dev' for BUILD_INFO, to override set environment variable OPSTRACE_BUILDINFO_PATH"
```
jgehrcke, line 107: >  override set environment variable OPSTRACE_BUILDINFO_PATH

Hm. I don't think this is logically correct.

It's more of a "seen NODE_ENV === "development", so do not require OPSTRACE_BUILDINFO_PATH to be set"

More importantly, in my experimentation with this approach I deliberately did _not_ import `log` because I ran into a circular import -- buildinfo.ts as well as log.ts are candidates for being imported super early.

jgehrcke:
> So as to not add further complexity to running the app locally

1) Call `make set-build-info-constants`. (We used that target before). That writes `buildinfo.json` to the root directory of the repository.
2) Set the environment variable `OPSTRACE_BUILDINFO_PATH` to point to that buildinfo.json file.

I can understand that you didn't want to bother with that, but if we can add these two steps to the scripts/make target for "running the app locally" the we have 'proper' build info even for local dev.

PR #1073, opened 2021-07-20
ci: run upgrade tests on GCP as part of PR pipeline

sreis:
As discussed in slack:

> I propose running the upgrades pipeline (for GCP only) in all PRs for the next two weeks. This way, we can quantify how much it affects developer productivity.

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -58,6 +58,25 @@ steps:
       - bash ci/check-if-docs-pr.sh || make ci-deploy-testremote-teardown
     artifact_paths:
       - "/tmp/opstrace-bk-${BUILDKITE_BUILD_NUMBER}-${BUILDKITE_COMMIT:0:7}-1-aws/bk-artifacts/**/*"
+  - label: "🔨 test upgrades (GCP)"
+    key: "testupgrades-gcp"
+    depends_on:
+      - "unit-tests"
+    env:
+      # Fetch the latest Opstrace CLI artifact from S3 and use that as the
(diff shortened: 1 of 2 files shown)


.buildkite/pipeline.yml
```diff
@@ -58,6 +58,25 @@ steps:
...
+      # cluster.
+      OPSTRACE_CLI_VERSION_TO: ${OPSTRACE_PREBUILD_DIR}/build/bin/opstrace
+      OPSTRACE_CLUSTER_NAME: "pr-upgr-bk-${BUILDKITE_BUILD_NUMBER}-${BUILDKITE_COMMIT:0:3}-g"
+      OPSTRACE_CLOUD_PROVIDER: "gcp"
+      GCLOUD_CLI_REGION: "us-west2"
+      GCLOUD_CLI_ZONE: "us-west2-a"
+    command:
+      - ci/test-upgrade/run.sh
+    artifact_paths:
+      - "/tmp/opstrace-bk-${BUILDKITE_BUILD_NUMBER}-${BUILDKITE_COMMIT:0:7}-1-gcp/bk-artifacts/**/*"
```
jgehrcke, line 79: This is the same as for `maintest-gcp` above -- that's probably an oversight, right? 

Which directory is this executed in? That is, what would `pwd` show in `run.sh`?

Let's make sure this build step has its own working directory (that no other things interfere with) and then let's create a `bk-artifacts` directory in there, and then collect that.

sreis: Yeah. I'll push a fix soon.

PR #1077, opened 2021-07-21
ui: add dashboard for pod metrics in K8s metrics integration

nickbp:
~NOTE: Haven't tested this Dashboard JSON content yet. It was built/exported from a Grafana instance and then edited in-place.~ (this has been done - works fine)

NEEDS FEEDBACK: This adds support for updating/resetting existing integration dashboards, with a "Reinstall Dashboards" button. This avoids trying to detect changes between the dashboards that are currently deployed vs what the App has available, which subjectively felt reasonable enough. Our call to Grafana enables a "overwrite" flag so as long as the new dashboards didn't remove a dashboard vs the old dashboards, we should be able to just overwrite in-place.
FWIW, another option for updating dashboards would be for the user to manually delete the integration folder, at which point under the current logic the "Install Dashboards" button should become available on its own. But this isn't super intuitive.

### DIFF
--- a/packages/app/src/client/integrations/k8sLogs/Show/InstallInstructions.tsx
+++ b/packages/app/src/client/integrations/k8sLogs/Show/InstallInstructions.tsx
@@ -88,7 +88,23 @@ export const InstallInstructions = ({
     saveAs(configBlob, configFilename);
   };
 
+  const dashboardButtonText = useMemo(
+    () =>
+      isDashboardInstalled ? "Reinstall Dashboards" : "Install Dashboards",
+    [isDashboardInstalled]
+  );
+
   const dashboardHandler = async () => {
+    // Delete existing folder, if any. For reinstalling/updating dashboards.
+    try {
+      await grafana.deleteFolder({ integration, tenant });
+    } catch (err) {
+      // Ignore 404 error - expected for initial dashboard install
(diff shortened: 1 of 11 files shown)


nickbp:
Planning on merging once CI's green. The button itself largely works for getting the dashboards updated/refreshed. The new dashboard being provided looks good now as well (checked how it looked against both an EKS cluster and an onprem/K3s cluster).

nickbp:
Quick rebase/push to kick CI

MoSattler (commented):
Hi @nickbp!

Just wanted to let you know that we're currently working on adding error handling with somewhat usable feedback for the user. Right now we're aiming to do that with [useNotificationService](https://github.com/opstrace/opstrace/blob/03f91eb8a57ef9b10f152d87a070df97a64d2b30/packages/app/src/client/services/Notification/NotificationService.tsx#L76). 

As you add some code here that makes some requests, may you consider adding some basic error handling with `useNotificationService` in places where it makes sense?

nickbp:
I was able to switch over to using `useNotificationService` for the places where RPCs were being made. I manually ran the new code and didn't see any issues in the happy path of installing a new integration, reinstalling the dashboard, and uninstalling the integration - however I didn't have a way of forcing the sad path.

PR #1088, opened 2021-07-22
ui(uninstall-integration-button): add error handling

MoSattler:
Adding user feedback for the uninstallation button for integrations, in case there are server or network problems. 

Screenshot 2021-07-23 at 14 55 00

# Have you...

### DIFF
--- a/packages/app/package.json
+++ b/packages/app/package.json
@@ -219,6 +219,7 @@
     "mini-css-extract-plugin": "0.9.0",
     "mini-svg-data-uri": "^1.2.3",
     "module-alias": "^2.2.2",
+    "msw": "^0.32.2",
     "nock": "^13.0.11",
     "nodemon": "^2.0.4",
     "nodemon-webpack-plugin": "^4.3.2",
--- a/packages/app/src/client/integrations/common/UninstallIntegrationBtn.tsx
+++ b/packages/app/src/client/integrations/common/UninstallIntegrationBtn.tsx
@@ -30,6 +30,8 @@ import graphqlClient from "state/clients/graphqlClient";
 import { deleteFolder } from "client/utils/grafana";
 
 import { Button } from "client/components/Button";
+import { useNotificationService } from "client/services/Notification";
+import { random } from "lodash";
 
 export const UninstallBtn = ({
   integration,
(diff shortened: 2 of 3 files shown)


packages/app/package.json
```diff
@@ -218,6 +218,7 @@
     "mini-css-extract-plugin": "0.9.0",
     "mini-svg-data-uri": "^1.2.3",
     "module-alias": "^2.2.2",
+    "msw": "^0.32.2",
```
MoSattler, line 221: Adding [msw](https://mswjs.io) for rest and gql mocking

packages/app/src/client/integrations/common/UninstallIntegrationBtn.tsx
```diff
@@ -44,28 +46,54 @@ export const UninstallBtn = ({
...
+          state: "error" as const,
+          title: "Could not delete grafana folder",
+          information: error.response.data.message ?? error.message,
+          handleClose: () =>
+            unregisterNotification({
+              id: messageId,
+              title: "",
+              information: ""
+            })
+        };
```
MoSattler, line 77: I believe there might be a case where an error would be expected: when a folder does not exist. 
Should be prevent user feedback being shown in this case?

jgehrcke: General guideline: distinguish retryable from non-retryable errors.

- Retryable errors should automatically lead to retrying, but not w/o indicating this to the user (spinning wheel, whatever). Retrying needs to hit a "giving up" criterion, in which case the _last_ retryable error detail should be exposed to the user.

- For non-retryable errors the assumption is that the detail should always be exposed to the user -- often the user can do something about them. Either by changing the input or by 'talk to an admin'.
 
- > when a folder does not exist. ... Should be prevent user feedback being shown in this case?

That's a beautiful example for non-retryable error, and the user can do something about it: _definitely_ show the feedback to the user :)


---------

I have not really looked at all this here -- but my suspicion is that may expose retryable erros to the user w/o actually retrying automatically before.

jgehrcke: And by the way: _thanks_ for diving into the error handling regime. Desperately needed, and one of my favorite topics -- often underestimated.

MatApple: I agree with @jgehrcke on the above but would add a caveat which is that retryable error handling gets complicated quite quickly in a UI, where you have to track state across retries, choose a good criteria to stop retrying, and communicate status to a user during this process. While I believe we can aim for that in the longer term, I'd prefer to start simple and iterate from there. Just showing an immediate error (whether it's retryable or not) to a user is a really good start and a very impactful improvement on the current UX. From there, adding a suggested action (which perhaps just says "retry again" if it's retryable) would be a good next step. The great thing about UI is that a user can just click a button a second time to retry so we can get quite far without building full retry logic (for now).

jgehrcke: @MatApple yes for sure. I introduced with "General guideline" and also said "I have not really looked at all this here" but I failed to state explicitly that I like landing any improvement / intermediate state towards that north star that I tried to describe :)

> Just showing an immediate error (whether it's retryable or not) to a user is a really good start and a very impactful improvement on the current UX

So: fully agree!

PR #1089, opened 2021-07-22
touch up some docs

triclambert:


.buildkite/pipeline.yml
```diff
@@ -74,7 +74,7 @@ steps:
       GCLOUD_CLI_REGION: "us-west2"
       GCLOUD_CLI_ZONE: "us-west2-a"
     command:
-      - ci/test-upgrade/run.sh
+      - bash ci/check-if-docs-pr.sh || ci/test-upgrade/run.sh
```
triclambert, line 77: Changing this means that this is no longer a docs PR.

triclambert:
Since the 3 builds have all begun, I will call my latest change to pipeline.yml successful and merge.

PR #1111, opened 2021-07-26
feat: update configuring alerts doc

triclambert:


triclambert:
Note:  we need to work on a better site preview mechanism, but for now, here is a rendered version:

image

MatApple:
Awesome, thanks for the update! 🎉

PR #1100, opened 2021-07-26
[wip] ui: upgraded to latest Playwright - v1.13.0

terrcin:

PR #1092, opened 2021-07-23
[wip] [poc] ui:don't run mobile browser tests

terrcin:
PR is testing out ideas from this issue: #1087

PR #1076, opened 2021-07-21
[WIP] ui: white screen of auth

terrcin:
todo: description

PR #1115, opened 2021-07-27
[wip] ui: white screen of auth - rolling back to getting user email from auth0 callback to browser

terrcin:
This is the [white screen of auth](https://github.com/opstrace/opstrace/pull/1076) PR with 8db043e07c4fb885b9980b2709560fcbdef7ac09 manually rolled back and https://github.com/opstrace/opstrace/pull/1114, https://github.com/opstrace/opstrace/pull/1100, and https://github.com/opstrace/opstrace/pull/1092 merged in.

There is a concern that with the CI logging in before each test is causing Auth0 rate limit issues due to changing session creation to calling the Auth0 API. This PR is about minimising that and stablising the `white screen of auth` change so that it can get merged.


jgehrcke:
> this is causing Auth0 rate limit issues 

Do / did we have indications for actual rate limiting?

jgehrcke:
I am curious about what our high-level problem description is about the "white screen" -- what went wrong in these cases, and which part of this larger patch is what we think the critical piece to consolidate that?

terrcin:
> Do / did we have indications for actual rate limiting?

When I added the Auth0 api `/userinfo` call coming from the express server the browser tests started behaving extremely slowing and failing alot. We were not exposing the details of these errors in the react app, just simply showing that an error occured or possibly presenting a white screen to the user as the react state isn't changed once a white screen is shown. When I tried running browser tests locally while I had a couple of PRs running on CI, also all at the `test-browser` stage, I got the rate limit error from Auth0. I waited a bit then ran them locally again without issue. 

Regarding the actual change here. The aim of this PR has been to simplify the code and flow around the user login process - reducing uncertainty about the current state and removing unneccessary code and dependencies on other steps having happened in parallel. I was expecting the problem to be a race condition as being able to login seemed to rely on dependent parts hopefully occuring in the correct order rather than a series of deliberate sequential steps and checks. Most notibly relying on `useCurrentUser` looked problematic to me. There were also a couple of catches that simply swallowed exceptions that could also have been causing users to get stuck on a white screen.

The new `withSession` component directly checks with the express server if there is a logged in user or not. If not withSession will try and create a new Opstrace session if the surfer has already authenticated with Auth0 - if they haven't they get presented with a Login button that sends them off to Auth0, when they come back the process starts again. The most fiddly bit was the `returnTo` behavior.

terrcin:
Also, I've intentionally left in the old `withAuthentication` component and `login` view to make it super easy to switch back to if neccessary. I'll create a PR in a week or two once `withSession` has had some real world usage.

jgehrcke:
> Regarding the actual change here. The aim of this PR has been to simplify the code and flow around the user login process - reducing uncertainty about the current state and removing unneccessary code and dependencies on other steps having happened in parallel. I was expecting the problem to be a race condition as being able to login seemed to rely on dependent parts hopefully occuring in the correct order rather than a series of deliberate sequential steps and checks. Most notibly relying on useCurrentUser looked problematic to me. There were also a couple of catches that simply swallowed exceptions that could also have been causing users to get stuck on a white screen.

Thanks for that explanation. Makes a lot of sense!

packages/app/src/client/components/withAuthentication/index.tsx
```diff
@@ -14,15 +14,17 @@
...
+
 import useCurrentUser, {
   useCurrentUserLoaded
 } from "state/user/hooks/useCurrentUser";
 import { setCurrentUser } from "state/user/actions";
+
 /**
- * withAccess wraps a component to ensure the component
+ * WithAuthentication wraps a component to ensure the component
  * is only mounted if access has been granted.
```
jgehrcke, line 28: > if access has been granted

Then let's call it `WithAuthorization`.

jgehrcke:
https://auth0.com/docs/policies/rate-limit-policy/authentication-api-endpoint-rate-limits#limits-for-production-tenants-of-paying-customers

That is, for us `/userinfo` is limtied by `User ID` to `5 requests per minute with bursts up to 10 requests`.

That also means that when we hit this in CI, it's tied to the account used in CI.

Issue #1102, opened 2021-07-26
ci instability: End-to-end alerts for ...: Expectation not fulfilled within 300 s


jgehrcke:
**NOTE, August 2**: the original main symptom tracked here was `Expectation not fulfilled within 30 s`.

Since then, code had changed, and a _different_ instability symptom (not unlikely as of the same root cause) popped up, showing as `Expectation not fulfilled within 300 s`. Usually we should have created a new ticket for the new symptom, but now it's ok to track the new symptom here. I've updated the title.

------------

```
[2021-07-26T01:09:32Z]   21 passing (26m)
[2021-07-26T01:09:32Z]   4 pending
[2021-07-26T01:09:32Z]   2 failing
[2021-07-26T01:09:32Z]
[2021-07-26T01:09:32Z]   1) End-to-end alert tests
[2021-07-26T01:09:32Z]        End-to-end alerts for default tenant:
[2021-07-26T01:09:32Z]      Error: Expectation not fulfilled within 30 s
[2021-07-26T01:09:32Z]       at Object.waitForQueryResult (testutils/index.ts:931:9)
[2021-07-26T01:09:32Z]       at processTicksAndRejections (node:internal/process/task_queues:96:5)
[2021-07-26T01:09:32Z]       at getE2EAlertCountMetric (test_alerts.ts:301:23)
[2021-07-26T01:09:32Z]       at testE2EAlertsForTenant (test_alerts.ts:365:17)
[2021-07-26T01:09:32Z]       at Context.<anonymous> (test_alerts.ts:422:5)
```

```
[2021-07-26T00:45:15Z] 2021-07-26T00:45:15.952Z info: Waiting for cortex E2E alerting metric for tenant=default job=testalerts-pqoch-job with zero value
[2021-07-26T00:45:15Z] 2021-07-26T00:45:15.952Z info: Cortex query parameter (object):
[2021-07-26T00:45:15Z] {
[2021-07-26T00:45:15Z]   "query": "e2ealerting_webhook_receiver_evaluations_total{job=\"testalerts-pqoch-job\"}"
[2021-07-26T00:45:15Z] }
[2021-07-26T00:45:15Z] 2021-07-26T00:45:15.952Z info: Cortex query parameters (query string):
[2021-07-26T00:45:15Z] query=e2ealerting_webhook_receiver_evaluations_total%7Bjob%3D%22testalerts-pqoch-job%22%7D
[...]

PR #1122, opened 2021-07-27
test-remote: test_alerts: bump timeout for waitForCortexMetricResult()

jgehrcke:
CC @nickbp 

motivated by https://github.com/opstrace/opstrace/issues/1102#issuecomment-887405723

It's unclear why certain HTTP requests are not responded to in a timely fashion (below 10 s). This is fishy, we should investigate.


jgehrcke:
The patch was covered, we have green CI signal alraedy -- merging. Then starting a small number of builds from main to assess current stability.

PR #1126, opened 2021-07-27
ui: remove CondRender

MoSattler:
Proposal: removing `<CondRender />` in favour of a pure JS(X) solution for conditional rendering.

For two reasons I think that this would be advantageous:
1. `<CondRender />` is [not super straightforward](https://github.com/opstrace/opstrace/blob/main/packages/app/src/client/utils/rendering.tsx#L33-L47).
1. Pure JS is react's recommended way for conditional rendering, see the [docs](https://reactjs.org/docs/conditional-rendering.html#inline-if-with-logical--operator).

# Have you...


jgehrcke (approved):
Thanks for those two reasons! :) From my pov we can merge this because you explained this so nicely. However, I don't know anything about react and maybe @mat and @terrcin can quickly leave feedback.

MatApple:
I agree with the direction to remove <CondRender /> in favour of a pure JS(X) solution for conditional rendering. It does add unnecessary complexity through a layer of indirection. Tagging @terrcin in here to comment since he had the vision for this component.

terrcin:
I was trying to re-create CondRender from memory based on it from my previous job. There it was more signifigant and featurefull rather than just simple boolean checking. It's never gone anywhere after the initial version of it was comitted so happy to drop the idea. :-)

terrcin (approved):

PR #1139, opened 2021-07-29
ui: turn js files to ts

MoSattler:
- Turning some JS files into TS

There might have been a reason why these files remain JS - so please let me know if this would be a mistake!

# Have you...


MatApple (approved):
LGTM!

terrcin (approved):
LGTM 2

I've checked out the branch and run it via telepresences and can verify that I can still correctly create grafana dashboards for i9n metrics and logs.

MoSattler:
@terrcin thank you for making sure it's all properly working!

PR #1140, opened 2021-07-29
chore: add loki query-frontend

sreis:
* Applies some config options from [loki production settings](https://github.com/grafana/loki/blob/main/production/ksonnet/loki/config.libsonnet).

Close https://github.com/opstrace/private/issues/35.

### DIFF
--- a/packages/controller/src/resources/loki/index.ts
+++ b/packages/controller/src/resources/loki/index.ts
@@ -30,7 +30,13 @@ import {
   ServiceAccount
 } from "@opstrace/kubernetes";
 import { State } from "../../reducer";
-import { roundDownToOdd, select, getBucketName } from "@opstrace/utils";
+import {
+  roundDownToOdd,
+  select,
+  getBucketName,
+  min,
+  roundDown
+} from "@opstrace/utils";
 import {
   deepMerge,
   getControllerConfig,
@@ -46,9 +52,14 @@ export function LokiResources(
   namespace: string
 ): ResourceCollection {
   const collection = new ResourceCollection();
-  const { infrastructureName, target, region, gcp } = getControllerConfig(
-    state
-  );
+  const {
+    infrastructureName,
+    target,
+    region,
+    gcp,
+    logRetentionDays
+  } = getControllerConfig(state);
+
   const dataBucketName = getBucketName({
     clusterName: infrastructureName,
     suffix: "loki"
@@ -167,16 +178,34 @@ export function LokiResources(
         }
       ])
     },
+    queryFrontend: {
+      // more than 2 is not beneficial (less advantage of queueing requests in
+      // front of queriers)
+      replicas: 2
+    },
+
+    memcachedResults: {
+      replicas: select(getNodeCount(state), [
+        { "<=": 4, choose: 2 },
+        { "<=": 9, choose: 3 },
+        {
+          "<=": Infinity,
+          choose: min(4, roundDown(getNodeCount(state) / 3))
+        }
+      ]),
+      resources: {}
+    },
     env: []
   };
 
   const memberlist_bind_port = 7947;
+  const grpc_server_max_msg_size = 41943040; // default (4 MB) * 10
 
   const lokiDefaultConfig = {
     server: {
       http_listen_port: 1080,
-      grpc_server_max_recv_msg_size: 41943040, // default (4 MB) * 10
-      grpc_server_max_send_msg_size: 41943040 // default (4 MB) * 10
+      grpc_server_max_recv_msg_size: grpc_server_max_msg_size,
+      grpc_server_max_send_msg_size: grpc_server_max_msg_size
     },
     auth_enabled: true,
     chunk_store_config: {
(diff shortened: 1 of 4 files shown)


packages/controller/src/resources/loki/index.ts
```diff
@@ -196,22 +225,56 @@ export function LokiResources(
       ingestion_rate_mb: 500,
       ingestion_burst_size_mb: 2000,
       // enforced in querier
-      max_entries_limit_per_query: 60000 // Per-tenant entry limit per query
+      max_entries_limit_per_query: 60000, // Per-tenant entry limit per query
+      // https://github.com/grafana/loki/blob/main/production/ksonnet/loki/config.libsonnet#L190-L198
+      reject_old_samples: true,
```
sreis, line 230: Forgot to link this article here
https://grafana.com/docs/loki/latest/design-documents/2021-01-ordering-constraint-removal/#flushing--chunk-creation

jgehrcke, line 230: I see that this is a significant change, and I have also had a peek into the doc you linked. It's a really important design change in Loki that's being discussed there! 

For the record, this is about changing `reject_old_samples` from `false` (default) to `true`:
```
# Whether or not old samples will be rejected.
# CLI flag: -validation.reject-old-samples
[reject_old_samples: <bool> | default = false]
```

Which means that one cannot add log streams to Loki anymore that start arbitrarily long ago in the past.

Two questions:
1) can we maybe make this change in a separate PR / effort? (is this somewhat tied to the introduction to the query frontend?

2) After reading https://grafana.com/docs/loki/latest/design-documents/2021-01-ordering-constraint-removal/#flushing--chunk-creation I am unclear how far the implementation already goes -- is the Loki that we have currently in Opstrace already set up for out-of-order writes (within a single stream)?

My general thinking is that once the Loki in our system _allows for_ out-of-order writes to a single log stream I'd love to reflect this in our test suite very explicitly.

sreis: > After reading https://grafana.com/docs/loki/latest/design-documents/2021-01-ordering-constraint-removal/#flushing--chunk-creation I am unclear how far the implementation already goes -- is the Loki that we have currently in Opstrace already set up for out-of-order writes (within a single stream)?

> My general thinking is that once the Loki in our system allows for out-of-order writes to a single log stream I'd love to reflect this in our test suite very explicitly.

So, [out of order logs do not seem to be supported upstream](https://github.com/grafana/loki/issues/1544#issuecomment-575772520).  

Then there's this interesting bit in the [blog post](https://grafana.com/docs/loki/latest/design-documents/2021-01-ordering-constraint-removal/#flushing--chunk-creation):

> Loki regularly combines multiple blocks into a chunk and "flushes" it to storage. In order to ensure that reads over flushed chunks remain as performant as possible, we will re-order a possibly-overlapping set of blocks into a set of blocks that maintain monotonically increasing order between them. 

With this in mind, I don't think it makes sense to accept logs with very old timestamps since chunks try to maintain monotonically increasing order for performance reasons.

Another issue I see is accepting data with a timestamp that conflicts with the log retention settings. Genuinely curious about what happens to data that is ingested with timestamps older than the log retention policy. Will it only be available while it's in ingester memory? I assume that as soon as it is flushed to storage and the compactor runs, it marks it for deletion. The test suite we have does not account for how the other components interact. Not saying it should, but it might lead to test things that Loki is not meant to do.

> could we maybe make this change in a separate PR / effort? (is this somewhat tied to the introduction to the query frontend)?

This PR is based on a branch from a few months ago to introduce the query frontend and update the Loki settings to follow what upstream recommends for production deployments. In this case, we should be conservative and follow Loki production settings, and reject samples that might cause an issue with chunk generation. @jgehrcke wdyt?

jgehrcke: Based on your reply I think I understand that the change to `reject_old_sample` _is_ unrelated to the introduction of the query frontend. 

I think I also understand that you would like to keep it in here.

That's fair.

It's just an important behavioral change that should not be silently piggy-backed onto an 'unrelated' PR, and I think now that we have discussed this here that is already not silent anymore.

> Genuinely curious about what happens to data that is ingested with timestamps older than the log retention policy

I think you bring up a great point about the undefined behavior / inconsistent behavior here. In CI, I think the system is configured to have a retention period of 7 days for logs. Being able to insert log data older than that is simply flawed. I never thought about it that way before, bit it's pretty obvious. To me, this is as of now the best / _strongest_ reason to change the test suite to not used fixed timestamp log entries anymore far from the past.

test/test-remote/test_loki_log_ingest_api.ts
```diff
@@ -150,7 +151,7 @@ suite("Loki API test suite", function () {
   });
 
   test("insert w/ cntnrzd FluentD(loki plugin), then query", async function () {
-    const sampleTimeRFC3339nano = "2010-10-10T10:10:01.123456789Z";
+    const now = timestampToRFC3339Nano(ZonedDateTime.now(ZoneId.UTC));
```
jgehrcke, line 154: update: see discussion above. This change makes sense because we do not want to test inserting data older than the retention period.

test/test-remote/test_loki_log_ingest_api.ts
```diff
@@ -252,9 +253,7 @@ suite("Loki API test suite", function () {
     // @ts-ignore: TS2532: Object is possibly 'undefined'.
     const testname = testName(this);
 
-    // `sampletsns` is for example: "1286705401123456789"
-    const sampleTimeRFC3339nano = "2010-10-10T10:10:01.123456789Z";
-    const ts = ZonedDateTime.parse(sampleTimeRFC3339nano);
+    const ts = ZonedDateTime.now().minusMinutes(10);
```
jgehrcke, line 256: update: see discussion above. This change makes sense because we do not want to test inserting data older than the retention period.

test/test-remote/test_loki_log_ingest_api.ts
```diff
@@ -322,8 +322,8 @@ suite("Loki API test suite", function () {
             uniq: rndstring().slice(0, 5)
           },
           values: [
-            ["1286705401123456789", "aaa"],
-            ["1286705401123456789", "aaa"]
+            [now, "aaa"],
+            [now, "aaa"]
```
jgehrcke, line 326: update: see discussion above. This change makes sense because we do not want to test inserting data older than the retention period.

test/test-remote/test_loki_log_ingest_api.ts
```diff
@@ -352,8 +352,7 @@ suite("Loki API test suite", function () {
     const testname = testName(this);
 
     // Specify details of log record to be inserted.
-    const sampleTimeRFC3339nano = "2012-10-10T10:10:01.123456789Z";
-    const sampletimestamp = ZonedDateTime.parse(sampleTimeRFC3339nano);
+    const sampletimestamp = ZonedDateTime.now().minusMinutes(10);
```
jgehrcke, line 355: update: see discussion above. This change makes sense because we do not want to test inserting data older than the retention period.

jgehrcke (approved):
:rocket:

jgehrcke:
Let's finally land this. Thanks, @sreis!

PR #1141, opened 2021-07-30
ui: Add radio option for log format to K8s log integration

nickbp:
Adds an option for specifying whether the cluster is running CRI/containerd or dockerd. This is required by the underlying log scraper - depending on the runtime, the paths for the data are different and they have totally different formats. Previously, the code was hardcoded to CRI/containerd even though the underlying yaml template had an option for dockerd. There was just a TODO for exposing this as an option.

The result looks like this. Could probably be cleaned up a bit but I figured that this should be okay for now (baby's first react UI):
k8slogs-radio-help

### DIFF
--- a/packages/app/src/client/integrations/k8sLogs/Form.tsx
+++ b/packages/app/src/client/integrations/k8sLogs/Form.tsx
@@ -106,11 +106,13 @@ export const K8sLogsForm = ({ handleCreate }: Props) => {
                     {label:"dockerd", value:"docker"}
                   ]}
                   label="Container Runtime"
-                  helperText="The log format produced by the cluster depends on the runtime"
+                  helperText="The cluster container runtime affects the log format."
                 />
-                <Typography color="textSecondary" variant="subtitle1">Check the runtime using this command:</Typography>
-                <Typography color="textSecondary" variant="subtitle1">{"kubectl get nodes -o jsonpath='{.items[].status.nodeInfo.containerRuntimeVersion}'"}</Typography>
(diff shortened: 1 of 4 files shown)


nickbp:
@terrcin LMK if you have any pointers for handling the margins better in the help text. Is there a component or option for doing this? I wanted the output to look like the above screenshot, but I imagine the underlying `&nbsp;` hack isn't the correct way to accomplish this.

packages/app/src/client/integrations/k8sLogs/Form.tsx
```diff
@@ -22,24 +22,29 @@ import * as yup from "yup";
...
+  deployNamespace: yup.string().required(),
+  format: yup.string().required()
 });
 
 const defaultValues: Values = {
   name: "",
-  deployNamespace: "opstrace"
+  deployNamespace: "opstrace",
+  format: "cri"
 };
```
MoSattler, line 48: You can DRY this a bit up my deriving `Values` from `yup`.

```ts
const Schema = yup.object({
  name: yup.string().required(),
  deployNamespace: yup.string().required(),
  format: yup.string().required()
});

type Values = yup.TypeOf<typeof Schema>
```

MoSattler (commented):
Some typing suggestions!

nickbp:
Thanks! I've added the typing improvements and will check it once more before merging (depending on how CI goes)

nickbp:
GCP failed on `view_ring_health` in test-remote-ui, known-flaky.

Tested the UI changes by hand, and things worked as expected - the radio option worked as expected, the UI layout looked fine, and CRI and dockerd-format yaml configs were rendered correctly

PR #1143, opened 2021-07-30
ui: deduplication of schema validation

MoSattler:
Deduplication by depriving types from yup schema instead of typing it manually.

# Have you...

### DIFF
--- a/packages/app/src/client/integrations/exporterAzure/Form.tsx
+++ b/packages/app/src/client/integrations/exporterAzure/Form.tsx
@@ -29,16 +29,7 @@ import { Button } from "client/components/Button";
 import { Typography } from "client/components/Typography";
 import { ExternalLink } from "client/components/Link";
 
-type Values = {
-  name: string;
-  subscriptionId: string;
-  azureTenantId: string;
-  clientId: string;
-  clientSecret: string;
-  config: string;
-};
-
-const Schema = yup.object().shape({
+const Schema = yup.object({
   name: yup.string().required(),
   subscriptionId: yup.string().required(),
   azureTenantId: yup.string().required(),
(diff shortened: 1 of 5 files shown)


jgehrcke:
Just an anecdote: personally, I have sometimes found yup-derived types to become reaaaally complex. Resulting tsc-generated error messages were then extremely hard to read. So I remember having made the opposite change: from yup-derived types to explicit / manually prepared typescript types, to keep compiler errors understandable. Also, I had kind of sworn to myself to try to fade yup out of our code based during those two times when I tried to update it from the version we currently use, to the most recent release. Twice, I had to back out of that because it became intolerably painful.

YMMV and I didn't look at details here :).

MatApple:
The number one benefit of Yup is to have runtime validation. I agree that we shouldn't use Yup where we don't gain anything from runtime validation - i.e. where we feel very confident in the data structure and type that we're passing. Having a bit of runtime validation upfront on unknown data can prevent a hornet's nest of hard-to-track-down problems during runtime, where the JS code fails because it received a type that is different than what is expected (i.e. what the developer _said the typings_ **should be**). Or worse, it doesn't fail explicitly but succeeds and leaves some downstream part of the system to fail with no breadcrumbs back to the origin of the problem

MatApple:
Perhaps there is a better way to use Yup and typescript typings such that we don't have this pain @jgehrcke and we still get the benefit of runtime validation?

jgehrcke:
> The number one benefit of Yup is to have runtime validation.

Ack.

> Perhaps there is a better way to use Yup and typescript typings such that we don't have this pain @jgehrcke and we still get the benefit of runtime validation?

Well, it's really all about dropping the yup-derived type kind of as soon as it makes sense.
1) do the validation
2) For the result, though, do a type-coercion into a "simple / vanilla" TypeScript type.

Done for example here: https://github.com/opstrace/opstrace/blob/255c79563fb760ed4b8b1c564fe99bfcc62866fb/packages/cli/src/create.ts#L109

The right-hand side of the expression are all yup-derived types. Left-hand side is a vanilla TS type. Further down the code path from here, the TSC error messages will not contain yup-specific thingies.

Issue #47, opened 2020-12-01
uninstaller: gcp: does not recover from network resource 'projects/vast-pad-snip/global/networks/snip is already being used by 'projects/vast-pad-snip/global/firewalls/k8s-snip-node-hc'


jgehrcke:
Consumed all high-level retries, didn't break the dependency cycle:
```
...
2020-12-01T11:07:08.623Z info: Destroying VPC
2020-12-01T11:07:30.531Z info: VPC deletion has started with status: RUNNING
2020-12-01T11:07:37.189Z error: error during cluster teardown (attempt 5):
ApiError: The network resource 'projects/vast-pad-240918/global/networks/jpdev' is already being used by 'projects/vast-pad-<snip>/global/firewalls/k8s-3cc02042af359f14-node-hc'
    at new ApiError (/home/jp/dev/opstrace/node_modules/@google-cloud/common/build/src/util.js:59:15)
    at Util.parseHttpRespBody (/home/jp/dev/opstrace/node_modules/@google-cloud/common/build/src/util.js:194:38)
    at /home/jp/dev/opstrace/node_modules/@google-cloud/compute/src/operation.js:251:46
    at /home/jp/dev/opstrace/node_modules/@google-cloud/compute/src/operation.js:234:7
    at /home/jp/dev/opstrace/node_modules/@google-cloud/common/build/src/service-object.js:193:13
    at /home/jp/dev/opstrace/node_modules/@google-cloud/common/build/src/util.js:369:25
    at Util.handleResp (/home/jp/dev/opstrace/node_modules/@google-cloud/common/build/src/util.js:145:9)
    at /home/jp/dev/opstrace/node_modules/@google-cloud/common/build/src/util.js:434:22
    at onResponse (/home/jp/dev/opstrace/node_modules/retry-request/index.js:214:7)
    at /home/jp/dev/opstrace/node_modules/teeny-request/src/index.ts:325:11 {
  code: undefined,
  errors: [Array],
  response: undefined
}
2020-12-01T11:07:37.190Z error: JSON representation of err: {
  "errors": [
    {
      "code": "RESOURCE_IN_USE_BY_ANOTHER_RESOURCE",
      "message": "The network resource 'projects/vast-pad-240918/global/networks/jpdev' is already being used by 'projects/vast-pad-<snip>/global/firewalls/k8s-3cc02042af359f14-node-hc'"
    }
  ],
[...]

PR #1144, opened 2021-07-30
go: proxy: replace X-Scope-OrgID header

jgehrcke:
https://pkg.go.dev/net/http#Header.Add
https://pkg.go.dev/net/http#Header.Set

> Set sets the header entries associated with key to the single element value. It replaces any existing values associated with key.

We want to take full control in the proxy, and set the header based on the value provided by the authenticator. Any value set by the original HTTP client should be obliterated. Therefore, use `Set()`.

Exploiting this for cross-tenant data writing would still require valid authentication proof for a tenant.

### DIFF
--- a/go/pkg/middleware/proxy.go
+++ b/go/pkg/middleware/proxy.go
@@ -151,7 +151,7 @@ func (trp *TenantReverseProxy) HandleWithProxy(w http.ResponseWriter, r *http.Re
 	}
 
 	// Add the tenant in the request header and then forward the request to the backend.
-	r.Header.Add(trp.headerName, tenantName)
+	r.Header.Set(trp.headerName, tenantName)
 	trp.Revproxy.ServeHTTP(w, r)
 }
 


sreis (commented):
Should we update the other calls too?

```
> git grep Header.Add
go/pkg/ddapi/handler.go:241:    req.Header.Add("X-Prometheus-Remote-Write-Version", "0.1.0")
go/pkg/ddapi/handler.go:242:    req.Header.Add("Content-Encoding", "snappy")
go/pkg/graphql/access.go:41:            req.Header.Add("x-hasura-admin-secret", g.secret)
go/pkg/middleware/proxy.go:154: r.Header.Add(trp.headerName, tenantName)
```

jgehrcke:
> Should we update the other calls too?

I looked at them too and found the other three 'fine', from a security point of view:

```
> git grep Header.Add
go/pkg/ddapi/handler.go:241:    req.Header.Add("X-Prometheus-Remote-Write-Version", "0.1.0")
go/pkg/ddapi/handler.go:242:    req.Header.Add("Content-Encoding", "snappy")
go/pkg/graphql/access.go:41:            req.Header.Add("x-hasura-admin-secret", g.secret)
```

`go/pkg/middleware/proxy.go:154` is the one we're patching here.

sreis:
Hit a know flake #47 tearing down the GCP cluster.

PR #1149, opened 2021-08-02
fix(RingTable): error handling

MoSattler:
Fixing this [error reported by sentry](https://sentry.io/organizations/opstrace/issues/2535043152/?environment=production&project=5529515&query=is%3Aunresolved):

Screenshot 2021-08-02 at 09 41 34

The underlying problem was that while we handle errors in the response, we did not handle errors on the request level (e.g. no network connection). 
I believe this might be an issue we could potentially run into many places where we do requests.

# Have you...

### DIFF
--- a/packages/app/src/client/views/ring-health/RingHealth.test.tsx
+++ b/packages/app/src/client/views/ring-health/RingHealth.test.tsx
@@ -183,6 +183,34 @@ test("handles request errors", async () => {
 
   const errorMessage = "Something bad happened"
 
+  nock("http://localhost").get(tab.endpoint).replyWithError({
+    message: errorMessage,
+    code: 'SOMETHING_VERY_BAD_HAPPENED',
+  })
+
+  const container = renderComponent(
+    <RingHealth title="some title" tabs={[tab]} />
+  );
+
+  jest.runOnlyPendingTimers();
+
+  expect(
+    await container.findByText("Could not load table")
+  ).toBeInTheDocument();
+  expect(
+    await container.findByText(errorMessage)
+  ).toBeInTheDocument();
+});
+
+test("handles response errors", async () => {
+  const tab = {
+    title: "Tab",
+    path: `/tab`,
+    endpoint: "/tab-endpoint"
+  };
+
+  const errorMessage = "Something bad happened"
+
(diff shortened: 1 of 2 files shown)


jgehrcke:
> I believe this might be an issue we could potentially run into many places where we do requests.

Oh yes. Wherever we do an HTTP request we absolutely need to handle the case where the request could not be sent out (in a timely manner) or when the response did not arrive (in a timely manner).

"in a timely manner" requires us to define what that means.

So, what we really want is to rather strictly follow a certain pattern / recipe.

One part of that recipe is to specify a timeout constant for the connect() timeout and/or a timeout constant for the time it takes until we did send the request out. Any error happening that prevents sending out the request (which includes DNS resolution errors, TCP errors while sending the request data, or said _timeout error_ (based on the chosen timeout constant)) then needs to be _explicitly handled_, usually with a retry until a higher-level retry stop criterion is hit. Often these things are done or can be done using the HTTP client abstraction in use -- it just needs to be properly configured and used.

MatApple:
> So, what we really want is to rather strictly follow a certain pattern / recipe.

Fully agree with @jgehrcke. @MoSattler we have a utility in our utils library called `httpcl` which is a wrapper around the `Got` library. Can you see if we are able to replace all our Axios uses in the app with `httpcl`? If you have any questions - @jgehrcke would be more than willing to help (he built that utility)

MoSattler:
@MatApple yes, will check it out!

jgehrcke:
> Can you see if we are able to replace all our Axios uses in the app with httpcl

Thanks for advocating into that direction @mat! :) A quick note about that: won't work like that because `httpcl` in `@opstrace/utils` is based on `got` which does not work in the browser, but is meant for the nodejs runtime. We kind of did settle on axios seemingly and of course with a bit of care and love the behavior I described can be achieved with axios. I actually tried to do that here: https://github.com/opstrace/opstrace/commit/cc4464f49a3d930d6c21b7acde5cb5b658388937

I also had a pretty keen eye on `ky` for the past year, which is built by the same people that build `got`, just for the browser.

MoSattler:
@jgehrcke @MatApple so where does that leave us now in terms of actionables?

PR #1150, opened 2021-08-02
ui: setup subscription client after session is verified

terrcin:
This resolves all the WS setup errors we see in the dev console before you login etc...

### DIFF
--- a/packages/app/src/client/components/withSession/pages/Login.tsx
+++ b/packages/app/src/client/components/withSession/pages/Login.tsx
@@ -15,8 +15,7 @@
  */
 
 import React from "react";
-
-import { useCommandService } from "client/services/Command";
+import { useHotkeys } from "react-hotkeys-hook";
 
 import { Page } from "client/components/Page";
 import { Box } from "client/components/Box";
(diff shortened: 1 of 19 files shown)


packages/app/src/client/components/withSession/pages/Login.tsx
```diff
@@ -26,15 +25,7 @@ import { Typography } from "client/components/Typography";
...
-  useCommandService(
-    {
-      id: "login-on-enter",
-      description: "Login",
-      keybindings: ["enter"],
-      handler: onLogin
-    },
-    []
-  );
+  useHotkeys("enter", onLogin);
```
MatApple, line 28: what was the motivation for this change? Just curious since it's not obvious.

Was it something to do with switching the hierarchy of the following?
```
<WithSession>
          <Services>
```

packages/app/src/state/clients/graphqlClient/subscriptionClient.ts
```diff
@@ -62,13 +69,22 @@ const subscriptionClient = () => {
...
     }),
     link: from([subscriptionErrorLink, wsLink])
   });
+
+  return subscriptionClient;
 };
 
-export default subscriptionClient();
+export const stopSubscriptionClient = () => {
+  subscriptionClient = null;
```
MatApple, line 83: shouldn't we technically call `.stop()` if there is a subscriptionClient? This doesn't actually stop it

PR #1136, opened 2021-07-29
[wip] ui: trying to include the buildinfo file test-browser image

terrcin:

PR #1154, opened 2021-08-02
Commit message prefix consolidation / simplification

jgehrcke:
Towards writing release notes, commit message prefixes should become more reliable.

This
* adds the `test-browser` prefix because it has become a real thing; CC @terrcin please use that :-).
* removes prefixes we have not used at all or not really used.
* removes prefixes we have mainly used 'wrongly' (such as 'fix').
* tries to sharpen the meaning of various component-based prefixes.

As always this state is evolving... I just wanted to have a little bit of an improvement now that we're in the release notes game.


jgehrcke:
of course, we also may want to actually enforce these _in CI_ (don't do that yet).

jgehrcke:
related: https://github.com/opstrace/opstrace/pull/1136#issuecomment-891059222

Issue #1137, opened 2021-07-29
ci instability: Login did not succeed within 300 s (browser-based flow)


jgehrcke:
```
[2021-07-29T12:57:14Z]   1) test_ui_api
[2021-07-29T12:57:14Z]        "before all" hook for "test_grafana_datasource_proxy_loki_get_labels":
[2021-07-29T12:57:14Z]      Error: Login did not succeed within 300 s
```

reported by @sreis here:
https://github.com/opstrace/opstrace/issues/1003#issuecomment-887923961
https://github.com/opstrace/opstrace/issues/1003#issuecomment-889165653

jgehrcke:
what seems to happen: 
```
[2021-07-29T12:51:20Z]   pw:api waiting for selector "css=button[type=submit]" +1ms
[2021-07-29T12:51:20Z]   pw:api   selector resolved to visible <button name="submit" type="submit" aria-label="Log In" …>…</button> +2ms
[2021-07-29T12:51:20Z]   pw:api attempting click action +1ms
[2021-07-29T12:51:20Z]   pw:api   waiting for element to be visible, enabled and stable +0ms
[2021-07-29T12:51:20Z]   pw:api   element is visible, enabled and stable +20ms
[2021-07-29T12:51:20Z]   pw:api   scrolling into view if needed +0ms
[2021-07-29T12:51:20Z]   pw:api   done scrolling +1ms
[2021-07-29T12:51:20Z]   pw:api   checking that element receives pointer events at (640,551.48) +0ms
[2021-07-29T12:51:20Z]   pw:api   element does receive pointer events +3ms
[2021-07-29T12:51:20Z]   pw:api   performing click action +0ms
[2021-07-29T12:51:20Z]   pw:api   click action done +11ms
[2021-07-29T12:51:20Z]   pw:api   waiting for scheduled navigations to finish +0ms
[2021-07-29T12:51:20Z]   pw:api   navigations have finished +0ms
[2021-07-29T12:51:20Z]   pw:api <= page.click succeeded +1ms
[2021-07-29T12:51:20Z]   pw:api => page.waitForSelector started +0ms
[2021-07-29T12:51:20Z]   pw:api waiting for selector "text=Getting Started" to be visible +0ms
[...]

PR #1165, opened 2021-08-03
browser-tests: debugging playwright cookie

terrcin:
Have the Playwright browser tests set a special `opstrace-playwright` cookie the react app can read. When set the `AccessDenied` page will show extra error info sent from the server which should be included in test failure screenshots.

### DIFF
--- a/packages/app/src/client/components/withSession/pages/AccessDenied.tsx
+++ b/packages/app/src/client/components/withSession/pages/AccessDenied.tsx
@@ -16,6 +16,7 @@
 
 import React from "react";
 import { useAuth0 } from "@auth0/auth0-react";
+import Cookies from "js-cookie";
 
 import { loginUrl } from "client/components/withSession/paths";
 
@@ -32,6 +33,8 @@ export const AccessDeniedPage = ({ data }: { data: {} }) => {
   const message = user
     ? `Contact your administrator or log out and try again with a different account.`
     : "Contact your administrator or try again with a different account.";
+
+  const isPlaywright = Cookies.get("opstrace-playwright");
 
   return (
     <Page centered height="100vh" width="100vw">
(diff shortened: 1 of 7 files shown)


packages/app/src/client/components/withSession/pages/AccessDenied.tsx
```diff
@@ -33,6 +34,8 @@ export const AccessDeniedPage = ({ data }: { data: {} }) => {
     ? `Contact your administrator or log out and try again with a different account.`
     : "Contact your administrator or try again with a different account.";
 
+  const isPlaywright = Cookies.get("opstrace-playwright");
```
MatApple, line 37: Let's sync about this before we merge. Why do we need to use a cookie? 

The bit I'm missing is why we can't just send the error we get from Auth0 (or any other server-side error) back to the UI and then use the appropriate Error component to show the error. In the case of a 403, we use the AccessDenied component, but for all other errors we could just use our ErrorView component

MatApple (changes requested):

jgehrcke:
> When set the AccessDenied page will show extra error info sent from the server which should be included in test failure screenshots.

I agree with @MatApple that maybe we do not want to have this asymmetry between CI env and prod env. If this is about debuggability, there are two more things we can do:

- We have the collections cluster to inspect the API's logs. How to use that I have shown here: https://github.com/opstrace/opstrace/issues/1137#issuecomment-891082681.

- If that is a little too much work, we can use the approach in https://github.com/opstrace/opstrace/blob/8af9e91d27f8f8056cda1364cc587a8f3d39e579/ci/create-cluster-artifacts.sh to pull the API container logs straight into the BK artifacts.

terrcin:
not doing this change

PR #1166, opened 2021-08-03
ui: removing use of CondRender, was accidently added back in via a long lived PR

nickbp:
Signed-off-by: Nahum Wild <nahum@opstrace.com>


nickbp:
(fixes build breakage caused by #1141 )

nickbp:
OK, the build failure has been fixed per the green checkmark above, and I've separately manually checked that the dialog in question is still working fine with the fix. Given all this I'm going to merge this now so that `main` is fixed again sooner rather than later.

PR #1173, opened 2021-08-03
ci: try to debug / tackle instability 1068

jgehrcke:
see #/1068


jgehrcke:
Rebased and removed the reading-time commit, force-pushed.

PR #1176, opened 2021-08-04
ci: sleep before test-remote (instability #944)

jgehrcke:
see commit msg


jgehrcke:
better solution tracked here: https://github.com/opstrace/opstrace/issues/1179

PR #1178, opened 2021-08-04
ui: updated the README.md quick start section as there are now more steps

terrcin:
Also tweaked the example `.env` files


terrcin:
not waiting for buildkite and merging as this is a change to docs only

PR #1188, opened 2021-08-05
ui: add preconfigured axios

MoSattler:
Adding a preconfigured axios instance to toy with the thought if we can address some of the issues raised by @jgehrcke in #1149 [here](https://github.com/opstrace/opstrace/pull/1149#issuecomment-891062142).

If this is a viable approach to the problem, I can think of some things that still would need to be discussed:

1. What options do we want to preconfigure with what values?
1. Which one of those options should be easily overwritable by a caller?

I am sure there are more considerations to be made.

### DIFF
--- a/packages/app/src/client/utils/request.ts
+++ b/packages/app/src/client/utils/request.ts
@@ -23,8 +23,33 @@ const request = axios.create({
 });
 
 request.defaults.raxConfig = {
+  // For requests that return a transient error (5xx).
   retry: 3,
-  instance: request
+
+  // For transient errors on transport level (DNS resolution, TCP connect()
+  // timeout, recv() timeout)
+  noResponseRetries: 3,
+
+  // Constant delay between attempts.
+  backoffType: "static",
+  // Delay between attempts in ms
+  retryDelay: 4000,
+
+  // HTTP methods to automatically retry
+  httpMethodsToRetry: ["GET", "DELETE", "PUT"],
+
+  // The response status codes to retry. 2 tuple array: list of ranges.
+  statusCodesToRetry: [
+    [100, 199],
+    [429, 429],
+    [500, 599]
+  ],
+
+  onRetryAttempt: function (err) {
+    const cfg = rax.getConfig(err);
+    //@ts-ignore cfg possibly undefined
+    console.log(`Retry attempt #${cfg.currentRetryAttempt} -- error: ${err}`);
+  }
 };
 
 rax.attach(request);
--- a/packages/app/src/client/utils/request.ts
+++ b/packages/app/src/client/utils/request.ts
@@ -1,0 +1,32 @@
+/**
+ * Copyright 2021 Opstrace, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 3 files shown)


packages/app/src/client/utils/request.ts
```diff
@@ -0,0 +1,16 @@
+import axios from "axios";
+import * as rax from 'retry-axios';
+
+const request = axios.create({
+  // For config options see https://github.com/axios/axios#request-config
+  timeout: 0,
```
jgehrcke, line 6: Let's set this to something non-zero, i.e. let's find a timeout constant we're comfortable with here. As this affects various requests, there is generally no one-size-fits-all timeout constant. But maybe (and you know best) there is.

If requests are small (just a few bytes) and if the response generation is known to be fast (sub-second) then let's set this to `5 s` or `10 s` because then we "know" that something is wrong, and retrying after 5 to 10 seconds already might heal the situation.

jgehrcke:
Great starting point! I left a comment about choosing the timeout constant :) Setting this to 0 means "wait indefinitely". :)

packages/app/src/client/utils/request.ts
MoSattler: Completely agree.

MoSattler: One thing regarding that, that is not quite in the scope of this is, that we need to do a better job in communicating these possible large request times to the user.
It might not be obvious to the user that the request has still a chance of succeeding after 5s. We might want to choose to show a spinner (or any other kind of "loading" state) in the UI or in certain situations. In some situations we do not seem to do that.

packages/app/src/client/utils/request.ts
```diff
@@ -0,0 +1,32 @@
...
+
+const request = axios.create({
+  // For config options see https://github.com/axios/axios#request-config
+  timeout: 5000
+});
+
+request.defaults.raxConfig = {
+  retry: 3,
+  instance: request
+};
```
MoSattler, line 28: Is it safe to generally just retry? Or should we only retry in certain failure scenarios?

packages/app/src/client/utils/request.ts
jgehrcke: > We might want to choose to show a spinner (or any other kind of "loading" state) in the UI or in certain situations

100 % agree. For anything that's expected to take longer than N seconds, show a spinner or something like that.

packages/app/src/client/utils/request.ts
jgehrcke: Depends on the specific request.

- It's the right thing to retry for all failures that result in the request not being set or the response not being received. That includes transport layer problems (network problems) and service outages.

- When the request was sent and a response was received, some responses are retryable:
    - usually _all_ 5xx ones
    - usually a 429 one, but that may be special (and require a larger delay between attempts
    - depending on the _specifics_ of the request other responses are retryable

jgehrcke: axios-retry probably documents what it retries by default. I suppose that it does (1) perfectly by default. I suppose it also retries all 5xx by default.

jgehrcke: @MoSattler I just remembered that I played with that before. See this commit for inspiration: https://github.com/opstrace/opstrace/commit/cc4464f49a3d930d6c21b7acde5cb5b658388937

MoSattler: For now, I just copied the rax config you had in place there. At least as a starting point.

packages/app/src/client/utils/request.ts
MatApple: yes fully agree!

MatApple: > In some situations we do not seem to do that

Not because we don't want to, just haven't got there yet!

PR #1199, opened 2021-08-06
ui(integrations): handle install error

MoSattler:
Handling request errors for installing integrations.

This PR is a superset of #1200 

# Have you...

### DIFF
--- a/packages/app/src/client/views/integrations/all/Install.tsx
+++ b/packages/app/src/client/views/integrations/all/Install.tsx
@@ -20,6 +20,8 @@ import { useParams, useHistory } from "react-router-dom";
 
 import {
   integrationDefRecords,
+  NewIntegration,
+  NewIntegrationOptions,
   showIntegrationPath
 } from "client/integrations";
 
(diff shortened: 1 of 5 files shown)


packages/app/src/client/views/integrations/all/Install.tsx
```diff
@@ -50,37 +44,56 @@ export const InstallIntegration = () => {
...
-            createFolder({ integration, tenant }).then(() =>
-              dispatch(loadGrafanaStateForIntegration({ id: integration.id }))
-            );
-          }
+      });
+    } catch (error) {
+      registerNotification({
+        state: "error" as const,
+        title: "Could not install integration",
+        information: error.response.errors?.[0].message ?? error.message
```
MatApple, line 68: is it possible for `error.response.errors` to be an empty array?

MoSattler: At least not according to the [graphql-spec](https://github.com/graphql/graphql-spec/blob/main/spec/Section%207%20--%20Response.md#errors)
> The errors entry in the response is a non-empty list of errors, where each error is a map.

jgehrcke: Meaning that if `errors` is not `undefined`, it is a non-empty array.

Checking for the `errors` key is a real important concept when getting GraphQL responses -- I also learned about that only recently, and added corresponding code commentary here:

https://github.com/opstrace/opstrace/blob/5ade2201e34d35bf60894e42ae4f7d67968a1490/test/test-remote/test_ui_api.ts#L451

MoSattler: @jgehrcke @MatApple 
Looking at the code, I think it might be enough to just read `error.message` - because the library we're using ([graphql-request](https://github.com/prisma-labs/graphql-request)) does a decent job covering [handling the errors](https://github.com/prisma-labs/graphql-request/blob/master/src/types.ts#L25-L55) and covering most cases, [including the one @jgehrcke  mentioned](https://github.com/prisma-labs/graphql-request/blob/master/src/types.ts#L49-L53).

The question is, if we're happy with [their error message format](https://github.com/prisma-labs/graphql-request/blob/master/src/types.ts#L30-L33), which seems to be more geared toward logging than presenting to user.

MoSattler: BTW - yes, GraphQL will return 200, even if there are field errors. Our client library will throw regardless.

PR #1200, opened 2021-08-06
ui(integration): fix typing

MoSattler:
Right now, the props of the integration components are not type checked. This PR changes that. More info in the code annotations.

This PR is a subset of #1199

### DIFF
--- a/packages/app/src/client/integrations/types.ts
+++ b/packages/app/src/client/integrations/types.ts
@@ -14,14 +14,34 @@
  * limitations under the License.
  */
 
-export type IntegrationDef = {
+import { Integration } from "state/integration/types";
+import { Tenant } from "state/tenant/types";
+
+export type FormProps<Data> = {
+  handleCreate: (
+    data: {
+      name: string;
+      data: Data;
+    },
+    options?: { createGrafanaFolder: boolean }
+  ) => void;
+};
+
+export type ShowProps = {};
+
+export type StatusProps = {
+  integration: Integration;
+  tenant: Tenant;
+};
+
+export type IntegrationDef<FormData = {}> = {
   kind: string;
   category: string;
   label: string;
-  desc: string | React.ReactType;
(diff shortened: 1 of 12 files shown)


packages/app/src/client/integrations/types.ts
```diff
@@ -14,14 +14,34 @@
...
+    options?: { createGrafanaFolder: boolean }
+  ) => void;
+};
+
+export type ShowProps = {}
+
+export type StatusProps = {
+  integration: Integration;
+  tenant: Tenant;
+}
```
MoSattler, line 35: As all the integration `<Form />`, `<Status />` and `<Show />` were made to have the same interface, I explicitly added that contract as a type.

packages/app/src/client/integrations/types.ts
```diff
@@ -14,14 +14,34 @@
...
   category: string;
   label: string;
-  desc: string | React.ReactType;
-  Form: React.ReactType;
-  Show: React.ReactType;
-  Status: React.ReactType;
+  desc: React.ReactNode;
+  Form: React.ComponentType<FormProps<FormData>>;
+  Show: React.ComponentType<ShowProps>;
+  Status: React.ComponentType<StatusProps>;
```
MoSattler, line 44: `ReactType` is deprecated, and in fact just prevented type checking altogether.

packages/app/src/client/views/integrations/all/Show.tsx
```diff
@@ -32,10 +32,5 @@ export const ShowIntegration = () => {
 
   if (!integrationDef) return <NotFound />;
 
-  return (
-    <integrationDef.Show
-      integration={integration}
-      integrationDef={integrationDef}
-    />
-  );
+  return <integrationDef.Show />
```
MoSattler, line 35: The now working type checking indicated that these props are in fact not needed.

PR #1197, opened 2021-08-06
cli: installer: gcp: follow VPC creation via observing long-running operation, retry on all levels and emit error detail

jgehrcke:
Context: #293.

Also, bump GCP client libraries.

Issue #293, opened 2021-01-22
installer: gcp: SQLInstance setup: error `Incorrect Service Networking config for instance: <snip>:NETWORK_PEERING_DELETED` not expected


jgehrcke:
this error must be expected and retried locally:
```
[2021-01-22T15:55:07Z] 2021-01-22T15:55:07.168Z info: Ensure SQLInstance exists
[2021-01-22T15:55:07Z] 2021-01-22T15:55:07.499Z info: create SQLInstance
[2021-01-22T15:55:12Z] 2021-01-22T15:55:12.983Z error: error during cluster creation (attempt 1):
[2021-01-22T15:55:12Z] GaxiosError: Invalid request: Incorrect Service Networking config for instance: ci-shard-ddd:bk-3454-afb-g-1611330907168:NETWORK_PEERING_DELETED.
[2021-01-22T15:55:12Z]     at Gaxios._request (/snapshot/build/node_modules/gaxios/src/gaxios.ts:117:15)
[2021-01-22T15:55:12Z]     at processTicksAndRejections (internal/process/task_queues.js:97:5)
[2021-01-22T15:55:12Z]     at JWT.requestAsync (/snapshot/build/node_modules/google-auth-library/build/src/auth/oauth2client.js:343:18) {
[2021-01-22T15:55:12Z]   response: [Object],
[2021-01-22T15:55:12Z]   config: [Object],
[2021-01-22T15:55:12Z]   code: 400,
[2021-01-22T15:55:12Z]   errors: [Array]
[2021-01-22T15:55:12Z] }
```
https://buildkite.com/opstrace/prs/builds/3454#718aac8b-a178-4129-8570-54d30a13f53f/247-508

Consumes a high level retry to date, which is why it's usually not fatal.

(Retrying solved the problem)

jgehrcke:
seen again locally:
```
$ ./opstrace --version --log-level debug
29884d4a-ci
2021-01-25T10:41:14.631Z debug: BUILD_INFO_COMMIT: 29884d4
2021-01-25T10:41:14.631Z debug: BUILD_INFO_TIME_RFC3339: 2021-01-25 08:10:27+00:00
2021-01-25T10:41:14.631Z debug: BUILD_INFO_HOSTNAME: ip-10-0-1-14.us-west-2.compute.internal
2021-01-25T10:41:14.632Z debug: BUILD_INFO_BRANCH_NAME: main
[...]

PR #1204, opened 2021-08-06
cli: create: gcp: log more detail

jgehrcke:
follow-up after https://github.com/opstrace/opstrace/pull/1197, for #293.

### DIFF
--- a/lib/gcp/src/globalAddress.ts
+++ b/lib/gcp/src/globalAddress.ts
@@ -107,6 +107,7 @@ export function* ensureAddressExists({
       address &&
       (address.status === "RESERVED" || address.status === "IN_USE")
     ) {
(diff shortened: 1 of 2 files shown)


jgehrcke:
Tried this exact commit locally, led to successful cluster creation. Merging, so that CI sampling over the weekend is even more productive.

lib/gcp/src/globalAddress.ts
```diff
@@ -107,6 +107,7 @@ export function* ensureAddressExists({
       address &&
       (address.status === "RESERVED" || address.status === "IN_USE")
     ) {
+      log.info(`${addressName} details: ${JSON.stringify(address, null, 2)}`);
```
MatApple, line 110: Good call! 🍿

PR #1209, opened 2021-08-07
ui(UninstallIntegrationButton): improve error handling

MoSattler:
Making the error handling a bit more expressive.

### DIFF
--- a/packages/app/src/client/integrations/common/UninstallIntegrationBtn.tsx
+++ b/packages/app/src/client/integrations/common/UninstallIntegrationBtn.tsx
@@ -25,9 +25,12 @@ import { Tenant } from "state/tenant/types";
 import { installedIntegrationsPath } from "../paths";
 
 import { deleteIntegration } from "state/integration/actions";
-import graphqlClient from "state/clients/graphqlClient";
+import graphqlClient, {
+  getGraphQLClientErrorMessage,
+  isGraphQLClientError
+} from "state/clients/graphqlClient";
 
-import { deleteFolder } from "client/utils/grafana";
+import { deleteFolder, isGrafanaError } from "client/utils/grafana";
 
 import { Button } from "client/components/Button";
 import { useSimpleNotification } from "client/services/Notification";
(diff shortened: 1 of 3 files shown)


packages/app/src/client/integrations/common/UninstallIntegrationBtn.tsx
```diff
@@ -55,23 +57,31 @@ export const UninstallBtn = ({
...
-        state: "error" as const,
-        title: "Could not uninstall integration",
-        information: error.response.errors?.[0].message ?? error.message
-      });
+      let notification: Parameters<typeof registerNotification>[0]
+      if (isGraphQLClientError(error)) {
+        notification = {
+          state: "error" as const,
+          title: "Could not uninstall integration",
+          information: error.response.errors![0].message
```
MoSattler, line 69: needs the `!` because the `ClientError` is not very precisely typed. 
Added https://github.com/prisma-labs/graphql-request/pull/282 upstream to fix that, let's see what they say.

packages/app/src/state/clients/graphqlClient/index.ts
```diff
@@ -47,6 +47,14 @@ export const isGraphQLClientError = (error: Error): error is ClientError => {
   return error instanceof ClientError;
 };
 
+export const getGraphQLClientErrorMessage = (error: ClientError): string => {
+  try {
+    return error.response.errors![0].message;
+  } catch (e) {
+    return `GraphQL Error (Code: ${error.response.status})`;
```
jgehrcke, line 54: Is `error.response` guaranteed to not be `undefined`?

In other words: if there was a transient issue and the request was not sent out, or no response arrived -- are we sure we do not call this function?

MoSattler: Yes - as `error` must be a `graphql-request` [`ClientError`](https://github.com/prisma-labs/graphql-request/blob/58a6d369d075b7b6cbd5ed0d6b1917c35786fe03/src/types.ts#L25-L55) (see [parameters](https://github.com/opstrace/opstrace/pull/1209/files#diff-129b93e0d8c903400c4b13ba4324668831194e24a3bab17df813dcdda1368eb0R50)).

`ClientError.response` [must not be undefined](https://github.com/prisma-labs/graphql-request/blob/20f18d3be52c85b4a0928ed8fe9ebba736e26f4c/src/types.ts#L26) and [is provided by `graphql-request`](https://github.com/prisma-labs/graphql-request/blob/20f18d3be52c85b4a0928ed8fe9ebba736e26f4c/src/index.ts#L229).

jgehrcke (approved):
:rocket: let's get this in!

PR #1220, opened 2021-08-09
ci: remove gcp projects d+f, add a+b+c

jgehrcke:
Small sample result is that #293 is limited to ddd and fff.

Showing the data set soon.

PR #1221, opened 2021-08-09
ci: pr pipeline: upgr test: revert FROM change

jgehrcke:
With the workaround https://github.com/opstrace/opstrace/pull/1220 being in place we can revert this change here.


jgehrcke:
Merging early, as this is still part of 'unbreaking CI'.

Issue #1059, opened 2021-07-15
ci-instability: fluentbit(grafana loki plugin) with basic auth: Error: Expectation not fulfilled within 30 s


sreis:
**Describe the bug**
```
[2021-07-15T05:16:25Z]   1) Loki API test suite
[2021-07-15T05:16:25Z]        fluentbit(grafana loki plugin) with basic auth:
[2021-07-15T05:16:25Z]      Error: Expectation not fulfilled within 30 s
[2021-07-15T05:16:25Z]       at Object.waitForLokiQueryResult (testutils/logs.ts:265:9)
[2021-07-15T05:16:25Z]       at runMicrotasks (<anonymous>)
[2021-07-15T05:16:25Z]       at processTicksAndRejections (node:internal/process/task_queues:96:5)
[2021-07-15T05:16:25Z]       at Context.<anonymous> (test_loki_log_ingest_api.ts:236:5)
```

https://buildkite.com/opstrace/scheduled-main-builds/builds/2668#edafbedf-6ca8-4372-99af-9b69af463dc9/6-588

jgehrcke:
I looked at the fluentbit log output :
```
[2021-07-15T05:02:01Z] [2021/07/15 05:01:50] [ info] [sp] stream processor started
[2021-07-15T05:02:01Z] ������ï¿½[2021/07/15 05:01:50] [debug] [input:tail:tail.0] inode=234972070 file=/tmp/opstrace-test-remote/fbit-log-input--30-KfjcyTMJp0DW-.log promote to TAIL_EVENT
[2021-07-15T05:02:01Z] ������ï¿½[2021/07/15 05:01:50] [ info] [input:tail:tail.0] inotify_fs_add(): inode=234972070 watch_fd=1 name=/tmp/opstrace-test-remote/fbit-log-input--30-KfjcyTMJp0DW-.log
[2021-07-15T05:02:01Z] ������I[2021/07/15 05:01:51] [debug] [task] created task=0x7f85cf437500 id=0 OK
[2021-07-15T05:02:01Z] ������ï¿½[0] tail.0: [1626325310.546936056, {"log"=>"aaaaaaa"}]
[2021-07-15T05:02:01Z] [1] tail.0: [1626325310.546945240, {"log"=>"bbbbbb"}]
[2021-07-15T05:02:01Z] [2] tail.0: [1626325310.546945449, {"log"=>"cccccccc"}]
[...]

PR #1225, opened 2021-08-09
cli(upgrade): exit early if buildinfo matches

### LINKED ISSUE #1223: cli(upgrade): check controller version before running the upgrade
**Describe the bug**

Discovered while triaging #1217.

The `upgrade` command flow stops the opstrace controller (sets the number of replicas in the deployment to zero) and then proceeds to check the controller image tag version. If they don't match then the image tag version is updated and the deployment is restarted (sets the number of replicas back to one).

When the opstrace controller image tags match the number of replicas is never set back to one.

sreis:
Close #1223

Move the buildinfo version check to as early as possible and checks if the `cliMetadata.allCLIVersions` image tag of the Opstrace controller deployment matches the one defined in the CLI buildinfo. Older Opstrace versions will have an empty cliMetadata field, for these, check the Opstrace controller deployment image and check it matches buildinfo.

### DIFF
--- a/packages/upgrader/src/upgrade.ts
+++ b/packages/upgrader/src/upgrade.ts
@@ -43,31 +43,110 @@ import { waitForControllerDeployment } from "./readiness";
 
 const CONTROLLER_IMAGE_DEFAULT = `opstrace/controller:${BUILD_INFO.VERSION_STRING}`;
 
-//
-// Set the controller deployment image version to the one defined in buildinfo.
-// Returns boolean to indicate controller deployment rollout initiated or not.
-export function* upgradeControllerDeployment(config: {
-  opstraceClusterName: string;
-  kubeConfig: KubeConfig;
-}): Generator<Effect, boolean, State> {
-  // Exit if controller deployment does not exist.
+// Checks if the  cliMetadata.allCLIVersions image tag of the Opstrace
+// controller deployment matches the one defined in buildinfo. Older Opstrace
+// versions will have an empty cliMetadata field, for these, check the Opstrace
+// controller deployment image and check it matches buildinfo.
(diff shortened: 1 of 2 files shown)


packages/upgrader/src/upgrade.ts
```diff
@@ -43,31 +43,110 @@ import { waitForControllerDeployment } from "./readiness";
...
+    const idx = opstraceControllerConfig.cliMetadata.allCLIVersions.length - 1;
+    const lastCLIVersion =
+      opstraceControllerConfig.cliMetadata.allCLIVersions[idx].version;
+
+    const installedControllerImageTag = installedControllerVersion
+      ?.split(":")
+      .pop();
+    if (lastCLIVersion !== installedControllerImageTag) {
+      die(
+        `found Opstrace version mismatch, CLI version ${lastCLIVersion} does not match installed controller version ${installedControllerImageTag}`
```
jgehrcke, line 78: :+1: 

In this error message, "CLI version" sounds like "the version of the CLI that's currently being executed".

Maybe let's expose an implementation detail here and make sure the error message is "correct". This is like an `assert()` in code where we really think this invariant is never violated anyway. So when this blows up, the error message may be optimized for debuggability / ugly.

Let's maybe say "... lastCLIVersion in the controller config is ${lastCLIVersion}, and does not match ..."

sreis:
Tests were stuck in the `cli-tests-with-cluster.sh` because of the DRY_RUN_UPGRADES test. I'll have a better look tomorrow but I'll probably remove that since we run the upgrade tests as part of the PRs.

sreis:
AWS failed with instability #1059. Not related to the PR since the flake is related to Loki and the PR touches the upgrade command logic.

PR #1230, opened 2021-08-09
wip: remove most resource requests and limits

triclambert:
See if we can spin up a smaller instance by removing the resource limits and requests.


triclambert:
This may not be possible because EKS has varying pod count limits per instance type.

https://github.com/awslabs/amazon-eks-ami/blob/73693cd549a830cf6c2b502b84ada18b537491b0/files/eni-max-pods.txt#L294

PR #1232, opened 2021-08-10
controller: don't flood 'Leaving x as-is' when opstrace annotation is removed

nickbp:
We want the logs to periodically report when something isn't being managed by the controller.
This ensures that "periodically" means every 5 minutes rather than every 5 seconds.

### DIFF
--- a/lib/kubernetes/src/reconciliation/reconcile.ts
+++ b/lib/kubernetes/src/reconciliation/reconcile.ts
@@ -118,6 +118,12 @@ export type ReconcileResourceTypes = {
   Cortices: V1Alpha1CortexResources;
 };
 
+// Keep track of the last time we logged about missing 'opstrace' annotations.
+// This avoids flooding logs when something is being ignored by the controller,
+// while still ensuring that it's consistently reported and visible.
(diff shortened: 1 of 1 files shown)


lib/kubernetes/src/reconciliation/reconcile.ts
```diff
@@ -480,9 +516,23 @@ export function* reconcile(
...
-      updateCollection
+      updateCollection,
+      notOursCollection
     );
 
+    if (notOursCollection.length !== 0) {
+      const nowMillis = new Date().getTime();
+      if (nowMillis - notOursLastLogMillis >= NOT_OURS_LOG_TTL_MILLIS) {
+        // Could log on a single line, but in practice there should only be a few of these at any time
+        notOursCollection.forEach(r => {
```
jgehrcke, line 527: For such things it's recommended to use a monotonic time source, especially when using long-running processes such as the controller is.

That's why I have built the `mtine()` and related functions in `@opstrace/utils`. Happy to do a follow-up patch.

jgehrcke (approved):
:rocket: merging this now, thanks @nickbp.

As commented, we should really use a monotonic time source here.

triclambert:
Thanks, @nickbp.

PR #1233, opened 2021-08-10
[wip] test-browser: run single auth spec to save logged in state before executing the entire suite which reuses it

terrcin:
This change updates how the playwright browser tests are executed on CI. the `OPSTRACE_PLAYWRIGHT_REUSE_STATE` env var is now set to `true` which means that the login fixtures and hooks will look for and use a previously logged in state (file on disk), if it's not there a normal login will occur and then the logged in state will be saved. 

On CI playwright is now run twice, once with a single worker running a single test that logs the user in resulting in the logged in state being saving to disk. The second and main run is now back to using multiple workers each reusing the logged in state rather than logging in. This means we can have any number of workers and only login once per test run.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -686,8 +686,9 @@ test-browser:
 		-e OPSTRACE_CLUSTER_NAME \
 		-e OPSTRACE_CLOUD_PROVIDER \
 		-e OPSTRACE_INSTANCE_DNS_NAME \
+		-e OPSTRACE_PLAYWRIGHT_REUSE_STATE=true \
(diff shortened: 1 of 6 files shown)


terrcin:
Going to merge this PR despite the failure which is an [unrealed ci-instability](https://github.com/opstrace/opstrace/issues/1226). Additionally a [branch of this code](https://github.com/opstrace/opstrace/pull/1237) simply changing the number of workers from 3 to 6 completely passed.

PR #1237, opened 2021-08-10
[wip] test-browser: what is the effect of 6 vs 3 workers on playwright runtime on CI?

terrcin:


terrcin:
one test run doesn't see to have made any difference. #science

PR #1255, opened 2021-08-11
chore: remove shard-eee based (see #293)

triclambert:
The GCP project `ci-shard-eee` has been exhibiting #293 more frequently than the other projects.  Let's remove it from rotation for now.


triclambert:
Oh, I see now that JP did this in 39ff3787710a474b91e9d3cda26bed62e31558da.  Nvm.

PR #1257, opened 2021-08-11
ui: do not load Sentry when running on localhost 

terrcin:
Now by default exceptions won't be sent to Sentry from the React app, this can be overriden with an env var. Additionally I've added Sentry exception tracking to the Express server and include BUILD_INFO data in the metadata tags sent from both server and client.


terrcin:
This change also includes BUILD_INFO data as tags in the metadata sent to Sentry, for example:

image

PR #1264, opened 2021-08-12
controller: fix fluentd config on gke 1.18+

### LINKED ISSUE #1254: Make fluentd support GKE 1.18 log format
With #1171 we now handle pods logs on GKE 1.19 but we're not forcing the GKE node upgrade and instead rely on the automatic upgrades. The node upgrade will only occur when the default GKE version is bumped (sometime this month). 

Unfortunately, it looks like the fluentd CRI plugin doesn't handle pods logs on GKE 1.18.

We need to have the opstrace controller deploy a version of fluentd that is compatible with GKE 1.18.

sreis:
When the controller notices it's running on GKE 1.18+ the fluentd config will not include the CRI parser.

Couldn't find anything in the Kubernetes javascript client that would return the API server version. I had to improvise and resorted to querying the /version endpoint manually. So, when the controller runs the reconcile loop on GCP it will fetch the API server version and cache it locally to avoid spamming the API server when the reconcile loop runs. 

I tested this locally by creating an instance on GCP with the latest available release of Opstrace (GKE version 1.18) and upgrading Opstrace to the version built from this PR branch. 

Close #1254

### DIFF
--- a/packages/controller/src/resources/apis/systemlogs.ts
+++ b/packages/controller/src/resources/apis/systemlogs.ts
@@ -27,6 +27,72 @@ import { Tenant } from "@opstrace/tenants";
 import { KubeConfig } from "@kubernetes/client-node";
 import { getTenantNamespace, getControllerConfig } from "../../helpers";
 import { DockerImages, getImagePullSecrets } from "@opstrace/controller-config";
+import axios, { AxiosResponse } from "axios";
+import { log } from "@opstrace/utils";
+
+// Variable to store the cached GKE version to avoid spamming the Kubernetes API
+// server.
+let cachedGKEVersion: {
+  major: string | undefined;
+  minor: string | undefined;
+} = {
+  major: undefined,
+  minor: undefined
+};
+
+// Checks if the API server version matches the given major and minor version.
+function* isGKEVersion(
+  kubeConfig: KubeConfig,
+  major: string,
+  minor: string
+): Generator<unknown, boolean, AxiosResponse> {
+  // Fetch GKE version and cache it.
+  if (
+    cachedGKEVersion.major === undefined ||
+    cachedGKEVersion.minor === undefined
+  ) {
+    if (kubeConfig.getCurrentCluster() === null) {
+      throw new Error(
+        `kubeconfig not configured properly, it is missing a cluster definition`
+      );
+    }
+    const server = kubeConfig.getCurrentCluster()?.server;
+    const endpoint = `${server}/version`;
+
+    const res: AxiosResponse<string> = yield axios.get(endpoint, {
+      timeout: 10000
+    });
+
+    if (res.status !== 200) {
+      throw new Error(
+        `failed to fetch Kubernetes API server version: ${JSON.stringify(res)}`
+      );
+    }
+
+    cachedGKEVersion = JSON.parse(res.data);
+    log.info(
+      `got Kubernetes API server version: ${JSON.stringify(cachedGKEVersion)}`
+    );
+  }
+
+  if (cachedGKEVersion.major === undefined) {
+    throw new Error(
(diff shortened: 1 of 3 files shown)


packages/controller/src/resources/apis/systemlogs.ts
```diff
@@ -27,6 +27,75 @@ import { Tenant } from "@opstrace/tenants";
 import { KubeConfig } from "@kubernetes/client-node";
 import { getTenantNamespace, getControllerConfig } from "../../helpers";
 import { DockerImages, getImagePullSecrets } from "@opstrace/controller-config";
+import axios, { AxiosResponse } from "axios";
```
jgehrcke, line 30: let's use `httpcl` based on `got` when we are in the nodejs runtime as we are in the controller.

Much better timeout control and error handling.

Let's use axios only in the browser if we have to. (`ky`) would be my preferred option.

sreis: > let's use httpcl based on got when we are in the nodejs runtime as we are in the controller.

But the controller is not using `got` it's using `axios`. Here's the prior art:  

https://github.com/opstrace/opstrace/blob/main/packages/controller/src/tasks/cortexRules.ts#L63

https://github.com/opstrace/opstrace/blob/main/packages/controller/src/tasks/syncTenants.ts#L70

What's the history on that then?

sreis: Ok, I ended up moving to httpcl anyway. :shrug:

sreis:
I had trouble calling a generator and/or an async function in the synchronous function `SystemLogAgentResources`. I ended up `fork`ing a function call to fetch the GKE version in the background. 

This is how the controller log output looks like before I tweaked it a bit to be less verbose.
```
2021-08-13T10:54:40.076Z info: starting system rules reconciler
2021-08-13T10:54:40.116Z info: starting reconciliation
2021-08-13T10:54:40.228Z info: got Kubernetes API server version: {"major":"1","minor":"19+","gitVersion":"v1.19.12-gke.2100","gitCommit":"82b41aad731f2dda24cc5ca99e09328163474782",>
2021-08-13T10:54:41.161Z info: checking if GKE version is major=1 minor=18+ cachedVersion={"major":"1","minor":"19+","gitVersion":"v1.19.12-gke.2100","gitCommit":"82b41aad731f2dda24>
2021-08-13T10:54:41.161Z info: fluentd parser:   <parse>
        @type cri
        merge_cri_fields false
      </parse>
2021-08-13T10:54:41.197Z info: API update request 1 of 1
2021-08-13T10:54:41.198Z info: Updating 1 DaemonSets:
2021-08-13T10:54:41.198Z info: Updating DaemonSets: system-tenant/systemlog
2021-08-13T10:54:41.198Z info: Updating 1 ConfigMaps:
2021-08-13T10:54:41.198Z info: Updating ConfigMaps: system-tenant/systemlog-fluentd-config
2021-08-13T10:54:42.068Z info: Waiting for 1 active DaemonSets:
- Waiting for DaemonSet system-tenant/systemlog rollout to finish: 0 out of 3 new pods have been updated
2021-08-13T10:54:43.213Z info: checking if GKE version is major=1 minor=18+ cachedVersion={"major":"1","minor":"19+","gitVersion":"v1.19.12-gke.2100","gitCommit":"82b41aad731f2dda24>
2021-08-13T10:54:43.213Z info: fluentd parser:   <parse>
        @type cri
        merge_cri_fields false
      </parse>
2021-08-13T10:54:44.247Z info: checking if GKE version is major=1 minor=18+ cachedVersion={"major":"1","minor":"19+","gitVersion":"v1.19.12-gke.2100","gitCommit":"82b41aad731f2dda24>
2021-08-13T10:54:44.247Z info: fluentd parser:   <parse>
        @type cri
        merge_cri_fields false
      </parse>
```

PR #1270, opened 2021-08-13
ci: Use age, not count, when deciding whether to delete CI buckets

nickbp:
The current logic avoids deleting the 20 most recent buckets.
However if we have a lot of concurrent CI jobs (>5), this can delete buckets that are being used by live tests.

To avoid this, we can explicitly check that the bucket was created at least 6 hours ago before we start trying to delete it.
This assumes that CI runs generally don't take longer than 6 hours.
This is done via a short stdlib-only python script because doing it in shell script gets pretty messy.


nickbp:
WIP: I've successfully run this manually/locally, but want to check that buildkite can run it. Assuming that works, I'll port the GCP script to have the same logic that the AWS script now has.

jgehrcke:
> This is done via a short stdlib-only python script because doing it in shell script gets pretty messy.

Huh! Yeah, ok :) we could have changed the 20 to 100 to get into a 'sane' state again, but of course conceptually the time cutoff is the better solution :).

Issue #1130, opened 2021-07-28
test-remote: end-to-end alert test failed: Failed to get non-zero alerts metric value after 300s


jgehrcke:
```
[2021-07-28T01:03:17Z]   21 passing (9m)
[2021-07-28T01:03:17Z]   4 pending
[2021-07-28T01:03:17Z]   1 failing
[2021-07-28T01:03:17Z]
[2021-07-28T01:03:17Z]   1) End-to-end alert tests
[2021-07-28T01:03:17Z]        End-to-end alerts for default and system tenants:
[2021-07-28T01:03:17Z]      Error: Failed to get non-zero alerts metric value after 300s. Are alerts successfully reaching the e2ealerting pod?
[2021-07-28T01:03:17Z]       at waitForE2EAlertFiring (test_alerts.ts:398:13)
[2021-07-28T01:03:17Z]       at Context.<anonymous> (test_alerts.ts:458:5)
[2021-07-28T01:03:17Z] 
[2021-07-28T01:03:17Z]
```
https://buildkite.com/opstrace/prs/builds/5222#5e7d212e-6f7f-4b34-bdfa-db85f9845df3/2980-4923

sreis:
```
[2021-07-29T09:25:48Z]   21 passing (9m)
[2021-07-29T09:25:48Z]   4 pending
[2021-07-29T09:25:48Z]   1 failing
[2021-07-29T09:25:48Z]
[2021-07-29T09:25:48Z]   1) End-to-end alert tests
[2021-07-29T09:25:48Z]        End-to-end alerts for default and system tenants:
[2021-07-29T09:25:48Z]      Error: Failed to get non-zero alerts metric value after 300s. Are alerts successfully reaching the e2ealerting pod?
[2021-07-29T09:25:48Z]       at waitForE2EAlertFiring (test_alerts.ts:398:13)
[2021-07-29T09:25:48Z]       at Context.<anonymous> (test_alerts.ts:453:5)
```
https://buildkite.com/opstrace/upgrade-tests/builds/1096#83b08d29-36ad-40d5-8d0a-e0c43056daa2/4006-5881

jgehrcke:
Seen six days ago by @terrcin, reported here: https://github.com/opstrace/opstrace/issues/1102#issuecomment-887923388

jgehrcke:
```
[2021-08-03T08:57:41Z]   21 passing (8m)
[2021-08-03T08:57:41Z]   4 pending
[...]

PR #1285, opened 2021-08-16
ci: push/pull CI container image.

jgehrcke:
So far the CI container image was never pushed to docker hub.

It was bound to the lifetime of the VM / disk.

Sometimes, a build finds empty / fresh state (fresh disk) and then the CI container image needs to be built from scratch. This takes a few minutes. Can reduce the duration of that process by pulling in some image layers stored on docker hub. (the last N layers need to be built and pushed for each build, and that's at the _con_ side of this change -- let's monitor the impact).

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -12,7 +12,7 @@ steps:
       - docker ps
       - bash ci/fetch-main.sh
       - echo "--- make rebuild-ci-container-image"
-      - make fetch-secrets && set-dockerhub-credentials && rebuild-ci-container-image && make push-ci-container-image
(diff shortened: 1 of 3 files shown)


jgehrcke:
Hm. Docker credentials are set up in `$HOME/.docker/config.json` and yet 
```
denied: requested access to the resource is denied
[2021-08-16T18:19:22Z] make: *** [push-ci-container-image] Error 1
```
Knowing that this works from within the CI container I will now move the push of the current CI container image into said container.

jgehrcke:
`main-test-gcp` failed with a known instability: https://github.com/opstrace/opstrace/issues/1130#issuecomment-900061816

PR #1287, opened 2021-08-17
gcp: list service connections during and after creation

jgehrcke:


jgehrcke:
`main-test-gcp` failed with known instability: https://github.com/opstrace/opstrace/issues/1130#issuecomment-900151485

PR #1291, opened 2021-08-17
cli: gcp: destroy: consolidations around region-to-destroy-in and subnetwork deletion

jgehrcke:


jgehrcke:
passed for GCP, merging

PR #1293, opened 2021-08-17
ui: use custom request function

MoSattler:
Adding the [custom request function](https://github.com/opstrace/opstrace/blob/11f1de65e6ba279d532d58ae5d74b0ff3a87fb49/packages/app/src/client/utils/request.ts) to two more places, both of which have good test coverage. 

Once they have been proven to work well, we might want to roll it out to further places, wherever it make sense.

# Have you...

### DIFF
--- a/packages/app/src/client/utils/grafana.ts
+++ b/packages/app/src/client/utils/grafana.ts
@@ -36,6 +36,7 @@ import { Dashboard as K8sLogsSummaryDashboard } from "client/integrations/k8sLog
 import { Dashboard as K8sMetricsApiServerDashboard } from "client/integrations/k8sMetrics/Show/dashboards/apiserver";
 import { Dashboard as K8sMetricsKubeletDashboard } from "client/integrations/k8sMetrics/Show/dashboards/kubelet";
 import { Dashboard as K8sMetricsResourceDashboard } from "client/integrations/k8sMetrics/Show/dashboards/resource";
+import request from "./request";
 
 const makeUuid = (integration: Integration) => `i9n-${integration.id}`;
 const makeUrl = (tenant: Tenant, path: string) =>
@@ -56,7 +57,7 @@ export async function createFolder({
   integration,
   tenant
 }: FolderProps): Promise<FolderInfo> {
-  const responseData = await axios({
+  const responseData = await request({
     method: "post",
     url: makeUrl(tenant, "folders"),
     data: {
@@ -107,7 +108,7 @@ export async function deleteFolder({
 }: FolderProps): Promise<{
   id: number;
 }> {
-  const responseData = await axios({
+  const responseData = await request({
     method: "delete",
     url: makeUrl(tenant, `folders/${makeUuid(integration)}`),
     withCredentials: true
--- a/packages/app/src/client/utils/request.ts
+++ b/packages/app/src/client/utils/request.ts
@@ -19,7 +19,7 @@ import * as rax from "retry-axios";
 
 const request = axios.create({
   // For config options see https://github.com/axios/axios#request-config
-  timeout: 5000
+  timeout: 10000
 });
 
 request.defaults.raxConfig = {


packages/app/src/client/utils/grafana.ts
```diff
@@ -46,7 +47,7 @@ export async function createFolder({
   integration,
   tenant
 }: FolderProps): Promise<FolderInfo> {
-  const responseData = await axios({
+  const responseData = await request({
```
jgehrcke, line 50: can you add a quick statement about why we think that the timeout and retrying parameters are a good choice for this request?

packages/app/src/client/utils/grafana.ts
```diff
@@ -97,7 +98,7 @@ export async function deleteFolder({
 }: FolderProps): Promise<{
   id: number;
 }> {
-  const responseData = await axios({
+  const responseData = await request({
```
jgehrcke, line 101: same here

packages/app/src/client/utils/grafana.ts
MoSattler: Both requests are usually part of a larger "request cascade". E.g. [when uninstalling integrations](https://github.com/opstrace/opstrace/blob/ac1a8ee15120656fa7e6e416a0ff29949b03ef18/packages/app/src/client/integrations/common/UninstallIntegrationBtn.tsx#L57-L61). 
As those are not atomic transactions, it would be unfortunate if only some of the requests go through and others time out due to temp network issues. 

Though I would love to hear your input on this, as I am not 100% sure if this is actually the way to go.

jgehrcke: things we need to understand:

- expected request processing time on the API handler. Less than one second? Less than 60 seconds? What's a reasonable upper bound beyond which we can safely assume thing is wrong?
- idempotency: are these safe to retry (ok when applied multiple times)?

jgehrcke: > if only some of the requests go through and others time out due to temp network issues.

Oh, ouchie. That kind of state should of course ideally be managed in the backend to make this kind of pseudo-atomic :).

But yeah I get you now that we're here and have to manage this transaction in the UI we have to try real hard to make all the requests succeed.

> the timeout and retrying parameters

With that I really meant the retry count, retry strategy, and timeout constants (the specific numbers)

jgehrcke: For clarity and for posterity: the retrying parameterization for the `request` object landed here: https://github.com/opstrace/opstrace/pull/1188/files#diff-f0b50639134328642967d005943bc837e05ff06b143b5c7dc56edaef8b4975cbR22 -- global timeout 5 seconds, 3 retries. Sounds good for all those HTTP requests that are expected to be handled _blazingly fast_.

For some other context recently I did set rather custom retrying parameters: https://github.com/opstrace/opstrace/blob/606fa4810cdec52ff6c1314b7c7ff4151144214b/packages/app/src/client/components/withSession/index.tsx#L528

jgehrcke:
@MoSattler I didn't meant to stall this. If you feel good about this change, please resolve the conflict and get this merged.

If my questions / reasoning is still a little unclear, let me try another way: can we expect the http request handlers affected here to respond much faster than within 5 seconds? If the answer is "yes": great, then the global timeout of 5 seconds plus retries will be an improvement.

If the answer is "no", then it will be a perfectly normal scenario for them to take around five seconds or more to respond: then the global timeout 5 seconds will worsen the situation.

MoSattler:
@jgehrcke do we have any metrics on this?

jgehrcke:
> @jgehrcke do we have any metrics on this?

I don't think so. This can usually be derived from the nature of the HTTP request handler. In our engineering lives we see different kinds of HTTP handlers, after all not from too many different categories.

### 1) fast, no  or little I/O and completely _within_ machine boundaries:

* Some handlers only need to execute code with very little CPU effort to generate response data and don't need to do I/O at all. Think `write('hello')`. That tells you that HTTP response generation is expected to take far far far below one second, more within nanoseconds or microseconds.
* Others handlers do a little bit of _machine-local_ I/O, and then you also know that response generation is expected to be super fast, within millisconds. Example: nginx static file serve for small files (with linux fs cache).

### 2) fast, but with I/O crossing machine boundaries
* Other handlers spend most of their time doing I/O, interacting with an external service _across machine boundaries_. Common examples are fast database lookups / insertions. When this works this is expected to take a few milliseconds. That's the typical db-driven web app scenario. Can fail, but rarely as of things getting 'slow' -- or if things are slow, the are already 'broken', and that justifies retrying. Expected response generation time: milliseconds.

### 3) slow, crossing machine boundaries
* Some http request handlers do complex remote interaction with potentially slow services, either over the Internet or with a potentially inherently slow API, and that kind of interaction may itself be subject to retrying. In those cases, the HTTP response generation may take seconds or minutes. We also have something like this in our system. Examples: querying Loki or Cortex for logs or metrics. Of course, certain queries are expected to be served only within, say, 30 seconds! And then here's another example for an HTTP request handler that is simply expected to take "a while" before returning a response, as of slow I/O with other serivices: https://github.com/opstrace/opstrace/blob/46a2c6f6feb5d2c7300e18b83e37e19f7cbffc90/packages/app/src/server/routes/api/cortex/cortexRuntimeConfig.ts#L101


As far as I see, the HTTP requests in question here interact with the Grafana HTTP API and after all trigger simple database insertions / lookups.

So, that should be category (2).

Btw, Grafana in our cases uses the cloud provider postgres DB: https://github.com/opstrace/opstrace/commit/0d9252f753e332448558fc8742fd7ad3763ecbb9

--------

If you can agree that we're talking (2) here, and maybe if you can briefly support that with anecdotal evidence (seeing such an http request being responded to in < 100 ms) then we're good to go with the 5 second timeout :)

MoSattler:
I will go through your comment in more detail on Monday, just a short note already:
I don't have any opinion on how long the timeout should be. I am just not sure how to determine the proper value.

jgehrcke:
>  I am just not sure how to determine the proper value.

My last comment should really help with that then :)

MoSattler:
Indeed very helpful, I am looking into it now!

MoSattler:
I had some requests taking more than 100ms. To be cautious, I raised the timeout to 10s. If that goes well for a while, we might want to consider lowering it at a later time.

packages/app/src/client/utils/request.ts
```diff
@@ -19,7 +19,7 @@ import * as rax from "retry-axios";
 
 const request = axios.create({
   // For config options see https://github.com/axios/axios#request-config
-  timeout: 5000
+  timeout: 10000
```
jgehrcke, line 22: For the record: we discussed the numbers for two specific HTTP requests. 

This change applies globally where this HTTP client is used -- I assume this means: also for other HTTP requests than the two discussed here.

Let's merge this now.

PR #1319, opened 2021-08-24
ci: add gcp networks clean up script

sreis:
Helper script to clean up stranded GCP networks from CI builds.


sreis:
Skipped builds because it's just a helper script to run locally.

PR #1321, opened 2021-08-24
ui: Integration for monitoring an external CockroachDB instance

nickbp:
The underlying config templates should support CockroachDB running on K8s or in an onprem/baremetal environment, but at the moment the UI config dialog just implements the K8s version. Planning to add a baremetal selection to the dialog in a later revision once the K8s version is out the door.


nickbp:
Going to merge this now and then bring back the `InstallInstructions` tests that had been temporarily removed via `fd01422`, as a separate PR. Mainly want to get the broad type annotation build fixes merged quickly to reduce the risk of merge conflicts.

In theory with the fixes merged we would be able to [unpin the version of `tsc` again](https://github.com/opstrace/opstrace/pull/1335) if we wanted.

Issue #1312, opened 2021-08-21
gcp: destroy: try hard to deleteServiceConnect() before destroying VPC


jgehrcke:
It's been confirmed by GCP engineering that service connections can leak behind the scenes and accumulate towards a hard limit. They are not garbage-collected when all reference is lost from a GCP API user's point of view (they can become orphaned, and then accumulate to said limit).

Action item is to try hard to deleteServiceConnect() before destroying VPC (and the cloudsql instance?). Existence of the VPC is the only way to retain reference to the service connection.

jgehrcke:
(related to #293)

PR #1326, opened 2021-08-25
gcp: destroy: try hard to deleteServiceConnect() before destroying VPC

jgehrcke:
https://github.com/opstrace/opstrace/issues/1312

GCP projects have an invisible quota for service connections. It's more of a GCP-internal sanity limit, not a user-facing quota.

Service connections may accumulate behind the scenes, even when deleting all referencing resources (namely, VPC networks, global addresses, and Cloud SQL instances).

This 'accumulation', as of GCP support, is for allowing to _quickly_ revive a service connection.

Now, the per-project hard limit of these dangling service connections seems to be 1024 (also as of GCP support).

Running into that limit resulted in the https://github.com/opstrace/opstrace/issues/1349 madness of the recent weeks which we accidentally tracked within #293 for a while.

This patch is to make it so that we try really hard to delete service connections associated with the VPC, before deleting the VPC.

### DIFF
--- a/lib/gcp/src/cloudSQL.ts
+++ b/lib/gcp/src/cloudSQL.ts
@@ -14,6 +14,8 @@
  * limitations under the License.
  */
 
+import { strict as assert } from "assert";
+
 import { call, delay, CallEffect } from "redux-saga/effects";
 import { google, sql_v1beta4, servicenetworking_v1 } from "googleapis";
 import { log, SECOND, sleep } from "@opstrace/utils";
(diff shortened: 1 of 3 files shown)


jgehrcke:
Trying locally. Found two interesting things:


```
info: services.connections.list for services/servicenetworking.googleapis.com and network projects/492745505745/global/networks/jpaug17:
{
  "connections": [
    {
      "network": "projects/492745505745/global/networks/jpaug17",
      "reservedPeeringRanges": [
        "google-managed-services-jpaug17"
      ],
      "peering": "servicenetworking-googleapis-com",
      "service": "services/servicenetworking.googleapis.com"
    },
    {
      "network": "projects/492745505745/global/networks/jpaug17",
      "reservedPeeringRanges": [
        "google-managed-services-jpaug17"
      ],
      "peering": "cloudsql-postgres-googleapis-com",
      "service": "services/servicenetworking.googleapis.com"
    }
  ]
}

```

This is not the same connection object, twice. Difference is in cloudsql-postgres-googleapis-com vs .servicenetworking-googleapis-com

jgehrcke:
Wow this is frustrating. Summary: Deleting the service connection(s) fails with this error:

> Failed to delete connection; Producer services (e.g. CloudSQL, Cloud Memstore, etc.) are still using this connection

But the CloudSQL instance is long gone.

jgehrcke:
Waiting for GCP support feedback before proceeding.

PR #1341, opened 2021-08-31
chore: misc typing fixes and a bump to TS 4.4.2 (latest)

nickbp:
Plays it safe and pins to exactly 4.4.2 for now. Could go with 4.4.x instead?

The errors seen before in CI around catch types seems to indeed be a 4.4.x thing:
https://devblogs.microsoft.com/typescript/announcing-typescript-4-4-rc/#use-unknown-catch-variables

### DIFF
--- a/lib/kubernetes/package.json
+++ b/lib/kubernetes/package.json
@@ -33,6 +33,6 @@
     "@types/request": "^2.48.4",
     "jest": "^26.6.3",
     "ts-jest": "^26.4.4",
-    "typescript": "^4.1.2"
+    "typescript": "4.4.2"
   }
 }
--- a/lib/kubernetes/package.json
+++ b/lib/kubernetes/package.json
@@ -32,7 +32,6 @@
     "@types/jest": "^26.0.16",
     "@types/request": "^2.48.4",
     "jest": "^26.6.3",
-    "ts-jest": "^26.4.4",
-    "typescript": "4.4.2"
+    "ts-jest": "^26.4.4"
   }
 }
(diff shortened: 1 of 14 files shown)


lib/kubernetes/package.json
```diff
@@ -33,6 +33,6 @@
     "@types/request": "^2.48.4",
     "jest": "^26.6.3",
     "ts-jest": "^26.4.4",
-    "typescript": "^4.1.2"
+    "typescript": "4.4.2"
```
jgehrcke, line 36: Ah. Now I understand. 

This `typescript` should not be here at all :) This should simply use the top-level package.json dev dependecy.

package.json
```diff
@@ -46,7 +46,7 @@
     "rimraf": "^3.0.0",
     "source-map-support": "^0.5.19",
     "ts-node": "^10.2.1",
-    "typescript": "4.3.5",
+    "typescript": "4.4.2",
```
jgehrcke, line 49: Right, that should be the authoritative typescript version requirement/definition for all the npm packages we have in our code base (except for those exceptions that _have to_ be exceptions for some reason, and I don't think we have those).

lib/kubernetes/package.json
nickbp: Removed this one, as well as one in packages/cli/ that appeared to also be a duplicate. Meanwhile these ones did not extend the root package so I left them as-is (with 4.4.2):
```
test/browser/package.json
test/test-remote/looker/package.json
test/test-remote/package.json
```

Issue #1343, opened 2021-08-31
ci-instability: Creating SQLInstance failed, retrying: GaxiosError: 502


sreis:
**Describe the bug**

```
[2021-08-31T09:34:03Z] 2021-08-31T09:34:03.299Z debug: Creating SQLInstance failed, retrying: GaxiosError: <!DOCTYPE html>
[2021-08-31T09:34:03Z] <html lang=en>
[2021-08-31T09:34:03Z]   <meta charset=utf-8>
[2021-08-31T09:34:03Z]   <meta name=viewport content="initial-scale=1, minimum-scale=1, width=device-width">
[2021-08-31T09:34:03Z]   <title>Error 502 (Server Error)!!1</title>
[2021-08-31T09:34:03Z]   <style>
[2021-08-31T09:34:03Z]     *{margin:0;padding:0}html,code{font:15px/22px arial,sans-serif}html{background:#fff;color:#222;padding:15px}body{margin:7% auto 0;max-width:390px;min-height:180px;padding:30px 0 15px}* > body{background:url(//www.google.com/images/errors/robot.png) 100% 5px no-repeat;padding-right:205px}p{margin:11px 0 22px;overflow:hidden}ins{color:#777;text-decoration:none}a img{border:0}@media screen and (max-width:772px){body{background:none;margin-top:0;max-width:none;padding-right:0}}#logo{background:url(//www.google.com/images/branding/googlelogo/1x/googlelogo_color_150x54dp.png) no-repeat;margin-left:-5px}@media only screen and (min-resolution:192dpi){#logo{background:url(//www.google.com/images/branding/googlelogo/2x/googlelogo_color_150x54dp.png) no-repeat 0% 0%/100% 100%;-moz-border-image:url(//www.google.com/images/branding/googlelogo/2x/googlelogo_color_150x54dp.png) 0}}@media only screen and (-webkit-min-device-pixel-ratio:2){#logo{background:url(//www.google.com/images/branding/googlelogo/2x/googlelogo_color_150x54dp.png) no-repeat;-webkit-background-size:100% 100%}}#logo{display:inline-block;height:54px;width:150px}
[2021-08-31T09:34:03Z]   </style>
[2021-08-31T09:34:03Z]   <a href=//www.google.com/><span id=logo aria-label=Google></span></a>
[2021-08-31T09:34:03Z]   <p><b>502.</b> <ins>That’s an error.</ins>
[...]

PR #1348, opened 2021-09-01
ci: remove ci-shard-iii from rotation

sreis:
This project seems to be hitting some hidden quota related to #1343. Remove it from rotation to get green CI again.


sreis:
AWS test hit known flake #1226.

PR #1356, opened 2021-09-02
controller: Set up ServiceMonitor and an initial dashboard with current metrics

nickbp:
Still a bit barebones but gets things set up for more metrics later.

dashboard


triclambert:
💗

Issue #923, opened 2021-06-22
Tenants added via UI after cluster creation are using invalid certificate for cortex/dd/loki ingress


nickbp:
In a cluster named `ship`, I created a set of tenants after the cluster was already running. This was done via following the [blog post](https://opstrace.com/blog/introducing-dynamic-tenant-addition-and-token-management) to create the keys, as well as adding the tenant entities in the UI. I saw that the new tenant namespaces were created and looked like they were working fine.

Afterwards, I was seeing errors when attempting to push metrics to `cortex.<tenant>` endpoints from an external prometheus instance:
> ts=2021-06-22T00:34:03.583Z caller=dedupe.go:112 component=remote level=warn remote_name=78c788 url=https://cortex.labels-a.ship.opstrace.io/api/v1/push msg="Failed to send batch, retrying" err="Post \"https://cortex.labels-a.ship.opstrace.io/api/v1/push\": x509: certificate is valid for ingress.local, not cortex.labels-a.ship.opstrace.io"

After disabling cert validation, I saw that the metrics were arriving in cortex successfully, so it looks like the ingress itself was correctly configured, but it was just using the wrong certificate.

I then checked one of the ingress controller pods and saw that it was flooding the logs with cert file validation warnings like this:
> $ kubectl logs -n ingress nginx-ingress-controller-api-5ljjq
> [...]
> W0621 23:05:49.576164       6 controller.go:1181] Validating certificate against DNS names. This will be deprecated in a future version
> W0621 23:05:49.576174       6 controller.go:1186] SSL certificate "series-b-tenant/https-cert" does not contain a Common Name or Subject Alternative Name for server "loki.series-b.ship.opstrace.io": x509: certificate is valid for *.dev.ship.opstrace.io, *.prod.ship.opstrace.io, *.staging.ship.opstrace.io, *.system.ship.opstrace.io, dev.ship.opstrace.io, prod.ship.opstrace.io, ship.opstrace.io, staging.ship.opstrace.io, system.ship.opstrace.io, not loki.series-b.ship.opstrace.io
> W0621 23:05:49.576184       6 controller.go:1187] Using default certificate
[...]

PR #1371, opened 2021-09-07
controller: manage tenant certs separately

nickbp:
The controller currently creates one certificate/secret up-front, and then copies the secret to the tenant and application namespaces. The secret is then marked immutable, which leads to problems if the certificate is updated later or if tenant SANs are added asynchronously to the cert (see #923).

This PR switches to a model of one cert object per tenant + one for the application namespace. The downside is that this means Ntenants+1 certs will be created with LetsEncrypt when the cluster is first created, when previously it was one cert for everyone. However it avoids the need to copy certificates across namespaces and keeps each tenant on a granular certificate. It also removes the prior immutable setting, allowing certs to be renewed automatically when they're approaching their expiration.

An alternate strategy would be to create a single `*.cluster.domain.io` wildcard cert to be shared by everybody, thereby reducing the number of interactions with LetsEncrypt. However we would still likely want to remove the immutable flag on the secret in order to allow expiration updates.


nickbp:
Looks like the alternative method mentioned above of having `*.cluster.domain.io` wouldn't work, as the wildcard only works for one subdomain level, not multiple. For example `*.cluster.domain.io` is valid for `staging.cluster.domain.io` but not `cortex.staging.cluster.domain.io`. So we'll need to continue managing per-tenant subdomains as in this PR.

From [Letsencrypt](https://community.letsencrypt.org/t/acme-v2-production-environment-wildcards/55578):
> DNS names in certificates may only have a single wildcard character, and it must be the entire leftmost DNS label, for instance “*.example.com”. A single certificate can have wildcard DNS names for multiple base domains, and can also mix in non-wildcard names.

MatApple:
> as the wildcard only works for one subdomain level

We ran into that previously too!

PR #1395, opened 2021-09-14
ui: type refactor + prettier

MoSattler:
# Have you...

### DIFF
--- a/lib/dns/src/opstrace.ts
+++ b/lib/dns/src/opstrace.ts
@@ -84,11 +84,10 @@ async function loginPollRequest(
   let data: any;
   try {
     data = JSON.parse(resp.body);
-  } catch (err: any) {
+  } catch (err) {
     log.warning(
-      "could not parse response body as JSON: %s: %s",
-      err.code,
-      err.message
+      "could not parse response body as JSON: %s",
+      (err as SyntaxError).message
     );
 
     return undefined;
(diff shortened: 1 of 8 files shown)


lib/dns/src/opstrace.ts
```diff
@@ -84,11 +84,10 @@ async function loginPollRequest(
   let data: any;
   try {
     data = JSON.parse(resp.body);
-  } catch (err: any) {
+  } catch (err) {
     log.warning(
-      "could not parse response body as JSON: %s: %s",
-      err.code,
```
MoSattler, line 90: JSON.parse throws `SyntaxError`, which does not have a `code` member. See [mdn](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/JSON/parse#exceptions).

packages/app/src/client/components/NotificationsList/NotificationsList.tsx
```diff
@@ -40,21 +39,28 @@ export type NotificationsListProps = {
...
+  onDeleteAll
+}: NotificationsListProps) => {
   const classes = useStyles();
 
-  const listNode = React.useRef<any>();
+  const listNode = React.useRef<HTMLUListElement>();
   const [height, setHeight] = React.useState<number>(0);
 
   useEffect(() => {
-    setHeight(listNode.current?.clientHeight + 45);
```
MoSattler, line 57: Not quite sure if I understand this line.

if `listNode.current` is indeed `undefined` (as was anticipated with the `?` operator), it would evaluate to 
```typescript
undefined + 45
```
which is `NaN`

PR #1398, opened 2021-09-16
ui(cortex-config): fix error handling

MoSattler:
Guarding against improve typing of generic errors.

# Have you...


packages/app/src/state/cortex-config/sagas/cortexConfigSubscription.ts
```diff
@@ -116,7 +116,7 @@ export function cortexConfigSubscriptionEventChannel(): EventChannel<Actions> {
             emitter(actions.setCortexRuntimeConfig(res.data));
           })
           .catch(err => {
-            if (ServerError.isInstance(err.response.data)) {
+            if (axios.isAxiosError(err) && err.response && ServerError.isInstance(err.response.data)) {
```
MoSattler, line 119: `ServerError.isInstance(err.response.data)` alone will break if an generic, non-request error is thrown, as it doesn't have the `response` field.

PR #1407, opened 2021-09-20
loki: switch ingesters to podManagementPolicy=Parallel (DELETE/READD STATEFULSET TO UPDATE)

nickbp:
This allows getting out of a circular dependency where running ingesters are failing readiness probes due to another ingester being down, preventing that ingester from coming back.

This mirrors a similar change that had been made to the Cortex ingester StatefulSet in June 2021, see e2a2229.

# Have you...


nickbp:
Above upgrade failure looks like a tooling/env issue (CLI failing to download) rather than a problem with the change itself.

But I'd still really want to make sure this change passes upgrades before merging so I'll just retry CI later. It's not that urgent.

Issue #1409, opened 2021-09-21
cortex: Avoid returning 500-599 errors to remote_write clients


nickbp:
There are scenarios where Cortex can be overrun by push client requests. The server starts to return 500 errors, which the clients treat as retryable. The retries themselves can then prevent the clients from ever catching up, so they flood the server continuously. A small batch of 500 errors can then feed on itself and grow into an indefinite outage.

In this example, we can see 500 errors greatly outpacing other responses. The 500 errors are considered "retryable", while the 400 errors are due to the requested data being too old. There's even a gap in the middle of the chart where system metrics are being blocked by the thundering herd of retries, caused by global distributor queues filling up:
![134001156-d019639a-0363-43a9-a5ea-9f0d5f14f2c1](https://user-images.githubusercontent.com/35933/134093512-52fa3b73-4fe1-47c7-b27b-6b2c5ee53f18.png)

For debugging the current behavior, we can look at the Prometheus `remote_write` [HTTP error handling](https://github.com/prometheus/prometheus/blob/e582e907f8a97ffae27123b9632a254e5018a4a6/storage/remote/client.go#L233-L246) and [retry handling](https://github.com/prometheus/prometheus/blob/e582e907f8a97ffae27123b9632a254e5018a4a6/storage/remote/queue_manager.go#L1217-L1267):
- 500-599: retry after `remote_write.queue_config.min_backoff` (with expontential backoff up to `remote_write.queue_config.max_backoff`)
- 429 (if `remote_write.queue_config.retry_on_http_429` is enabled): retry after provided `Retry-After` header duration, or `min_backoff` -> `max_backoff` as above
- else: don't retry

The retry logic itself defaults to waiting `min_backoff=30ms` on the first retry, up to `max_backoff=100ms` on following retries. Each batch of `max_samples_per_send=500` has up to `batch_send_deadline=5s` to get sent. It looks like the default used to be `max_samples_per_send=100` but people were saying that [this was too low](https://github.com/prometheus/prometheus/issues/5166).

[...]

PR #1410, opened 2021-09-21
cortex: remap 'retryable' 5xx errors to non-retryable errors

nickbp:
When cortex is backing up (e.g. full distributor queues) it can return 500 errors.
If clients retry-loop on these errors, it can lead to nobody being able to catch up.

Note that this is separate from the existing 429->503 mapping, which stays as-is.

See also https://github.com/opstrace/opstrace/issues/1409

### DIFF
--- a/go/cmd/cortex/main.go
+++ b/go/cmd/cortex/main.go
@@ -39,15 +39,95 @@ var (
 	cortexDistributorURL     string
 	tenantName               string
 	disableAPIAuthentication bool
+	allowPushRetries         bool
 )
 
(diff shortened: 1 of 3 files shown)


nickbp:
Upgrade test failure is due to the CLI download URL returning an S3 error. Not really relevant to the error handling change. Also seen in unrelated PR #1407.

nickbp:
Deployed the patched `cortex-api` by hand and it quickly cleaned up an environment that had been stuck in this scenario. Given that it seems to do the job, going to go ahead and merge as-is

Issue #1260, opened 2021-08-12
Revisit implementation of data retention (who deletes what and why?)


jgehrcke:
For enforcing retention settings (for the actual deletion), I'd like to try rely only on one mechanism; ideally the built-in mechanism provided by Loki and Cortex. I had commented on that before [here](https://github.com/opstrace/opstrace/pull/149#issuecomment-742788800)

> I am hoping that we can have loki/cortex delete objects from index and bulk storage in a consistent fashion, based on retention settings we configure them with :).

https://github.com/opstrace/opstrace/pull/149 does not really clarify the role / relevance of the lifecycle-based object deletion for retention that we introduced there. 

Let's decide between the following:

1) Loki and Cortex built-in mechanism is the one we rely on. Only that.
2) Loki and Cortex built-in mechanism is the one we rely on. But to be really sure (whereas we don't know if we need that) we _also_ want to use a lifecycle-based deletion of old objects in a bucket. Maybe with a good amount of leeway. Say: let Loki and Cortex implement a data retention of N days, and then we make the lifecycle rule implement an (N+L) days policy
3) We only want the lifecycle-based solution.

I think we should work with (1) for a while and see if we need the safety net provided by (2).

Or maybe I am missing something and we need (2) :).

----

August 2021, built-in retention primitives:

For Cortex with blocks storage: use `compactor_blocks_retention_period`. See https://github.com/cortexproject/cortex/blob/40b623b10f01ad10f22dc59968335b954e65b3e5/docs/configuration/config-file-reference.md.

For Loki, use the compactor retention configuration: https://github.com/grafana/loki/blob/055c35e47abb9f1891b6775ad79d5856597d4246/docs/sources/operations/storage/retention.md#compactor

jgehrcke:
As of today, the Loki per-tenant retention config can be set via the runtime config file, i.e. changes take effect w/o having to restart components

```
# Feature renamed to 'runtime configuration', flag deprecated in favor of -runtime-config.file (runtime_config.file in YAML).
# CLI flag: -limits.per-user-override-config
[per_tenant_override_config: <string>]
```
with the per-tenant override config file allowing for per-tenant retention settings, as for example for tenant `29` in this example:
```
overrides:
    "29":
        retention_period: 168h
        retention_stream:
        - selector: '{namespace="prod"}'
          priority: 2
          period: 336h
        - selector: '{container="loki"}'
          priority: 1
          period: 72h
```

PR #1418, opened 2021-09-23
chore: add missing reference for data retention parameters

triclambert:
Adding missing data retention parameters:  `log_retention_days` and `metrics_retention_days`.

### DIFF
--- a/docs/references/configuration.md
+++ b/docs/references/configuration.md
@@ -95,7 +95,7 @@ gcp:
     zone_suffix: a
 ```
 
-Note: the example above shows the defaults.
+The example above shows the defaults.
 
 **Note:**
 we develop and test mainly with `region` set to `us-west2`.
@@ -263,5 +263,5 @@ metrics_retention_days: 365
 Notes:
 
 * These options currently cannot be specified on a per-tenant basis.
-* These options currently cannot be changed after initial cluster creation.
+* These options (as with the other options in the configuration file) currently cannot be changed after initial cluster creation.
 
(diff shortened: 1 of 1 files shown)


triclambert:
Changes can be previewed here:  https://preview.opstrace.com/1418/docs/references/configuration.html#log_retention_days-and-metrics_retention_days

docs/references/configuration.md
```diff
@@ -247,3 +246,22 @@ Note that the controller does not get deployed when initiating `create` with the
...
+
+```yaml
+log_retention_days: 90
+metrics_retention_days: 365
+```
+
+Notes:
+
+* These options currently cannot be specified on a per-tenant basis.
+* These options currently cannot be changed after initial cluster creation.
```
jgehrcke, line 266: > These options currently cannot be changed after initial cluster creation.

Fair enough! But this might make it sound like the other parameters can be changed after initial cluster creation :) Most of them cannot. We do not have a config mutation concept, we never paid attention to this. If any one of the parameters can currently be changed upon "upgrade" then this work so to say as of being lucky, not by design.

One way to approach this is to qualify individual parameters one-by-one in manual testing, and then add a note in the lines of "this parameter can be changed ... ".

jgehrcke:
For connecting the dots: https://github.com/opstrace/opstrace/issues/1260

docs/references/configuration.md
triclambert: Actually, none of the config file parameters can be changed during the upgrade.  So I'll note that and merge.

jgehrcke: > Wait, is this correct?

I think I tried to answer that with my comment above. Let me know which part needs clarification.

Issue #1302, opened 2021-08-19
graphql-hasura maintenance bump: make 1.3 -> 2.x transition


jgehrcke:
See https://github.com/opstrace/opstrace/blob/b00d2a20701f2455f1340bb1b23a5de9b85011ca/packages/app/Dockerfile.graphqlserver#L1 

We are on 1.x, the ecosystem has moved forward to 2.x with several point releases.

MoSattler:
This transition would alleviate a security problem we have where a dependency in app, `hasura-cli@1.3.3`, adds vulnerable packages to our app.
There is a hot fix for this problem in https://github.com/opstrace/opstrace/pull/1433, but the proper way of handling this would be updating all the tooling to version >2.

PR #1433, opened 2021-09-28
chore: update hasura-cli/axios

MoSattler:
Forcing the `hasura-cli@1.3.3` dependency to use an updated `axios@0.21.1`, as it's required `axios@0.19.1` has the known vulnerability https://github.com/advisories/GHSA-4w2v-q235-vp99.

While this version of axios is technically out of [`hasura-cli@1.3.3` specified range](https://github.com/jjangga0214/hasura-cli/blob/v1.3.3/package.json#L94), this update should be safe, because:
1. The version bump [does NOT contain any breaking changes](https://github.com/axios/axios/blob/master/CHANGELOG.md).
1. [`axios` is NOT extensively used by `hasura-cli`](https://github.com/jjangga0214/hasura-cli/search?q=axios)

Though, this change should be seen as a hot fix. The proper way of achieving a safe version of `Hasura-cli/axios` would be to update `hasura-cli@1.3.3` to `hasura-cli@>2`, where this issue is fixed. Though there are [many breaking changes](https://github.com/hasura/graphql-engine/blob/master/CHANGELOG.md) to consider. See also https://github.com/opstrace/opstrace/issues/1302 regarding the update.

# Have you...

### DIFF
--- a/package.json
+++ b/package.json
@@ -59,7 +59,8 @@
     "**/**/lodash": "^4.17.19",
     "**/**/object-path": "^0.11.5",
     "**/**/serialize-javascript": "^3.1.0",
-    "**/**/yargs-parser": "^20.2.6"
+    "**/**/yargs-parser": "^20.2.6",
+    "**/app/hasura-cli/axios": "0.21.1"
   },
   "scripts": {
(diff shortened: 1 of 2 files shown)


MatApple (approved):
Thanks @MoSattler LGTM!

PR #1446, opened 2021-09-30
chore: update prometheus to 2.10.0

MoSattler:
Addresses [CVE-2019-3826](https://nvd.nist.gov/vuln/detail/CVE-2019-3826)

Also needed to do some changes to pointers due to a change in prometheus (https://github.com/prometheus/prometheus/pull/4957) where TimeSeries' signature changes

> changes to the proto definitions, made it so that certain fields can't be null which results in []type rather than []*type and therefore less object allocations. - [source](https://github.com/prometheus/prometheus/pull/4957#issuecomment-469163748)

# Have you...


go/go.mod
```diff
@@ -19,7 +19,7 @@ require (
 	github.com/morikuni/aec v1.0.0 // indirect
 	github.com/prometheus/client_golang v1.7.1
 	github.com/prometheus/common v0.10.0
-	github.com/prometheus/prometheus v2.5.0+incompatible
+	github.com/prometheus/prometheus v1.8.2-0.20190525122359-d20e84d0fb64
```
MoSattler, line 22: Corresponds to prometheus [version 2.10.0](https://github.com/prometheus/prometheus/tree/v2.10.0)

MoSattler:
I am not sure why, but these changes are causing a lint failure: https://buildkite.com/opstrace/prs/builds/5918#4b5b275c-7554-4340-8a04-47b3e921c46d/1393-1942

jgehrcke:
> Addresses CVE-2019-3826

For the record: that CVE is about a security issue in the Prometheus UI. Does not at all apply to our usage of the Prometheus lib here in this context.

jgehrcke:
@sreis can you review the code changes here? :)

jgehrcke:
Having built the unit and integration tests for the datadog-compatible API I can at least say that I am happy to merge if these tests pass.

sreis (approved):
lgtm but in the future, instead of a WIP commit please add more details on why the change is required. It could be something like: The pointer changes are required because the [WriteRequest struct was updated](https://github.com/prometheus/prometheus/blob/v2.10.0/prompb/remote.proto#L23).

PR #1448, opened 2021-10-01
go authenticator: migrate to jwt lib v4

jgehrcke:
from https://github.com/dgrijalva/jwt-go to https://github.com/golang-jwt/jwt
https://github.com/golang-jwt/jwt/blob/main/MIGRATION_GUIDE.md

Added tests for more confidence that this does not introduce an embarrassing security regression.

### DIFF
--- a/go/README.md
+++ b/go/README.md
@@ -1,4 +1,14 @@
 
+# Run unit tests
+
+The unit tests manage Docker containers, i.e. the Docker daemon needs to be running.
+
+
+```
+cd go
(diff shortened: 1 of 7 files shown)


jgehrcke:
```
[2021-10-01T10:34:36Z] pkg/authenticator/jwt.go:56:33: SA1019: jwt.StandardClaims is deprecated: Use RegisteredClaims instead for a forward-compatible way to access registered claims in a struct. (staticcheck)
```

Issue #1465, opened 2021-10-13
Post-PoC: Implement ClickHouse replication


nickbp:
ClickHouse replication requires either running a ZK instance or enabling the pre-production "[ClickHouse Keeper](https://clickhouse.com/docs/en/operations/clickhouse-keeper/)" feature.

Leaning towards trying out ClickHouse Keeper first and seeing how it goes, since it's already 'pre-production' despite only being added to the docs [in July](https://github.com/ClickHouse/ClickHouse/commit/623b368a707cf17713cf38079161439b9b0c7a6e#diff-380a02c367deb97304024ceaf14298817d5ccc43eaa2b84a9ec46752906e4b6a). If it doesn't work then we can revert to using ZK via [this operator](https://github.com/pravega/zookeeper-operator), using ClickHouse [docs](https://github.com/Altinity/clickhouse-operator/blob/master/docs/zookeeper_setup.md) and [example](https://github.com/Altinity/clickhouse-operator/blob/master/deploy/zookeeper/quick-start-persistent-volume/zookeeper-3-nodes.yaml) [configs](https://clickhouse.com/docs/en/operations/tips/#zookeeper).

Issue #1462, opened 2021-10-13
E2E PoC: Multi-tenant authenticated OTLP ingest


nickbp:
Add API authenticated endpoint for OTLP input for receiving spans. As with existing APIs, the endpoint should check auth headers then route the data to the correct tenant.

- gRPC at [port 4317](https://github.com/open-telemetry/opentelemetry-collector/tree/main/exporter/otlpexporter), https/protobuf at [port 4318](https://github.com/open-telemetry/opentelemetry-collector/blob/main/exporter/otlphttpexporter/testdata/config.yaml#L28)
- gRPC ingress may need [support enabled](https://kubernetes.github.io/ingress-nginx/examples/grpc/)
- For client auth, [gRPC](https://github.com/open-telemetry/opentelemetry-collector/blob/main/exporter/otlpexporter/testdata/config.yaml#L27-L32) supports both custom headers and “authenticator extensions” incl [bearer](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/extension/bearertokenauthextension), while [HTTPS](https://github.com/open-telemetry/opentelemetry-collector/blob/main/exporter/otlphttpexporter/testdata/config.yaml#L28) supports custom headers only
- Jaeger doesn't natively support OTLP yet so we need to translate to Jaeger/gRPC. Can either directly embed [OTLP->Jaeger translation](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/9d729d282feeff238bb8add32b011419415f56ca/pkg/translator/jaeger/traces_to_jaegerproto.go) within auth checker, or deploy per-tenant otel-collector-contrib instances to translate/forward. Direct embed would reduce the hops and could be easier overall to implement/test.

nickbp:
WIP branch: `nick/tracing-ingress`

Going to just try directly translating OTLP->Jaeger in the authenticated api handler, feels like it'll be simpler overall vs configuring and running a separate otlp-collector-contrib container.

nickbp:
The translation is implemented and wasn't too bad. However need to test it. Was able to do it via a custom module on top of the existing otel-collector code. After finishing it I happened to find this doc describing what I ended up doing, so that seems like a good sign: https://opentelemetry.io/docs/collector/custom-auth/

[...]

Issue #1461, opened 2021-10-13
E2E PoC: Multi-tenant deployment via Controller


nickbp:
Take the result of #1460 and start integrating it into Opstrace, adding multi-tenant support along the way.

Update controller to support tenant create/destroy
- Controller client for listing/creating/destroying ClickHouse DBs. [One DB per tenant](https://github.com/jaegertracing/jaeger-clickhouse/blob/main/guide-multitenancy.md), synced from tenant list.
- Create [per-tenant Jaeger instances](https://github.com/jaegertracing/jaeger/issues/2509), each pointing to respective ClickHouse DBs.
- Create per-tenant Jaeger dashboard endpoint similar to current Grafana.

Note that #1462 separately covers implementing the ingest endpoint for sending spans from outside the Opstrace instance.

nickbp:
WIP branch: https://github.com/opstrace/opstrace/tree/nick/tracing-controller

nickbp:
Implemented via #1471

PR #1471, opened 2021-11-02
controller: Deploy basic multitenant Jaeger/ClickHouse tracing stack

nickbp:
Updates the controller to deploy a minimal multitenant tracing stack (implements #1461).

- CRDs/operators for ClickHouse and Jaeger.
- Deploys a shared ClickHouse instance (single-pod for now), where each tenant has its own user/database.
    - Implements ClickHouse client that automatically creates/deletes tenant databases/users in ClickHouse.
- Deploys per-tenant Jaeger instances (each all-in-one for now).
    - Configures authenticated `<tenant>.<cluster>.opstrace.io/jaeger` ingress for viewing the Jaeger UI - equivalent to the `/grafana` ingress.

Notes:
- An ingress for sending traces hasn't been enabled yet, that'll be done next (#1462). The UI endpoint meanwhile shows traces from Jaeger reporting on itself.
- There are a few `TODO(nickbp)`s left at this point but it felt like a good point to get the current iteration out the door:
    - Implement better change detection for the `V1Jaeger` CRD
    - Update ClickHouse DB name when replication is added (#1465)
    - Ensure everything being deployed is using `imagePullSecrets`, in some places this may need underlying operator support?


jgehrcke:
Trying to revert this so that we can produce a release from `main` with more minimal changes compared to the last release.

PR #1478, opened 2021-11-16
bump: update several package versions with findings from dependabot

triclambert:


triclambert:
Merging without review.  We can revert if needed.

PR #1482, opened 2021-11-17
chore: AWS custom DNS name

sreis:
Changelog of changes in the branch:

* Require the `custom_dns_name`, `custom_auth0_client_id`, and `custom_auth0_domain` parameters to be set when creating/upgrading an instance.
* Change AWS CI job to use a custom DNS name in the `opstraceaws.com` root domain.
* Create a hosted zone in Route53 for the new instance.
* Add an NS entry in the opstraceaws.com root zone for the new instance hosted zone.
* On teardown, delete the instance hosted zone and remove the NS entry in the `opstraceaws.com` root zone.
* Disables upgrade tests in CI PR pipeline because it does not support running with a custom domain name. 

**NOTE**
*After merging the PR, we need to disable the scheduled upgrade tests because that pipeline does not support custom domain names.*

### DIFF
--- a/packages/installer/src/aws.ts
+++ b/packages/installer/src/aws.ts
@@ -20,7 +20,6 @@ import yaml from "js-yaml";
 import AWS from "aws-sdk";
 import { all, call, fork, join } from "redux-saga/effects";
 
-import { ensureDNSExists } from "@opstrace/dns";
 import {
   createVPC,
   ensureSubnetsExist,
@@ -57,11 +56,7 @@ import {
   RDSSubnetGroupRes
 } from "@opstrace/aws";
 
-import {
-  getClusterConfig,
-  getDnsConfig,
-  LatestClusterConfigType
-} from "@opstrace/config";
+import { getClusterConfig, LatestClusterConfigType } from "@opstrace/config";
 
 import {
   ConfigMap,
@@ -69,7 +64,7 @@ import {
   createOrUpdateConfigMapWithRetry
 } from "@opstrace/kubernetes";
 
-import { log, getBucketName, sleep, entries } from "@opstrace/utils";
+import { log, getBucketName, sleep, entries, die } from "@opstrace/utils";
 
 import { getAWSConfig } from "@opstrace/config";
 import { EnsureInfraExistsResponse } from "./types";
@@ -126,15 +121,12 @@ export function* ensureAWSInfraExists(): Generator<
   // State-mutating API calls below.
   yield call([new ServiceLinkedRoleRes(ccfg.cluster_name), "setup"]);
 
-  if (ccfg.custom_dns_name === undefined) {
-    yield call(ensureDNSExists, {
-      opstraceClusterName: ccfg.cluster_name,
-      dnsName: getDnsConfig(ccfg.cloud_provider).dnsName,
-      target: ccfg.cloud_provider,
-      dnsProvider: ccfg.cloud_provider
-    });
-  } else {
-    log.info("skip DNS setup, custom DNS name set");
+  if (!ccfg.custom_dns_name) {
+    die("custom_dns_name is not set");
+  }
+
+  if (!ccfg.custom_auth0_client_id || !ccfg.custom_auth0_domain) {
+    die("custom_dns_name and custom_auth0_domain are not set");
   }
 
   const lokiBucketName = getBucketName({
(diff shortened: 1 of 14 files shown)


sreis:
I tried the yup schema validation initially but the changes were not worth the trouble. Anyway, this PR should be reverted in the near future to bring back the functionality.

@jgehrcke feel free to add a commit to change the message to your liking and merge when you are ready to cut a release.

jgehrcke:
Two manual test cases:

```
2021-11-17T12:30:07.540Z error: invalid instance config document: custom_dns_name is a required field
```

```
2021-11-17T12:30:37.473Z error: invalid instance config document: custom_auth0_domain is a required field
```

sreis:
> I just went through the trouble, and pushed. In an abstract way I would say that the trouble is the following: an automatic migration cannot be performed when starting to require new parameters. I would say from the start that was 'flawed' about the migration approach we took for the user-given config: a framework that could take care of parameter rename, but which cannot do anything about additional information that must be provided by the user. My opinion: in the future, we can resort to documenting breaking changes and ask the user to implement them, including parameter renames.

Another option would be to introduce a new schema, V3, with the fields marked as required, and add a V2toV3 upgrade function which would notify the user it would need to define the new fields. This is where I stopped because the changes are (supposedly) temporary.

Thank you for trying to improve things though.

jgehrcke:
> Thank you for trying to improve things though.

Thanks for saying this -- and to be clear, this was / is supposed to be an improvement through simplification. The patch itself is complex and I'd rather not have it on this branch either, but the result after all is a simplification that I quite like for now.

packages/installer/src/aws.ts
```diff
@@ -126,15 +121,12 @@ export function* ensureAWSInfraExists(): Generator<
...
-      dnsProvider: ccfg.cloud_provider
-    });
-  } else {
-    log.info("skip DNS setup, custom DNS name set");
+  if (!ccfg.custom_dns_name) {
+    die("custom_dns_name is not set");
+  }
+
+  if (!ccfg.custom_auth0_client_id || !ccfg.custom_auth0_domain) {
+    die("custom_dns_name and custom_auth0_domain are not set");
```
jgehrcke, line 129: (this, too)

packages/installer/src/gcp.ts
```diff
@@ -58,18 +53,12 @@ export function* ensureGCPInfraExists(
...
-
-    log.info("DNS name has been set up: %s", dnsname);
-  } else {
-    log.info("skip DNS setup, custom DNS name set");
+  if (!ccfg.custom_dns_name) {
+    die("custom_dns_name is not set");
+  }
+
+  if (!ccfg.custom_auth0_client_id || !ccfg.custom_auth0_domain) {
+    die("custom_dns_name and custom_auth0_domain are not set");
```
jgehrcke, line 61: these are effectively asserts/noops now.

packages/installer/src/gcp.ts
```diff
@@ -58,18 +53,12 @@ export function* ensureGCPInfraExists(
...
 
-  if (ccfg.custom_dns_name === undefined) {
-    const dnsConf = getDnsConfig(ccfg.cloud_provider);
-    const dnsname = yield call(ensureDNSExists, {
-      opstraceClusterName: ccfg.cluster_name,
-      dnsName: dnsConf.dnsName,
-      target: ccfg.cloud_provider,
-      dnsProvider: ccfg.cloud_provider
-    });
-
```
jgehrcke, line 69: Note: might want to re-introduce code path involving Opstrace DNS service for internal dev cluster deployment.

jgehrcke:
> After merging the PR, we need to disable the scheduled upgrade tests because that pipeline does not support custom domain names.

Disabled, by archiving it.
`Pipeline upgrade-tests archived successfully.`

Context: 
https://forum.buildkite.community/t/switch-to-disable-pipelines/1144/3

> Sorry for the delay :bowing_man:t2:‍♂ We have added the ability to Archive Pipelines 31.

jgehrcke:
> Disables upgrade tests in CI PR pipeline because it does not support running with a custom domain name.

OK. At the core of this is I think our code (de)duplication challenge between https://github.com/opstrace/opstrace/blob/8370f532245b7d930b6e27802b30f73200ed153e/ci/deploy-testremote-teardown.sh and https://github.com/opstrace/opstrace/blob/8370f532245b7d930b6e27802b30f73200ed153e/ci/test-upgrade/create-cluster.sh etc.

